PR #132, opened 2020-10-23
Persistent Volume setup

ilchebedelovski:


build/helm/sifnode/templates/sifnode-deployment.yaml
```diff
@@ -64,14 +64,14 @@ spec:
...
-#          - name: data
-#            mountPath: /data
-#      volumes:
-#      - name: data
-#        persistentVolumeClaim:
-#          claimName: {{ .Values.pvc.name }}
-#      {{- with .Values.nodeSelector }}
+          volumeMounts:
+          - name: data
+            mountPath: /data
```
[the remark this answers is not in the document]
intl-man: Yeah, my bad @ilchebedelovski - @utx0 is correct. I was only thinking about the blockchain database itself, but there are other config/keys that do require persistent storage.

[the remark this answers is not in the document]
intl-man: The mount point could just be `/root`, and then that'd cover both.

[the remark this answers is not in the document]
ilchebedelovski: I have changed the mountPath

build/Rakefile
```diff
@@ -37,9 +37,23 @@ namespace :cluster do
     puts "Deploy cluster config: #{path(args)}"
     system("cd #{path(args)} && terraform apply") or exit 1
     puts "Cluster #{path(args)} created successfully"
+
+    puts "Deploy the EBS CSI plugin on k8s"
+    cmd = %Q{kubectl apply -k "github.com/kubernetes-sigs/aws-ebs-csi-driver/deploy/kubernetes/overlays/stable/?ref=master"}
```
utx0, line 42: Hey @ilchebedelovski what's this and why is it needed? Also this might not be the best place for it as we want these rake tasks to support gcp, az,etc in the furture too.

utx0 (changes requested):

build/Rakefile
ilchebedelovski: @utx0 ideally, this should be placed in the Terraform module, this is the CSI plugin deployment from here https://aws.amazon.com/premiumsupport/knowledge-center/eks-persistent-storage/ , but that `-k` option is not supported for kubectl in Terraform, that's why I have placed in the Rakefile.

Maybe we can place it in sifnode:install function in Rakefile, what do you think?

utx0: I would prefer this is added to the aws terraform modules some how. Any reason that cant happen?

utx0:
@ilchebedelovski can you also rebase this branch with `develop` The Rakefile structure has also changed there

build/Rakefile
ilchebedelovski: It is because we cannot execute `kubectl apply -k` over the kubectl provider in Terraform, this is important because we are installing whole directory and -k means `-k, --kustomize='': Process a kustomization directory.`.

utx0: Would this help: https://github.com/kbst/terraform-provider-kustomization

ilchebedelovski: @utx0 I have solved the issue with the kustomization provider, thanks

utx0 (approved):
This all looks good. @ilchebedelovski can you confirm for me that you have done and full e2e test of this code starting from zero to hero on all this?

build/terraform/providers/aws/main.tf
```diff
@@ -70,3 +84,53 @@ module "eks" {
...
+        "ec2:DeleteSnapshot",
+        "ec2:DeleteTags",
+        "ec2:DeleteVolume",
+        "ec2:DescribeInstances",
+        "ec2:DescribeSnapshots",
+        "ec2:DescribeTags",
+        "ec2:DescribeVolumes",
+        "ec2:DetachVolume"
+      ],
+      "Resource": "*"
```
intl-man, line 110: We could probably lock this down to the actual cluster resource we've created... but for now it'll be ok.

intl-man (approved):

ilchebedelovski:
@intl-man @utx0 I've made end-to-end testing on the Persistent Volume setup, there aren't any issues during the deployment process. The roles are created and attached successfully, the CSI plugin is installed with the `Kusotomization` provider and the volumes are created and attached automcatically from the Sifnode Helm Chart.

PR #137, opened 2020-10-24
Add under construction message to frontend repo

ryardley:
As we are now going open source this adds an under construction message to the frontend repo.


utx0 (approved):
lol

PR #141, opened 2020-10-26
[ChainOps] Feature/cli faucet

utx0:
This PR adds two features:

* Add an update to the build system to enable either a testnet or mainnet binary build. This then enables adding testnet only features. This th following;
* Added liangping's ping faucet module. 

This needs tesing.


utx0:
stale PR

PR #142, opened 2020-10-26
Ui components

novelja:
- `sass-loader` added for compilation of SCSS files
- `normalize-scss` added to reset the browser styles
- initialized files for variables, mixins, and global styles

- VueJS components folder structure and basic components with basic props/slots initialized:
- Elements (small components used to build blocks):
-- SifButton
-- SifInput
-- SifSwitch
-- SifSelect
-- InputGroup
-- IconButton
-- Dropdown

- Block (combinations of more element-size components)
-- CurrencyForm: input/select combo for selecting tokens
-- SifApp: Representing the main "app card" from the design, combining multiple elements

- Layout (grid components for design purposes and layout size components that are usually used only once per page)
-- Header - app header
-- Menu
-- Main - container for the main app card

- `/ui` route added for components showcase


ui/app/src/components/Blocks/CurrencyForm.vue
```diff
@@ -0,0 +1,83 @@
...
+
+  methods: {
+    onChange(event, data) {
+      console.log(data);
+    }
+  }
+}
+</script>
+
+<style lang="scss">
```
ryardley, line 63: Sorry if this is a stupid question from a React guy but Is there any reason we are not using scoped scss? One of the great things you get out of Vue is CSS Modules out of the box so you can avoid having to use strict naming convention systems like SMACSS and BEM and unnecessary nesting.

ryardley (commented):
1. Nice work. This is quite clear how this is setup and things are looking great. 

1. I **much** prefer the way you have put together the modal as it makes it easier to communicate between components. Perhaps we can take your approach but use [Teleport](https://v3.vuejs.org/guide/teleport.html) to make a general global Dialog component?

Following are my opinionated opinions please take with a grain of 🧂

1. Happy to be shown otherwise but I am not yet convinced that "Block" and "Element" from BEM are great abstractions with which to organise components. BEM is a 2014 attempt to solve grouping and reusability problems in a globally namespaced environment which we don't have because things that change together can just be placed near each other in folders. Component trees are self similar and fractal and it seems that well named nested folders organised by category allows for more encapsulation/refactorability. Having said that I think the main differentiator should be is this a 'feature-specific' component like a "swap panel" versus a 'general` one like a "button".

1. I love BEM in a global single file CSS situation but because of the inbuilt modularity of vue I am not convinced we necessarily need to follow such a verbose rigid naming convention for classnames. (See other [comment](https://github.com/Sifchain/sifnode/pull/142#discussion_r512499660))

ui/app/src/components/Blocks/CurrencyForm.vue
ryardley: I think the main casualty is readability as you go from this:

```html
<div class="menu">
  <div class="menu__item">
    <a class="menu__link menu__link--active" href="#">Discover</a>
  </div>
  <div class="menu__item">
    <a class="menu__link" href="#">New Wallet</a>
  </div>
  <div class="menu__item">
    <a class="menu__link" href="#">Convert</a>
  </div>
  <div class="menu__item">
    <a class="menu__link" href="#">Deposit</a>
  </div>
</div>
```

```html
<div class="menu">
  <div class="item">
    <a class="link active" href="#">Discover</a>
  </div>
  <div class="item">
    <a class="link" href="#">New Wallet</a>
  </div>
  <div class="item">
    <a class="link" href="#">Convert</a>
  </div>
  <div class="item">
    <a class="link" href="#">Deposit</a>
  </div>
</div>
```

ryardley: Also you might want to pull out the item to its own component and make dingbat__item a block. 

```html
<div class="dingbat__item">
  <a class="dingbat__link" href="#">New Wallet</a>
</div>
```

with styles

```scss
.dingbat {
  ...

  &__item {
    ...
  }

  &__link {
    ...
  }
}
```

This becomes quite different:

```html
<div class="item">
  <a class="item__link" href="#">New Wallet</a>
</div>
```

with styles

```scss
.item {
  ...
  &__link {
      ...
  }
}
```

Where as if we use scoped scss we can generally get away with just cutting and pasting both style and markup.

```html
<div class="item">
  <a class="link" href="#">New Wallet</a>
</div>
```
```scss
.item {
  .link {
    ...
  }
}

```

ui/app/src/components/Elements/SifSelect.vue
```diff
@@ -0,0 +1,143 @@
...
+            @click="selectOption($event, token)">
+            <span>{{ token.name }}</span>
+          </div>
+        </div>
+      </div>
+    </div>
+  </div>
+</template>
+
+<script>
```
ryardley, line 29: We probably need to work out what our TypeScript policy is and have a discussion around that.

ryardley:
Gonna close this in favour of the other UI branch

PR #144, opened 2020-10-27
fixed asymmetric withdraw

Jedi2002:


x/clp/types/msgs.go
```diff
@@ -116,10 +116,10 @@ func (m MsgRemoveLiquidity) ValidateBasic() error {
 	if !m.ExternalAsset.Validate() {
 		return sdkerrors.Wrap(ErrInValidAsset, m.ExternalAsset.Symbol)
 	}
-	if !(m.WBasisPoints > 0) {
+	if !(m.WBasisPoints > 0) || m.WBasisPoints >= 10000 {
 		return sdkerrors.Wrap(ErrInvalidWBasis, strconv.Itoa(m.WBasisPoints))
 	}
-	if m.Asymmetry > 10000 || m.Asymmetry < -10000 {
+	if m.Asymmetry >= 10000 || m.Asymmetry <= -10000 {
```
khdegraaf, line 122: I believe an asymmetry of +-10,000 is perfectly fine, as long as wBasis isn't 100% too and you are the only validator.  Is this limitation what you wanted @AustinoBombino?

khdegraaf (changes requested):
Not sure about one detail, see my comment.

x/clp/types/msgs.go
Jedi2002: Added some checks for rejecting transactions if the pool is too shallow . 
If you are the only lp , and try to withdraw everything ( wBasis 10000 ,Asymmetry -10000 /0/10000 ) it will reject the tx because it will result in the pool being too shallow. 

If we have more than 1 lp , you would be able to withdraw  everything you have added  ( wBasis 10000 ,Asymmetry -10000 /0/10000 ) 

For cases with wbasis less than 10000 , it still works the same .

x/clp/types/msgs.go
```diff
@@ -116,7 +116,7 @@ func (m MsgRemoveLiquidity) ValidateBasic() error {
 	if !m.ExternalAsset.Validate() {
 		return sdkerrors.Wrap(ErrInValidAsset, m.ExternalAsset.Symbol)
 	}
-	if !(m.WBasisPoints > 0) {
+	if !(m.WBasisPoints > 0) || m.WBasisPoints >= 10000 {
```
AustinoBombino, line 119: @Jedi2002 I think we want to change this to` || m.WBasisPoints > 10000`

x/clp/types/msgs.go
AustinoBombino: Yeah we should accept any values within the ranges of wBasis: 0 to 10000 and asymmetry -10000 to 10000, but while processing the withdrawal check to make sure that the withdrawal/swap won't overdraw the pool and throw an error if it will.

khdegraaf (approved):
Looks good to me now.

PR #152, opened 2020-10-29
remove the generate bin step since all solidity bin file checked in

juniuszhou:
Some tester get all kinds of error when run init scripts to generate bin files. comment the step out and use the bin file already checked in. Will add it back, reference issue 1.


utx0 (approved):
approved

PR #155, opened 2020-10-30
feature/ebrelayer-genesis

intl-man:
Integrate ebrelayer.


intl-man:
It might be an idea to split out `sifnodecli` and `ebrelayer` into their own containers..

cmd/ebrelayer/txs/signature.go
```diff
@@ -8,19 +8,13 @@ import (
...
 
 	"github.com/Sifchain/sifnode/cmd/ebrelayer/types"
 )
 
 // LoadPrivateKey loads the validator's private key from environment variables
 func LoadPrivateKey() (key *ecdsa.PrivateKey, err error) {
-	// Load config file containing environment variables
-	if err := godotenv.Load(); err != nil {
-		log.Fatal("Error loading .env file", err)
-	}
```
ElliotFriedman, line 22: So now there is no more config file like we discussed and there is only environment variables that we read in

ElliotFriedman (changes requested):
Could you please add some documentation on how to run this docker image and set the env vars?

cmd/ebrelayer/txs/signature.go
intl-man: Yes, I'm not a fan of having a `.env` file floating around in a production environment. I feel it's better to set the environment variables when spinning up the environment.

build/rakelib/genesis.rake
```diff
@@ -14,19 +14,30 @@ namespace :genesis do
     end
 
     desc "Boot the new scaffolded network in docker-compose"
-    task :boot, [:chainnet] do |t, args|
-      if args[:chainnet] == nil
+    task :boot, [:chainnet, :eth_address, :eth_keys, :eth_websocket] do |t, args|
+      trap('SIGINT') { puts "Exiting..."; exit }
```
utx0, line 18: whats the trap do?

cmd/ebrelayer/txs/signature.go
```diff
@@ -8,19 +8,13 @@ import (
 
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/crypto"
-	"github.com/joho/godotenv"
```
utx0, line 11: This could be used in place of the removed line and then its still possible to use `.env` for local development too:

`import _ "github.com/joho/godotenv/autoload"`

utx0 (commented):
couple of minor comments

cmd/ebrelayer/txs/signature.go
utx0: Yeah I would gitignore all `.env` file and then they can only be used for local development.

build/rakelib/genesis.rake
intl-man: Just a way to exit a program gracefully, without ruby throwing errors when a `SIGINT` is sent to the process. Without it, you end up with something like the following:

```
rakelib/genesis.rake:83:in `system'
/rakelib/genesis.rake:83:in `build_docker_image'
build/rakelib/genesis.rake:39:in `block (3 levels) in <top (required)>'
```

cmd/ebrelayer/txs/signature.go
intl-man: Nice - will add.

build/rakelib/genesis.rake
utx0: oh nice.

utx0 (approved):
LGTM

ElliotFriedman (approved):

PR #156, opened 2020-10-30
Fix vue event issues by avoiding hyphens in events

ryardley:
There was a strange bug in vue that appears to be resolved by avoiding hyphens in events. I suspect a framework bug.


ryardley:
@novelja

PR #169, opened 2020-11-02
Feature/mock api

utx0:
* cleaned up swagger spec
* created a mock server setup .


api/openapi/sifchain-openapi.yml
```diff
@@ -1,14 +1,150 @@
...
-  title: CLP_GET_REQUESTS
-  description: CLP_GET_REQUESTS
-  version: '0.1'
+  title: Sifnode - OpenAPI spec
+  description: OpenAPI spec for sifnode
+  version: '0.0.1'
 servers:
   - url: 'http://0.0.0.0:1317'
 paths:
+  /clp/getPools:
```
intl-man, line 9: @utx0 Do we want the verb names included in the endpoint name? Maybe we should rename this `/pools` ? (and similarly with the other endpoints too).

utx0: tihs is from cosmos remember.

utx0: This is how the clp module works in cosmos-sdk land.

intl-man (approved):

jkbrooks (approved):

Issue #175, opened 2020-11-03
CLP Genesis 


intl-man:
When generating a new genesis file, the following error is returned:

```
ERROR: failed to validate genesis state: failed to unmarshal clp genesis state: UnmarshalJSON cannot decode empty bytes
```

as it appears that the initial CLP Genesis state is emtpy.

utx0:
@Jedi2002 I think we have a need in clp to make sure the genesis export state is exported correctly and completely. Otherwise we could have big upgrade issues. cc @jkbrooks @AustinoBombino @musnit. We had to commect some of the code out as this was blocking us from deployments.

musnit:
yeh i think you're right

intl-man:
@utx0 @musnit On further investigation, this was my fault - I was marshalling genesis with an old (pre-CLP) struct. Since been fixed and merged (#177).

PR #177, opened 2020-11-03
hotfix/genesis

intl-man:
Genesis hotfix.

1. Resolves the issue in #175 (however this fix is just temporary).
2. Fixes a problem where the ebrelayer starts too quickly when the chain is only at block height 0 (just increased the delay).


utx0 (approved):
temp hot fix approved

PR #178, opened 2020-11-03
[Peggy] Some unit testing on x/ethbridge/keeper

Joe-mcgee:


x/ethbridge/keeper/keeper_test.go
```diff
@@ -0,0 +1,255 @@
...
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/stretchr/testify/require"
+	"strconv"
+	"strings"
+	"testing"
+)
+
+var (
+	cosmosReceivers, _                         = CreateTestAddrs(1)
+	amount, _                                  = strconv.ParseInt("10", 10, 64)
```
khdegraaf, line 16: This is very inelegant.  Should just do `amount int64 = 10`

x/ethbridge/keeper/keeper_test.go
```diff
@@ -0,0 +1,255 @@
...
+	ethereumSender                             = types.NewEthereumAddress("0x627306090abaB3A6e1400e9345bC60c78a8BEf57")
+	//BadValidatorAddress                        = sdk.ValAddress(CreateTestPubKeys(1)[0].Address().Bytes())
+)
+
+func TestProcessClaimLock(t *testing.T) {
+	ctx, keeper, _, _, _, validatorAddresses := CreateTestKeepers(t, 0.7, []int64{3, 3}, "")
+	validator1Pow3 := validatorAddresses[0]
+	validator2Pow3 := validatorAddresses[1]
+
+	nonce, err := strconv.Atoi("1")
```
khdegraaf, line 29: This should just be `nonce := 1`

x/ethbridge/keeper/keeper_test.go
```diff
@@ -0,0 +1,255 @@
...
+	require.Equal(t, receiverCoins.String(), "10stake")
+
+	// duplicate processSuccessClaim
+	err = keeper.ProcessSuccessfulClaim(ctx, claimString)
+	require.NoError(t, err)
+
+	//receiverCoins = bankKeeper.GetCoins(ctx, cosmosReceivers[0])
+	//require.Equal(t, "10stake", receiverCoins.String())
+	//actually is 20, could this be infinite mint?
+}
```
khdegraaf, line 190: Check to see if this behavior is correct or not.  @juniuszhou , could you comment?  If so, remove comment.  If behavior is not correct or represents work in progress, then comment should reflect that.

x/ethbridge/keeper/keeper_test.go
```diff
@@ -0,0 +1,255 @@
...
+	require.Equal(t, status.Text, oracle.SuccessStatusText)
+
+}
+
+func TestProcessClaimBurn(t *testing.T) {
+	ctx, keeper, _, _, _, validatorAddresses := CreateTestKeepers(t, 0.7, []int64{3, 3}, "")
+	validator1Pow3 := validatorAddresses[0]
+	validator2Pow3 := validatorAddresses[1]
+
+	nonce, err := strconv.Atoi("1")
```
khdegraaf, line 87: See comment above.

khdegraaf (changes requested):
A few things should be improved...

Joe-mcgee:
> A few things should be improved...

Thanks for the Feedback Kevin I'll get to making these changes

x/ethbridge/keeper/keeper_test.go
Joe-mcgee: sounds good, if curious, the intent was to mimic the input from the cli [here](https://github.com/Sifchain/sifnode/blob/eef610f3be6e361d2b6f6d3a5865d5a3d4634e51/x/ethbridge/client/cli/tx.go#L77)

x/ethbridge/keeper/keeper_test.go
juniuszhou: Yes, we can mint the coin lots of time if call the ProcessSuccessfulClaim directly. For external interface, we have the OracleID to prevent duplicated claim. The behavior is correcct.

juniuszhou (dismissed):

khdegraaf (approved):

utx0 (approved):
LGTM

PR #198, opened 2020-11-04
feature/readme-update

intl-man:
Cleanup.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -6,14 +6,18 @@
 
 1. Install golang and set GOPATH in your env. https://golang.org/doc/install
 
+2. Setup your node.
+
+If you're a new node operator, please perform the following:
+
 ```
-# clone:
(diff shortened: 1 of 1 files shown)


readme.md
```diff
@@ -62,31 +120,9 @@ f27548f03a4179b7a4dc3c8a62fcfc5f84be15ff@35.166.247.98:28004
 dd35505768be507af3c76f5a4ecdb272537e398f@35.166.247.98:28006
 ```
 
-# Verfiy
```
utx0, line 65: why you dropping this bit?

intl-man: I guess the thinking was that it won't exactly match what we had there, but perhaps I can re-word it. I'll add something else.

utx0 (approved):
LGTM

PR #199, opened 2020-11-04
[Peggy] Feature/peggy test coverage

juniuszhou:
Add two python scripts for test the peggy functionalities, also could be used to do smoking test after the deployment. Use the python to write test flow also be better to add more test cases in the future.


ElliotFriedman (approved):

scripts/README.md
```diff
@@ -11,3 +11,13 @@ Scripts for local development and testing.
 ```
 ./scripts/setup-e2e.sh
 ```
+
```
khdegraaf, line 14: This is an inaccurate or incomplete set of steps to run the tests.  I can get the python scripts to run if I run the manual steps in peggy-e2e-test.md at the beginning.  Although it took me two tries.  Rrunning peggy-basic-test.py worked, but then running peggy-e2e-test.py after failed due to 

> 2020/11/04 09:13:18 VM Exception while processing transaction: revert Not enough locked assets to complete the proposed prophecy
>calvins-mbp:testnet-contracts kevindegraaf$ ebrelayer init tcp://localhost:26657 ws://localhost:7545/ 0x30753E4A8aad7F8597332E8137

but worked if I ran it again.  Should fix ebrelayer to not crash here?  In any event, documentation and script should work easily and automatically.  Goal is for this to be turned on as a true smoke test that could be run in Github Actions check.

khdegraaf (changes requested):
This needs to work automatically from script, and it isn't quite there yet.

scripts/README.md
juniuszhou: I will add description for test environment for local node and testnet. For the relayer crash problem, I will create an issue in github and not deal with it in this pr.

khdegraaf: OK

khdegraaf (approved):

scripts/README.md
khdegraaf: Should try to fully automate in Github Actions at some point too.

PR #210, opened 2020-11-06
Hotfix fix issue 208

juniuszhou:
If the symbol of burn token not start with prefix 'c', the relayer will crash.


cmd/ebrelayer/txs/parser.go
```diff
@@ -56,6 +56,7 @@ func EthereumEventToEthBridgeClaim(valAddr sdk.ValAddress, event *types.Ethereum
 	case ethbridge.BurnText:
 		if !strings.Contains(symbol, defaultPrefix) {
 			log.Fatal("Can only relay burns of 'PEGGY' prefixed tokens")
+			return witnessClaim, errors.New("symbol of burn token must start with prefix")
```
ElliotFriedman, line 59: Maybe we should have this message be a bit more verbose. i.e. prefix 'c'

ElliotFriedman (changes requested):

ElliotFriedman (approved):

PR #213, opened 2020-11-06
Feature/test framework

khdegraaf:
Create a copy of cosmos-sdk SimApp for Sifchain, and redo CLP module to use it.  Should increase test coverage a bit, and allow full module integration testing.


x/clp/keeper/keeper.go
```diff
@@ -12,7 +12,7 @@ import (
 // Keeper of the clp store
 type Keeper struct {
 	storeKey   sdk.StoreKey
-	cdc        *codec.Codec
+	Cdc        *codec.Codec
```
Jedi2002, line 15: I think it would be a better Idea , to create getter for this isntead of exposing it directly

Jedi2002 (dismissed):
Looks good , minor comment

x/clp/keeper/keeper.go
khdegraaf: Fair point, I'll fix this.

kingpinXD (dismissed):
Lgtm

simapp/app_test.go
```diff
@@ -0,0 +1,53 @@
...
+	app.InitChain(
+		abci.RequestInitChain{
+			Validators:    []abci.ValidatorUpdate{},
+			AppStateBytes: stateBytes,
+		},
+	)
+	app.Commit()
+
+	// Making a new app object with the db, so that initchain hasn't been called
+	app2 := NewSimApp(log.NewTMLogger(log.NewSyncWriter(os.Stdout)), db, nil, true, map[int64]bool{}, 0)
```
utx0, line 34: whats the point with the second app object?

simapp/app_test.go
```diff
@@ -0,0 +1,53 @@
...
+	//_, _, err = app2.ExportAppStateAndValidators(false, []string{})
+	//require.NoError(t, err, "ExportAppStateAndValidators should not have an error")
+	require.NotNil(t, app2)
+}
+
+// ensure that black listed addresses are properly set in bank keeper
+/* func TestBlackListedAddrs(t *testing.T) {
```
utx0, line 41: What this all about?

utx0: maybe just some clean up?

utx0 (commented):
Looks like this is coming alone nicely, however @khdegraaf I might need to do a pairing session on some of this to make sure I understand all the moving parts here correctly.

simapp/app_test.go
khdegraaf: I took the cosmos-sdk version of simapp stuff, and trimmed it down and modified it to match our app.  Some of this I could just remove, but need to talk about whether it is stuff we will support long-term or not.  I know some of it is functionality we just don't support yet.  Like Slashing and Genesis.  Others may not be relevant.  Probably need to have a discussion on some of the details.

simapp/app_test.go
khdegraaf: Make sure we can do this twice.  Fuzz testing will do this over and again multiple times.

khdegraaf:
OK, I cleaned up things we discussed @utx0

utx0 (approved):
Nice one!

PR #218, opened 2020-11-07
Hotfix fix issue 206

juniuszhou:
relayer need handle if failed in calling ethereum smart contract.


cmd/ebrelayer/relayer/ethereum_test.go
```diff
@@ -1,16 +1,16 @@
...
-	tmProvider      = "Node"
-	ethProvider     = "ws://127.0.0.1:7545/"
-	contractAddress = "0x00"
-	privateKeyStr   = "ae6ae8e5ccbfb04590405997ee2d52d2b330726137b875053c36d94e974d162f"
-	// rpc              = "http://localhost:7545"
-	rpc              = ""
-	validatorMoniker = "user1"
-	chainID          = "sifchain"
-	web3Provider     = "ws://localhost:7545/"
-)
```
ElliotFriedman, line 13: Why are we commenting this out? If we comment it out because we don't need it, let's remove it from the codebase.

ElliotFriedman (changes requested):

cmd/ebrelayer/relayer/ethereum_test.go
utx0: Ive removed this whole file as its all deadcode.

utx0 (approved):

utx0 (approved):

PR #221, opened 2020-11-07
Enable solc optimizer in truffle to get contract size down below 24Kb

ElliotFriedman:


ElliotFriedman:
Because the optimizer is enabled, contract compiled bytecode size for bridgebank is under 24kb, which means we should be able to deploy to testnet.

utx0 (approved):
Nice one.

PR #222, opened 2020-11-08
Fix up swap to use correct CLP formulae

ryardley:
* Hook up backend API
* Use correct swap amount formula: `S = (x*X*Y)/(x + X)^2`
* Calculate reverse swap formula: `x = ( -2*X*S + X*Y - X*sqrt( Y*(Y - 4*S) ) ) / 2*S`
* Refactor entire swap engine to use composite pairs and double swaps over native asset ATK -> RWN -> BTK


ryardley:
@novelja I updated the backend formulae for the swap calculator and hooked it up to the backend. We probably need to provide screens in accordance with joergs designs to facilitate this.

PR #233, opened 2020-11-09
Feature/better codecov calcs

khdegraaf:
Add logic to ignore certain files we can't cover, or know are covered elsewhere in code coverage calcs.  Also, some minor code coverage improvements.


cmd/ebrelayer/relayer/cosmos.go
```diff
@@ -1,5 +1,7 @@
 package relayer
 
+// DONTCOVER
```
utx0, line 3: Whats the reasoning with all the `DONTCOVER` ?

cmd/ebrelayer/txs/.env
```diff
@@ -0,0 +1,3 @@
+# Replace sample private key for an active validator address
+# For local testing, use accounts[1] private key from truffle
```
utx0, line 2: i'd prefer we didnt commit `.env` file into the repo so we can config things locally for iso dev/testing

cover.sh
```diff
@@ -0,0 +1,9 @@
+go test -v ./... -coverprofile=coverage.txt -covermode=atomic
```
utx0, line 1: Can we put this script in `./scripts`  :)

utx0 (commented):
some comments and minor changes

cmd/ebrelayer/relayer/cosmos.go
khdegraaf: The code coverage will now notice this, and excluded the file from code coverage stats.

cover.sh
khdegraaf: sure

cmd/ebrelayer/txs/.env
khdegraaf: OK, I'll fix this.

khdegraaf: done.

cover.sh
khdegraaf: done.

cmd/ebrelayer/relayer/cosmos.go
utx0: Sorry my question was more why are we excluding these files?

khdegraaf: These are core files in ebrelayer.  Can't mock these without significant work that is better focused elsewhere.  There are a variety of other files excluded from code coverage (via settings we copied from cosmos-sdk).  Like all top-level main.go files, as well as client directories that do boiler-plate io plumbing.  That is essentially what we are doing here.

utx0: oh ok I get it now.  Thanks

utx0 (approved):
Looking good.

PR #241, opened 2020-11-10
Update Makefile for auto versioning

k0kk0k:
Version is taken from the tab https://github.com/Sifchain/sifnode/tags and the unnecessary text is cut off


utx0:
Hey @k0kk0k, thanks for the PR.  We are acutally wanting to put in place a SemVer based versioning system as aposed to using the manual release tags that we are currently using. Let me know if you are interested in helping. If so maybe do some research and let me know what you think would be a good approach and if we agreed we would be happen to offer a bounty for this task. cc @jkbrooks

utx0:
closed. we'll handle this at some point soon.

PR #244, opened 2020-11-10
Update readme.md replace symbol

Levash0v:
When configured from:
https://github.com/Sifchain/sifnode/blob/develop/readme.md
In the line: --moniker < moniker > \\
it is necessary to replace the <> symbol with ""
because otherwise it fails (hosted OS - Ubuntu 20.04)

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -126,11 +126,11 @@ sifnodecli tx staking create-validator \
     --commission-rate 0.1 \
     --amount 1000000000rowan \
     --pubkey $(sifnoded tendermint show-validator) \
-    --moniker <moniker> \
+    --moniker "moniker" \
     --chain-id monkey-bars \
     --min-self-delegation 1 \
     --gas auto \
-    --from <moniker> \
+    --from "moniker" \
     --keyring-backend file
 ```
 
--- a/readme.md
+++ b/readme.md
@@ -6,6 +6,7 @@
 
 - [Ruby 2.6.x](https://www.ruby-lang.org/en/documentation/installation)
 - [Golang](https://golang.org/doc/install)
+- [Node.js >=12.19.0](https://nodejs.org/en/)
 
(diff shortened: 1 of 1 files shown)


Levash0v:
Add to requirements
[Node.js >=12.19.0](https://nodejs.org/en/)

utx0:
Hey thanks for the PR @Levash0v , just wondering why you think nodejs is a requirement? So far this isnt a requirement to get a sifnode up and running at all? Did you get an issue of sorts at all?

readme.md
```diff
@@ -6,6 +6,7 @@
 
 - [Ruby 2.6.x](https://www.ruby-lang.org/en/documentation/installation)
 - [Golang](https://golang.org/doc/install)
+- [Node.js >=12.19.0](https://nodejs.org/en/)
```
utx0, line 9: Hey thanks for the PR @Levash0v , just wondering why you think nodejs is a requirement? So far this isnt a requirement to get a sifnode up and running at all? Did you get an issue of sorts at all?

Levash0v: When installing Starport locally, the following packages are required:
_Prerequisites:_ If you want to install Starport locally, make sure to have Golang> = 1.14 and Node.js> = 12.19.0 installed on your system._
During the initial setup, I installed two packages Noda. js and Golang (Ruby did not install it, this requirement appeared later) the node started, later installed the Ruby package

utx0: yeah ok I get it now. We only used starport to scaffold the app and arent really using it now we have the repo under development. We also cut out the frontend vue code that came with the scaffolder and are building out our own frontend foundations. ill have to get ride of the referrence to starport in the doco now and its really not a requirement any more.

utx0:
closing this as we arent really using starport anymore and nodejs isnt a dep for running a sifnoded process.

PR #245, opened 2020-11-10
Update smart contracts and tests to be upgradeable

ElliotFriedman:


musnit:
@ElliotFriedman looks ok but this is a significant PR - should get other Solidity decs to review too and then get it audited again.

smart-contracts/contracts/BridgeBank/BridgeBank.sol
```diff
@@ -16,27 +17,30 @@ import "./WhiteList.sol";
...
+    bool private _initialized;
+
+    using SafeMath for uint256;
 
     /*
-     * @dev: Constructor, sets operator
+     * @dev: Initializer, sets operator
      */
-    constructor(
+    function initialize(
```
ilyabukalov, line 29: not sure about versioning (truffle/openzepplin/solc) but it was available inheritance from "initializable" which implements this initialise functionality and adds specific to language syntax "public initialisable" in the end of constructor. Whether to use it or not - I don't have strong opinion but you could check by yourself and make sure how it works

ilyabukalov:
Overall good, just be aware that deployProxy (which now in migration script) does not automatically recognize the existence of the old proxy, so it will create each time new instance with new address allocating the new storage as well. Since we go to upgradability support this migration script should be either prevented of accidental call with additional separated script for upgrade OR some additional command line flag should be implemented in migration script for truffle which will either go to deployment or to upgrading eg. truffle migrate --blabla

ilyabukalov:
One more small comment as well. Don't forget later to commit the files from .openzepplin folder. It is used for the future upgrades (they store proxies addresses and their all implementations). If you loose this information after deployment - the future upgrades may not be possible.

ilyabukalov (approved):

smart-contracts/contracts/BridgeBank/BridgeBank.sol
ElliotFriedman: I would have used the openzeppelin implementation of initializable, however, it was implemented as an abstract contract which gave a compiler error with the current version of solc that we are using.

ilyabukalov: I believe compilation error is normal behaviour (public initialisable after constructor where "initialisable" is not real modifier). It should be compiled by openzeppelin cli. It is kind of "hack" in solc language. At least how I remember it. There is package @openzeppelin/cli which can do this in scripting as well. But I think your approach however better because it provides more direct control and flexibility.

utx0:
Hey @ElliotFriedman just looking at some of these comments and wondering do we need to create some follow up tickets at all?

smart-contracts/contracts/BridgeBank/BridgeBank.sol
ElliotFriedman: I would use openzeppelin cli, but the problem with that is this project is done in truffle so I don't feel like that would make sense. Additionally, truffle will recompile if the build folder is built by another compiler that doesn't put out the artifacts it's expecting.

ElliotFriedman:
> Hey @ElliotFriedman just looking at some of these comments and wondering do we need to create some follow up tickets at all?

@utx0 I can add some documentation to how all of the upgradeable stuff works. I don't think we need any follow up tickets.

PR #247, opened 2020-11-11
Feature/clp handlers refactor

Jedi2002:


AustinoBombino (commented):
@Jedi2002 Everything in here looks good to me so far.

One thing that Aidan suggested though was to move all of the state setting (setPool etc) back into the keepers: https://github.com/Sifchain/sifnode/pull/44#discussion_r515289289

So my interpretation of that is we would handle all of the computation the way you have it now and at the end call a simple createPool or addLiquidity etc that lives in the keeper that takes the actual computed state changes and performs them.

utx0:
@Jedi2002 this looks like it needs more work correct? If so can you update the label now?

AustinoBombino:
@Jedi2002 I think Aidan was suggesting that the functions actually be keeper functions. The parts you pulled out and put in the executors looks good but do you think they'll work as actual keeper functions?

Jedi2002:
> @Jedi2002 I think Aidan was suggesting that the functions actually be keeper functions. The parts you pulled out and put in the executors looks good but do you think they'll work as actual keeper functions?

Thats actually a good idea. Should be doable ,I am passing in the keeper as an argument anyways.

AustinoBombino (approved):
Looks great

PR #263, opened 2020-11-13
remove bin file from repo.

juniuszhou:
These bin files are not needed from deployment point of view. It is just intermediate files to generate go source code.


utx0 (approved):
Perfect.

PR #268, opened 2020-11-15
Update readme.md

c29r3:
- parse moniker via awk  
- added sync progress script


utx0:
If you would like to submit the script into a PR that would be fine, otherwise I cant accept this as its point to a non-offical script that could be changed and this sort of pratices is bad from a security point of view. For now im closing this.

PR #269, opened 2020-11-16
Feature/truffle cicd

utx0:
* Test smart contract in a PR. 
* Some minor doco updates


smart-contracts/.env.example
```diff
@@ -11,7 +11,7 @@ LOCAL_PROVIDER="http://localhost:7545"
 # Local validator count must be at least 1 and no more than 8 
 LOCAL_VALIDATOR_COUNT = 3
 # The initial power associated with each local validator
-LOCAL_INITIAL_VALIDATOR_POWERS = 50,1,1
+LOCAL_INITIAL_VALIDATOR_POWERS = 33,33,33
```
juniuszhou, line 15: some test case assume the first validator has the most of power, then the claim or peophecy can be finilized after the transaction issued from fist validator.

utx0: This is only a `.env.example` and should be made as real world as possible and not used for tests. You will see in the `smart-contracts/migrations/2_next.js` script I have set it up with default if these values arent set and these will then kick in for any test run.

smart-contracts/migrations/2_next.js
```diff
@@ -16,55 +16,48 @@ module.exports = function(deployer, network, accounts) {
...
-  // Input validation for general env variables
-  if (process.env.CONSENSUS_THRESHOLD.length === 0) {
-    return console.error(
-      "Must provide consensus threshold as environment variable."
-    );
-  }
-  consensusThreshold = process.env.CONSENSUS_THRESHOLD;
+  let consensusThreshold = (process.env.CONSENSUS_THRESHOLD === 0 ? process.env.CONSENSUS_THRESHOLD:70)
+  let localValidatorCount = Number(process.env.VALIDATOR_COUNT === 0 ? process.env.VALIDATOR_COUNT:1)
+  let initialPowers = (process.env.INITIAL_VALIDATOR_POWERS === 0 ? process.env.INITIAL_VALIDATOR_POWERS.split(","):[100])
```
utx0, line 21: Note to self. Add a fail/exit condition for a mainnet deployment if these values havent changed from the defaults.

smart-contracts/migrations/2_next.js
```diff
@@ -16,55 +16,48 @@ module.exports = function(deployer, network, accounts) {
...
-  let consensusThreshold;
-
-  // Input validation for general env variables
-  if (process.env.CONSENSUS_THRESHOLD.length === 0) {
-    return console.error(
-      "Must provide consensus threshold as environment variable."
-    );
-  }
-  consensusThreshold = process.env.CONSENSUS_THRESHOLD;
+  let consensusThreshold = (process.env.CONSENSUS_THRESHOLD === 0 ? process.env.CONSENSUS_THRESHOLD:70)
```
ElliotFriedman, line 19: I think the ternary operator here needs to be reversed. If the consensus threshold is 0, we should default to 70, otherwise use the consensus threshold from the env var.

`  let consensusThreshold = (process.env.CONSENSUS_THRESHOLD === 0 ? 70 : process.env.CONSENSUS_THRESHOLD)
`

smart-contracts/migrations/2_next.js
```diff
@@ -16,55 +16,48 @@ module.exports = function(deployer, network, accounts) {
...
+  let localValidatorCount = Number(process.env.VALIDATOR_COUNT === 0 ? process.env.VALIDATOR_COUNT:1)
```
ElliotFriedman, line 20: same comments for validator and power variable. I believe the ternary operator should be reversed so we don't have an incorrect configuration of 0 validators and 0 powers.

utx0:
closing as its too stale now. Porting the goods into a new PR.

PR #270, opened 2020-11-16
Burn Issue Fix

ElliotFriedman:


juniuszhou (approved):

utx0 (approved):
Nice work

ElliotFriedman:
This will need to be reworked. Whitelists should only include one type of token. Either cosmos native or ethereum native, but not both.

smart-contracts/contracts/BridgeBank/BridgeBank.sol
```diff
@@ -197,7 +231,7 @@ contract BridgeBank is BankStorage, CosmosBank, EthereumBank, WhiteList {
         bytes memory _recipient,
         address _token,
         uint256 _amount
-    ) public payable onlyWhiteList(_token) {
+    ) public payable onlyEthTokenWhiteList(_token) validSifAddress(_recipient) {
         string memory symbol;
```
ElliotFriedman, line 235: There is an ethereum token whitelist and a cosmos token whitelist. This keeps the concerns separate and avoids attack vectors where a user could try to burn a whitelisted token

smart-contracts/contracts/BridgeBank/CosmosWhiteList.sol
```diff
@@ -17,16 +17,16 @@ contract WhiteList is WhiteListStorage {
...
 
     /*
      * @dev: Modifier to restrict erc20 can be locked
      */
-    modifier onlyWhiteList(address _token) {
+    modifier onlyCosmosTokenWhiteList(address _token) {
         require(
-            getTokenInWhiteList(_token),
+            getCosmosTokenInWhiteList(_token),
             "Only token in whitelist can be transferred to cosmos"
```
stbeyer, line 30: The error message should probably change

stbeyer (approved):
Looks good. Left a comment about error string, since they should be different on the two whitelists, I suppose

utx0 (approved):
All look good to me mate. Merging now.

PR #274, opened 2020-11-17
[ChainOps] feature/rake

intl-man:
An idea for moving rake into the project root.


build/rakelib/helpers.rake
```diff
@@ -1,4 +1,8 @@
 # check_args checks to make sure the required args are passed in
+def cwd
```
utx0, line 2: why call it `cwd`? why `pwd`?

utx0 (dismissed):

build/rakelib/helpers.rake
utx0: yeah cool, now i get your reasoning. `cwd` works for me.

utx0 (approved):
nice work.

PR #291, opened 2020-11-19
Websocket

ilyabukalov:


utx0 (approved):
Tested and working. However it has a minor bug with the cosmso side and infact creates a new go routine for every failure case. But the eth side works and thats good enough for now. Thanks bro.

ilyabukalov:
yes I know, I investigated this as well. It is actually not bug on my understanding. I went through original go code of tendermint and documentation as well. How I see it: the web socket and subscription is maintained correctly on tendermint level (relatively "lower level' comparing to our development) and interface to interrupt (any explicit stop or implicit notification/callback) is not exposed on user level. I have tried many different scenarios of interrupting cosmos backend and non of callbacks/channel event were received. It seems indeed high level interface is stateless (in tendermint docs it is said as well with even more ambiguity: you are not allowed to close listening channel). So I assume the current behaviour is correct according to tendermint design and what you see "interrupt/reconnect" in console - it is not exposed client interface but tendermint internals which we cannot impact anyhow.

PR #296, opened 2020-11-19
[ChainOps] Updates in the block-explorer configmap

ilchebedelovski:


deploy/helm/block-explorer/templates/configmap-settings.yaml
```diff
@@ -21,45 +21,44 @@ data:
...
                     "displayNamePlural": "CHOTS",
                     "fraction": 1000000
                 }
-
             ],
             "gasPrice": 0.02,
             "coingeckoId": "sif"
         },
-        "genesisFile": "https://raw.githubusercontent.com/Sifchain/networks/feature/genesis/testnet/monkey-bars-testnet-3/genesis.json",
+        "genesisFile": "https://raw.githubusercontent.com/Sifchain/networks/feature/genesis/testnet/latest.json",
```
intl-man, line 61: For now this works. At some state we will need to parameterise this and the other options.

intl-man (approved):

deploy/helm/block-explorer/templates/configmap-settings.yaml
```diff
@@ -21,45 +21,44 @@ data:
...
             "powerReduction": 1000000,
             "coins": [
                 {
-                    "denom": "trwn",
-                    "displayName": "TRWN",
-                    "displayNamePlural": "TRWNS",
+                    "denom": "rowan",
+                    "displayName": "ROWAN",
+                    "displayNamePlural": "ROWAN",
                     "fraction": 1000000
```
intl-man, line 31: I think these factions are wrong. These look like 1e6 to me. The precision we use is 1e8. I think these all need to be changed to `100000000`.

intl-man (changes requested):

deploy/helm/block-explorer/templates/configmap-settings.yaml
utx0: Hey actually its 1e18 due to the way peggy tokens are getting minted from eth:

1 eth = 1000000000000000000 wei
1 eth = 1e18 wei

utx0: https://eth-converter.com/

intl-man (approved):

PR #304, opened 2020-11-20
Feature/integration env testrunner

Joe-mcgee:
In this PR
- scaffolds a localnet through docker with ./scripts/start-integration-env.sh
- tears down localnet with ./scripts/stop-integration-env.sh
- fills ceth from eth on validator with extra user1 account, transfers rowan there
- runs the ./scripts/peggy-basic-test-docker.py tests


utx0:
Hey @banshee and @khdegraaf should we just close this out now with the new work that is taking place?

banshee:
Yes, I think this is obsolete.

PR #306, opened 2020-11-21
Feature/clp state export

Jedi2002:
Added Import and export for Genesis state for CLP 
Added unit Tests
Refactored keeper.go into smaller files so that they are easier to manage


x/clp/genesis.go
```diff
@@ -1,21 +1,53 @@
...
 import (
+	"fmt"
 	"github.com/Sifchain/sifnode/x/clp/types"
 	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/pkg/errors"
 	abci "github.com/tendermint/tendermint/abci/types"
 )
 
-func InitGenesis(ctx sdk.Context, keeper Keeper, bankKeeper types.BankKeeper, data types.GenesisState) (res []abci.ValidatorUpdate) {
+func InitGenesis(ctx sdk.Context, keeper Keeper, data types.GenesisState) (res []abci.ValidatorUpdate) {
```
banshee, line 11: Do we normally name the result if we don't use the name and the name doesn't have any extra meaning?

banshee (approved):

x/clp/test/test_common.go
```diff
@@ -50,9 +50,14 @@ func CreateTestApp(isCheckTx bool) (*simapp.SimApp, sdk.Context) {
 }
 
 func CreateTestAppClp(isCheckTx bool) (sdk.Context, keeper.Keeper) {
+	ctx, app := GetSimApp(isCheckTx)
+	return ctx, app.ClpKeeper
+}
+
+func GetSimApp(isCheckTx bool) (sdk.Context, *simapp.SimApp) {
```
khdegraaf, line 57: This function seems redundant, and no one else is calling it.  What is its purpose?

khdegraaf (approved):
Seems overall decent from a test perspective.

x/clp/genesis.go
Jedi2002: Thats just following the coding style set by the sdk , we can definatly remove it for now . But it might be usefull when we do have more fields in the struct

x/clp/test/test_common.go
Jedi2002: I just broke down the CreateTestAppClp , into two functions ,for code clarity . Might use just the simapp for some future tests

PR #310, opened 2020-11-24
[Frontend] Feature/background events

ryardley:
* Background refresh pool balances for swap

To test:

Terminal 1: `yarn chain:sif`
Terminal 2: `yarn chain:eth`
Terminal 3: `yarn chain:migrate && yarn core:watch`
Terminal 4: `yarn app:serve`

Then launch the app http://localhost:8080

Setup a swap:
1. Connect wallet
1. Select tokens and amounts

Once up try to remove liquidity from a pool:

```bash
sifnodecli tx clp remove-liquidity \
 --from akasha \
 --sourceChain ETH \
 --symbol ETH \
 --ticker cbtk \
 --wBasis 9901 \
 --asymmetry -1
```

Notice the price output will change in accordance with the new pool amount from the swap.


ui/core/src/api/MarketService/index.ts
```diff
@@ -1,80 +1,67 @@
...
 }: MarketServiceContext) {
+  let ws: ReconnectingWebSocket;
   const sifClient = new SifUnSignedClient(sifApiUrl);
-  const poolMap = new Map<string, Pool>();
+
+  let poolHandler: PoolHandlerFn = () => {};
+
+  async function setupPoolWatcher() {
+    await new Promise((res, rej) => {
+      ws = new ReconnectingWebSocket("ws://localhost:26657/websocket");
```
59023g, line 25: This is good. Will need to tie this into a chain config file (which we'll need for keplr wallet too)

59023g: I'd put a TODO here or create a chain config like this: https://github.com/chainapsis/keplr-example/blob/master/src/main.js#L19

ui/core/src/api/MarketService/index.ts
```diff
@@ -1,80 +1,67 @@
...
+            params: {
+              query: `tm.event='Tx'`,
+            },
+          })
+        );
+        // This assumes every transaction means an update to pool values
+        // Subscribing to all pool addresses could mean having a tone of
+        // open connections to our node because there is no "OR" query
+        // syntax so have chosen to go with debouncing getPools for now.
+        ws.onmessage = async (...argoids) => {
```
59023g, line 41: Makes sense. So what you're saying is eventually we'll want some logic to parse between messages?

59023g (approved):
Looks good

ui/core/src/api/MarketService/index.ts
ryardley: Yep we definitely need to sort out config including a coin whitelist

ui/core/src/api/MarketService/index.ts
ryardley: Well ideally we would subscribe to specific pools through a single connection so we only get notified when those specific pools are updated. Cosmos query language doesn’t allow for OR queries for some reason so the only way to do this is to subscribe over multiple channels. So one per query. If we have 50 coins that’s 50 connections per browser.

PR #312, opened 2020-11-25
[Frontend] Get vercel working with lerna in the monorepo

ryardley:
# Problem 

We had a problem where because we were using a npm filesystem package running `yarn` within the `./ui` folder meant that code was cached and therefore stale which meant that development experience was super slow as you have to constantly clean out `./ui/app/node_modules/ui-core` to see changes. This was done initially to allow for vercel deployment. This PR means we don't have any devx issues and can easily deploy to vercel.

# Setup for a vercel deployment

To deploy setup a vercel deployment with the following config and run `yarn vercel` within the `./ui` folder:

image

# Done in this PR

* Added `lerna` to handle `./core` as a linked module (`ui-core`) within yarn workspaces.
* Added `vercel` as a dev dependency to `./ui` as it is deployment tooling.


59023g:
You kept Vercel as I had it set, right?

ryardley:
> You kept Vercel as I had it set, right?

I have not changed any settings in the vercel account nor anything in vercel.json

59023g (approved):

PR #317, opened 2020-11-26
[Core] Logger

ilyabukalov:


cmd/ebrelayer/siflogger/siflogger.go
```diff
@@ -0,0 +1,125 @@
...
+		option = log.AllowNone()
+	default:
+		panic("incorrect level")
+	}
+	filterNumber++
+	filter := log.NewFilter(e.logger, option)
+	e.logger = filter
+}
+
+func (e *SifLogger) SetFilterForLayer(level Level, keyvals ...interface{}) {
```
juniuszhou, line 81: Obviously the keyvals' data structure is map. Don't need judge the length is even if input is a map.

ilyabukalov: it is my question for the team as well. Do we really need a map? My opinion and idea was to have only keys without values (I call them "tags"), because when you want isolate something in logging you use only one word. I don't understand the case where it could be for example Tag=1 and somewhere else Tag=2. Logically speaking, if you have several and many values for Tag then for what reason you wanted to split your code to Tag itself? Not sure I understand fully logging concept of tendermint. BTW they have reserved key "module" hardcoded in tendermint internals, so from this point of view is possible to have "module"="1" and "module"="2". But other than that I think simple chain of tags is enough like in example in the current code (look at goroutines): 
logger = root.Tag("relayer")
logger_eth = logger.Tag("eth")
logger_cosmos = logger.Tag("cosmos")
logger_tendermint_client = logger_cosmos.Tag("client")
so we will be able to filter by "eth" OR by "cosmos" including all embedded. And additionally filter by tenderming client inside "cosmos" ONLY.

juniuszhou: If you think the value not useful, just remove it. the input as key/tag list is fine.

utx0 (changes requested):
Ok great first start. Couple of things. 
1. You need to fix the build/linting issues on this PR
2. If might actually be an idea that we split this out into its own `pkg/siflogger` package so when we are read we can use this across the whole project.
3. Also be great for this packages to have a README.md with usage doco

ElliotFriedman:
I wanted to check in where we are with this PR? It would be great to have this working on the next release of peggy.

cmd/ebrelayer/siflogger/logger.go
```diff
@@ -0,0 +1,233 @@
+package siflogger
```
Jedi2002, line 1: I think we can create a separate package for the logger interface , instead of putting it under ebrelayer ?

cmd/sifnoded/main.go
```diff
@@ -15,31 +18,109 @@ import (
...
+		err = logger.SetFilterForLayerFromConfig("module", config.LogLevel, cfg.DefaultLogLevel())
+		if err != nil {
+			return err
+		}
+		context.Logger = logger.With("module", "main")
+
+		return nil
+	}
+}
+
```
Jedi2002, line 112: Probably shoulld add this to simapp as well

cmd/ebrelayer/siflogger/logger.go
ilyabukalov: yes, @utx0 could you please advise name of package and folder location for package files.

utx0: `pkg/siflogger`

cmd/sifnoded/main.go
ilyabukalov: did I understand correctly that simapp is used only for testing purposes? Since I have not found any entry point except test_ functions.
And just in case to clarify again: this logger will be used **one per process**, not per module and not per package. As I understand by now we have only 2 processes overall to run chain: ebrelayer and sifnoded.
Is that correct?

Regarding simapp I can add it as well but it does not give any value if simapp is used only for dry testing (or please clarify if I got it wrong?).

Jedi2002: The simapp can be run seprately
You are right it is just for testing , but it might still be usefull to be able redirect output with loglevels instead.
I do realise it does not need to be part of this pr , we can add that in seperatly,

ilyabukalov:
Documentation added
https://github.com/Sifchain/sifnode/blob/b2e85ed458337453731fe35cce0d120176d05d7b/pkg/siflogger/README.md

utx0:
As this PR still needs a bit of testing and review from the test about our logging approach. For now I have put this on hold as its a nice to have and not urgent for our up and coming release.

utx0:
close due to stale PR. Can pick up the package in there another time.

PR #335, opened 2020-12-02
Token relayer cleanup

ElliotFriedman:
Fix the issue in the relayer where it capitalizes the token symbol before passing it to ethereum.

Update the documentation around this so that you don't think you need to run the script to wire a contract up to the bridgebank.


ElliotFriedman:
Comment from James was that we need to have a folder that has scripts for everyday development to make life easier and so we don't step on each other's toes when moving scripts around.

utx0 (approved):
LGTM

PR #336, opened 2020-12-02
Feature/Upgrade Module

Jedi2002:


app/app.go
```diff
@@ -186,18 +200,62 @@ func NewInitApp(
...
+		}
+		err = app.clpKeeper.SetPool(ctx, pool)
+		if err != nil {
+			return
+		}
+		clp.InitGenesis(ctx, app.clpKeeper, clp.DefaultGenesisState())
+	})
+	app.SetStoreLoader(bam.StoreLoaderWithUpgrade(&types.StoreUpgrades{
+		Added: []string{clp.ModuleName},
+	}))
```
intl-man, line 220: @Jedi2002 - this looks like some test logic, no?

intl-man (changes requested):

app/app.go
Jedi2002: THis PR is not ment to be merged , This piece of code is used to make sure the steps in the document 
https://docs.google.com/document/d/1AqWfSBK8m8XFi23aQ5j3SLhO7AcFWFKG0SWRjUDT3EE/edit , works as expected .

Adding the upgrade module in itself does not need any code, We would just need to write snippents like this when we want to do live upgrades  , the snipets would define the migration ( old binary -> new binary )  .

Lastly this snippet is pretty simple ,real upgrades would require much more complicated migrationss ,and every upgrade would be unique to the the old and new state .

intl-man: @Jedi2002 Gotcha. So far it all looks good to me. One thing I'd be keen on us doing is using the auto-download feature:

https://github.com/cosmos/cosmos-sdk/tree/master/cosmovisor#auto-download

thoughts??

Jedi2002: Yeah , I tried using that , But I was not really sure if that would be reliable  . I am thinking along the lines , that validators might have different configurations in terms of firewals ,os security etc , and the auto download feature might get blocked due to that .

I feel publishing documents which a validator can follow might be better , becuase in that case the we can pinpoint the issue easier , debugging the system security params for a validator might not be something we would get into . 

Would love to discuss more on this though @intl-man

scripts/upgrade.sh
```diff
@@ -0,0 +1,19 @@
...
+#original="voting_period\": \"172800000000000"
+#new="voting_period\": \"10"
+#while read a; do
+#    echo "${a//$original/$new}"
+#done < $DAEMON_HOME/config/genesis.json > $DAEMON_HOME/config/genesis.json.t
+#mv $DAEMON_HOME/config/genesis.json{.t,}
+
+cosmovisor start >> sifnode.log 2>&1  &
+sleep 10
+yes Y | sifnodecli tx gov submit-proposal software-upgrade testupgrade --from shadowfiend --deposit 100000000stake --upgrade-height 5 --info '{"binaries":{"linux/amd64":"https://srv-store2.gofile.io/download/K9xJtY/sifnoded.zip?checksum=sha256:8630d1e36017ca680d572926d6a4fc7fe9a24901c52f48c70523b7d44ad0cfb2"}}' --title testupgrade --description testupgrade
```
Jedi2002, line 13: @intl-man Tested this on my machine which runs linux , would be great if you could test it on some other os/arch

intl-man (dismissed):

app/app.go
```diff
@@ -224,18 +262,30 @@ func NewInitApp(
...
 		staking.NewAppModule(app.stakingKeeper, app.AccountKeeper, app.SupplyKeeper),
+		upgrade.NewAppModule(app.UpgradeKeeper),
 		oracle.NewAppModule(app.OracleKeeper),
 		ethbridge.NewAppModule(app.OracleKeeper, app.SupplyKeeper, app.AccountKeeper, app.EthBridgeKeeper, app.cdc),
 		clp.NewAppModule(app.clpKeeper, app.bankKeeper, app.SupplyKeeper),
 		faucet.NewAppModule(app.faucetKeeper, app.bankKeeper, app.SupplyKeeper),
+		gov.NewAppModule(app.govKeeper, app.AccountKeeper, app.SupplyKeeper),
 	)
 
+	//// there is nothing left over in the validator fee pool, so as to keep the
```
utx0, line 273: This is code needed?

app/app.go
```diff
@@ -216,6 +231,29 @@ func NewInitApp(
 		keys[faucet.StoreKey],
 		app.bankKeeper)
 
+	skipUpgradeHeights := make(map[int64]bool)
```
utx0, line 234: Can we add some comments around this code?

scripts/init.sh
```diff
@@ -39,5 +37,15 @@ sifnoded validate-genesis
 
 
 
+#mkdir -p $DAEMON_HOME/cosmovisor/genesis/bin
```
utx0, line 40: Can we remove this or move it to its own test script?

scripts/upgrade.sh
```diff
@@ -0,0 +1,19 @@
+#!/usr/bin/env bash
+
+
+#original="voting_period\": \"172800000000000"
```
utx0, line 4: is this still needed?

utx0 (changes requested):
Some minor clean up bits

scripts/init.sh
Jedi2002: we do need this , to create directories but I think they can be made part of the deployment scripts rather than these shell scripts

utx0 (approved):

PR #338, opened 2020-12-02
feature wait 50 blocks to process Ethereum events

juniuszhou:
For Ethereum, the block not finalized and the reorg happened sometimes. After ebrelayer received an envent from Ethereum, we put it on the event queue, process them one by one after 30 blocks produced.


cmd/ebrelayer/relayer/ethereum.go
```diff
@@ -171,26 +182,51 @@ func (sub EthereumSub) Start(completionEvent *sync.WaitGroup) {
...
+				delete(sub.EventsBuffer.Buffer, newHead.Number)
+			}
+
+			// Iterate find out older enough events
+			for key, value := range sub.EventsBuffer.Buffer {
+				sub.Logger.Info(fmt.Sprintf("New header is %d stored header is %d", newHead.Number, key))
+				tmpKey := big.NewInt(30)
+				tmpKey.Add(tmpKey, key)
+				// event happened 30 blocks ago
+				if newHead.Number.Cmp(tmpKey) > 0 {
```
ElliotFriedman, line 206: If you could add docs around this process, it would be helpful in enabling others to ramp up to speed with what peggy is doing.

ElliotFriedman (changes requested):

cmd/ebrelayer/relayer/ethereum.go
juniuszhou: Write an adr for ebrelayer to explain current ebrelayer design and improvement in the future.

docs/adr-002-ebrelayer-eth-subscribe.md
```diff
@@ -0,0 +1,42 @@
...
+
+## Status
+
+*Proposed*
+
+## Context
+In this ADR, we discuss the solution for ebrelayer how to subscribe the events from Ethereum and process these events.
+### Summary
+
+For ebrelaye, it just need subscribe the events from CosmosBridge smart contract, both LogLock event and LogBurn event, then process both events and send transaction to Sifchain. Basically, there are two problems.
```
ElliotFriedman, line 15: Overall, this is good docs, some minor changes, I do believe it is subscribed to the bridgebank, not the cosoms bridge, and `ebrelaye` should turn be `ebrelayer`

ElliotFriedman (changes requested):

ElliotFriedman:
Testing locally, ceth going across the bridge works perfectly! Will now test wiring up erowan and making sure it gets minted on the cosmos side as rowan after 30 blocks.

ElliotFriedman (dismissed):

cmd/ebrelayer/relayer/ethereum.go
```diff
@@ -171,26 +182,51 @@ func (sub EthereumSub) Start(completionEvent *sync.WaitGroup) {
 			completionEvent.Add(1)
 			go sub.Start(completionEvent)
 			return
+		case err := <-subHead.Err():
+			sub.Logger.Error(err.Error())
+			completionEvent.Add(1)
+			go sub.Start(completionEvent)
+			return
+		case newHead := <-heads:
```
ElliotFriedman, line 190: I am getting some issues running this locally. The tx's are going through instantly when I do the following:

1. run `./test/integration/start-integration-env.sh` from the root directory

2. Query the account value and take note of the ceth balance.

3. Now, burn some more eth on the ganache chain. This results in the sifnode account being given ceth instantly without the 30 block wait.

Happy to walk you through what I'm doing on my local machine so that we can debug.

ElliotFriedman (changes requested):

cmd/ebrelayer/relayer/ethereum.go
ElliotFriedman: When I checked out to commit `b826e670a5381e3d1c70a32b54da541e712a5553`, which was before I merged the code with develop, everything worked fine on your branch except for transfers that were over or equal to 1e19 as that code wasn't on your branch before you started.

cmd/ebrelayer/relayer/ethereum.go
```diff
@@ -171,26 +182,51 @@ func (sub EthereumSub) Start(completionEvent *sync.WaitGroup) {
...
+			events, ok := sub.EventsBuffer.Buffer[newHead.Number]
+			if ok && events.BlockHash != newHead.Hash() {
+				// Deal with block reorg, delete all events with old hash value
+				sub.Logger.Info(fmt.Sprintf("Block reorg found old hash is %v new hash is %v", events.BlockHash, newHead.Hash()))
+				delete(sub.EventsBuffer.Buffer, newHead.Number)
+			}
+
+			// Iterate find out older enough events
+			for key, value := range sub.EventsBuffer.Buffer {
+				sub.Logger.Info(fmt.Sprintf("New header is %d stored header is %d", newHead.Number, key))
```
ElliotFriedman, line 202: I just realized that we have a logic error here. This picks up events and stores them in the q, but it doesn't check that this event is still valid and that TX did not end up getting dis included in a block because the block was orphaned, which was the problem we are trying to solve.

Here, we will need to check on the transaction that was included in this event and make sure that it is in fact in a block that is 30 blocks ago, if it is not in a previous block, we delete it from the q, if it got included in a later block, we add it again to the queue.

ElliotFriedman (changes requested):

cmd/ebrelayer/relayer/ethereum.go
juniuszhou: I believe the we can check it via code logic before. If we get the event, which block hash not the same with the buffered one. Then we removed all events with old block hash.
			if ok && events.BlockHash != newHead.Hash() {
				// Deal with block reorg, delete all events with old hash value
				sub.Logger.Info(fmt.Sprintf("Block reorg found old hash is %v new hash is %v", events.BlockHash, newHead.Hash()))
				delete(sub.EventsBuffer.Buffer, newHead.Number)
			}

ElliotFriedman: That solution sounds good to me.

ElliotFriedman: We have to somehow query to check if this old block was valid, if it is invalid, like you said, we throw all of the old stuff out.

smart-contracts/scripts/sendTransfer.js
```diff
@@ -0,0 +1,20 @@
...
+  ******************************************/
+module.exports = async (cb) => {
+  const Web3 = require("web3");
+
+  let provider = new Web3.providers.HttpProvider(process.env.LOCAL_PROVIDER);
+
+  const web3 = new Web3(provider);
+  try {
+    const accounts = await web3.eth.getAccounts();
+    await web3.eth.sendTransaction({from: accounts[8], to: accounts[9], value: 1})
```
ElliotFriedman, line 12: Could we please loop over this 30 times in this file vs having the python file trigger this 30 times? It should make things a lot faster as truffle exec has a bunch of overhead of spinning up the compiler and checking on the artifacts and then connecting to the network.

juniuszhou: This loop over 30 times added into python script.

banshee: I think we need this the other way around - there should be a yarn command that generates N blocks.  Like @ElliotFriedman said, doing this in python is really slow, but we also want to be able to do it in shell scripts.

test/integration/start-integration-env.sh
```diff
@@ -100,10 +100,6 @@ yarn --cwd $BASEDIR/smart-contracts peggy:lock ${ADDR} 0x00000000000000000000000
 export USER1ADDR=$(cat $NETDEF | yq r - "[1].address")
 echo "export USER1ADDR=$USER1ADDR" >> $envexportfile
 
-sleep 5
-yarn --cwd $BASEDIR/smart-contracts peggy:lock ${USER1ADDR} 0x0000000000000000000000000000000000000000 $(to_wei 10)
-sleep 5
-
```
banshee, line 106: Here's where we'd want to use the generateNBlocks yarn command from a shell script

banshee (changes requested):
I think this is one case where we really do want to improve the integration tests before we merge this.  We should definitely have tests that check the balances before and after 30 blocks, for example.

test/integration/start-integration-env.sh
juniuszhou: Agree, will optimize the script.

smart-contracts/scripts/sendTransfer.js
ElliotFriedman: @juniuszhou Please add a loop inside of the javascript file to make sendTransfer.js mine 30 blocks and remove the loop in the python file. This way things will execute much faster when tests are run.

ElliotFriedman: You can use the advance block function from openzeppelin to advance blocks https://forum.openzeppelin.com/t/how-to-simulate-a-date-in-the-future-with-truffle-tests/908/3

https://docs.openzeppelin.com/test-helpers/0.5/api#advanceblock

ElliotFriedman:
I'm not seeing anything here that looks incorrect to me. I talked with the peggy althea team, they use 50 blocks before allowing transactions through.

I think we should bump up to 50 blocks waited for security reasons.

Before I run a final test against this PR, please fix all of the integration tests and then tag me so I can run a final test before signing off on it.

smart-contracts/scripts/sendTransfer.js
banshee: Can we get a method that advances N blocks, not 30?  The tests should check the balance before and after 

> You can use the advance block function from openzeppelin to advance blocks https://forum.openzeppelin.com/t/how-to-simulate-a-date-in-the-future-with-truffle-tests/908/3
> 
> https://docs.openzeppelin.com/test-helpers/0.5/api#advanceblock

Does this bignumber issue matter to us? https://github.com/OpenZeppelin/openzeppelin-test-helpers/issues/70

cmd/ebrelayer/relayer/ethereum.go
```diff
@@ -171,26 +182,47 @@ func (sub EthereumSub) Start(completionEvent *sync.WaitGroup) {
...
+			go sub.Start(completionEvent)
+			return
+		case newHead := <-heads:
+			sub.Logger.Info(fmt.Sprintf("New header %d with hash %v", newHead.Number, newHead.Hash()))
+
+			// Add new header info to buffer
+			sub.EventsBuffer.AddHeader(newHead.Number, newHead.Hash(), newHead.ParentHash)
+
+			for {
+				thirty := big.NewInt(30)
```
banshee, line 197: The tests need a way to get this constant.

banshee (changes requested):

ElliotFriedman:
Apparently we can check the status of an event that has previously mined. I would prefer our implementation uses this method by querying the event and tx hash instead of the recursive check that we currently use. https://ethereum.stackexchange.com/questions/68012/is-it-possible-to-get-duplicates-of-the-log-events-on-ethereum-when-we-have-a-fo

cmd/ebrelayer/relayer/ethereum.go
juniuszhou: will set a constant for it.

ElliotFriedman, line 197: This needs to be 50 blocks per the new requirement.

ElliotFriedman (changes requested):

ElliotFriedman (approved):

test/integration/start-integration-env.sh
banshee: We've added these tests

cmd/ebrelayer/relayer/ethereum.go
banshee: We're just hardcoding to 50 in the tests

banshee (approved):

PR #340, opened 2020-12-03
Create peggy tutorial

jkbrooks:


ElliotFriedman:
We should also note that the address they should be sending a lock tx to is this address on ropsten: `0x80E6A5D9a855D855AD6Fc9912d685b8dFdE24104`

docs/peggy tutorial
```diff
@@ -0,0 +1,47 @@
...
+```
+INFURA_PROJECT_ID="abcdebeedabcdebeed"
+ETHEREUM_PRIVATE_KEY="0x00000000000000000000000000000000000000"
+git clone https://github.com/Sifchain/sifnode
+cd sifnode/smart-contracts
+echo -e "ETHEREUM_PRIVATE_KEY=$ETHEREUM_PRIVATE_KEY\r\nINFURA_PROJECT_ID=$INFURA_PROJECT_ID" > .env
+curl -sL https://deb.nodesource.com/setup_14.x | sudo bash -
+sudo apt install nodejs
+sudo npm install -g truffle
+npm install dotenv
```
banshee, line 12: Maybe point them to code that has all the dependencies (something like test/integration/setup-linux-environment.sh), or we have to commit to updating this document every time we add dependencies.

docs/peggy tutorial
```diff
@@ -0,0 +1,47 @@
+First install dependencies:
+
+```
+INFURA_PROJECT_ID="abcdebeedabcdebeed"
+ETHEREUM_PRIVATE_KEY="0x00000000000000000000000000000000000000"
```
banshee, line 5: Seems weird to have these env vars in a section called dependencies.

ElliotFriedman: We need to do a proper write up of the docs for peggy. This write up should cover all of these things instead of putting it all into a single doc. For now, this will do.

docs/peggy tutorial
ElliotFriedman: Feel free to request changes

ElliotFriedman (approved):

PR #344, opened 2020-12-04
Clp/remove sourcechain

Jedi2002:
Removed Sourcechain
Removed Ticker 

Symbol is the only way to work with CLP right now


scripts/bug267.sh
```diff
@@ -1,16 +1,21 @@
 #!/usr/bin/env bash
 
 
+rm -rf ~/.sifnoded
```
banshee, line 4: Should this whole file be deleted?

scripts/upgrade.sh
```diff
@@ -0,0 +1,19 @@
+#!/usr/bin/env bash
```
banshee, line 1: This needs some comments or doc.  When do you run it?

scripts/bug267.sh
Jedi2002: Yeah , To clean the node .

scripts/upgrade.sh
Jedi2002: https://docs.google.com/document/d/1AqWfSBK8m8XFi23aQ5j3SLhO7AcFWFKG0SWRjUDT3EE/edit

This was more of a research task , I havent added documentation for it yet , the upgrades stuff is nowhere near being finalized . You could still use the above doc for reference

utx0:
Hey @intl-man we are having issue with this PR due to sifgen tool updates that need to happen. Can you please take a look and either help @Jedi2002 with this or if its a simple thing make the needed changes so the intergration tests can pass. Thanks mate.

intl-man:
> Hey @intl-man we are having issue with this PR due to sifgen tool updates that need to happen. Can you please take a look and either help @Jedi2002 with this or if its a simple thing make the needed changes so the intergration tests can pass. Thanks mate.

Done. @Jedi2002 this should be good now, let me know if otherwise.

app/app.go
```diff
@@ -27,6 +29,9 @@ import (
 
 	"github.com/Sifchain/sifnode/x/ethbridge"
 	"github.com/Sifchain/sifnode/x/oracle"
+
+	"github.com/cosmos/cosmos-sdk/x/upgrade"
```
utx0, line 33: Hey mate, looks like you have merged upgrade module stuff into this PR?

utx0 (changes requested):
Hey @Jedi2002 I'm not too sure what has happened here but I am seeing changes for the upgrade/gov module stuff in the `app.go` file. I therefore cant merge this as this needs to be tested separately

utx0 (approved):
Outside of my minor comment this is great work. We can deal with a clean up of the scripts whenever really.

PR #347, opened 2020-12-07
[ChainOps] Sifnode UI Helm Chart and auto build on trigger

ilchebedelovski:


intl-man:
@utx0 recommend we close as I don't think this is relevant any more.

utx0:
No longer needed.

PR #351, opened 2020-12-08
[Frontend] UI/pool interface

novelja:
## Pool and Add Liquidity UI styles

**Pool List (`/pool`)**
- [x] need a way to read the list of pools user joined

**Add Liquidity**
- [x] confirmation screens


ui/app/src/views/PoolPage.vue
```diff
@@ -2,34 +2,68 @@
...
-  color: #333;
-  background: #aaa;
-  margin-bottom: 1rem;
+<style scoped lang="scss">
+.heading {
+  display: flex;
+  align-items: center;
+
+  h3 {
+    @include title16;
```
ryardley, line 56: Have we got an explanation of these different titles etc. anywhere? Would love to see a styleguide page or our components displayed in storybook.

ui/app/vue.config.js
```diff
@@ -1,26 +1,26 @@
...
-          @import "normalize-scss";
-          @import "@/scss/reset.scss";
-          @import "@/scss/typography.scss";
-          @import "@/scss/variables.scss";
-          @import "@/scss/utilities.scss";
-          @import "@/scss/mixins.scss";
-        `
-      }
+    css: {
+        extract: {
```
ryardley, line 10: Is this reformatting required? Maybe I am wrong but it looks like prettier isn't setup correctly or wasn't setup correctly either on the project or in editor.

ui/core/src/hooks/addLiquidityCalculator.ts
```diff
@@ -105,11 +105,11 @@ export function usePoolCalculator(input: {
     const bAmount = toField.fieldAmount.value;
     if (!aAmount || !bAmount) return "";
     if (bAmount.equalTo("0")) return "";
-    return `${aAmount
+    return `<span class="number">${aAmount
```
ryardley, line 108: I don't think this markup should be here. We would be much better off returning primitives from here and formatting them in the consumer.

ryardley (changes requested):
Nice work. Main concern I have is adding formatting to the liquidity calculator as that really should have no knowledge of how it's output is used.

ryardley:
> need a way to read the list of pools user joined

There is an api that has been built for this. I can hook it up for you. cc @59023g in the meantime it might work to just mock it using the equivalent output of `api.MarketService.getPools()`

ui/app/src/components/shared/PriceCalculation.vue
```diff
@@ -6,15 +6,16 @@
...
 <script>
 export default {
-
 }
 </script>
 
 <style lang="scss">
 .price-calculation {
-  min-height: 120px;
+  height: 120px;
```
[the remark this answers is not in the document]
novelja: Yeah, I also like to keep these dynamic. But from my understanding, Joerg aims to keep the height of the whole panel at a fixed height, which further implies we should make some of these components at a fixed height also. He also uses scrollbars extensively throughout the design for cases such as this.

ui/app/src/views/PoolPage.vue
novelja: Not yet. There are a lot of different copy/type styles and with the new designs, new styles are emerging. I'll look into the best solution to organize these.

ui/core/src/hooks/addLiquidityCalculator.ts
novelja: Agree. I'll need this as two separate strings for styling purposes. Is it fine if I returned an object here with `value` and `copy`?

ui/app/src/components/shared/PriceCalculation.vue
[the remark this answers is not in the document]
ryardley: We are going to have to make this responsive though anyway. I think we should be careful here.

ui/core/src/hooks/addLiquidityCalculator.ts
ryardley: yup it should return whatever is the easiest for the UI without knowing anything about the way the UI is rendering the info

ryardley:
Ok @novelja if you can rebase develop with this you should have access to read the list of pools user joined by using the following:

```ts
await actions.clp.getLiquidityProviderPools()
```

59023g (approved):

59023g:
Tickets created for the following: 
- Display Pool Tokens Received on Confirmation Screen https://app.asana.com/0/1198643968827245/1199660948423348
- Display Liquidity Provider's Pool Share https://app.asana.com/0/1198643968827245/1199660948423343

ryardley (approved):

PR #355, opened 2020-12-08
Blockscience Rebalancing Architecture

AustinoBombino:
Architecture documents for our implementation of Blockscience's Rebalancing Policy proposal: https://hackmd.io/@shrutiappiah/By450P3ID


docs/5.Minting-Module.md
```diff
@@ -0,0 +1,693 @@
...
+// validate minter
+func ValidateMinter(minter Minter) error {
+	if minter.Inflation.IsNegative() {
+		return fmt.Errorf("mint parameter Inflation should be positive, is %s",
+			minter.Inflation.String())
+	}
+	return nil
+}
+
+// NextInflationRate returns the new inflation rate for the next hour.
```
Jedi2002, line 82: Would we use block time ,or number of blocks to keep a track of time ?

docs/5.Minting-Module.md
```diff
@@ -0,0 +1,693 @@
...
+
+// default minting module parameters
+func DefaultParams() Params {
+	return Params{
+		MintDenom:                   sdk.DefaultBondDenom,
+		InflationRateChange:         sdk.NewDecWithPrec(13, 2),
+		InflationMax:                sdk.NewDecWithPrec(20, 2),
+		InflationMin:                sdk.NewDecWithPrec(7, 2),
+		GoalBonded:                  sdk.NewDecWithPrec(67, 2),
+		BlocksPerYear:               uint64(60 * 60 * 8766 / 5), // assuming 5 second block times
```
Jedi2002, line 212: This might change , if the number of trasactions increase . Would that affect the calculation since we are assuming this to be a constant  .

docs/5.Minting-Module.md
AustinoBombino: This comment is from the cosmos mint module and isn't even accurate, I'll remove it

docs/5.Minting-Module.md
AustinoBombino: This is pulled straight from the cosmos mint module. 

I'll add to the doc to make clearer what the actual changers are vs what is standard cosmos

docs/5.Minting-Module.md
AustinoBombino: Nextinflationrate is calculated every block, not every hour

AustinoBombino:
Liquidity Pool doc is up.
There are a couple outstanding questions that have cropped up around the asymmetric add and withdrawal functions that I need to resolve with the Blockscience team, but the changes outlined in this doc will be needed regardless of the outcome of that conversation.

Jedi2002 (approved):

intl-man:
@AustinoBombino What's the status with this? This seems stale and perhaps it should be closed?

AustinoBombino:
Changes will be coming for the rebalancing module so these docs will need to be updated. Closing for now

PR #357, opened 2020-12-08
[Frontend] Feature/Add getPoolsByLiquidityProvider() and allow env configuration of websocket url

ryardley:
1. Adds the `getPoolsByLiquidityProvider()` function
[Pool: Connect List Pools API](https://app.asana.com/0/1198643968827245/1199227170208737) 

2. Adds configuration via env var for connecting the websocket endpoint
The following env vars now need to be set during deployment: 

* `VUE_APP_SIFNODE_API=https://somehost`
* `VUE_APP_SIFNODE_WS_API=ws://somehost:26657/websocket`


ui/core/src/api/ClpService/ClpService.ts
```diff
@@ -15,12 +15,13 @@ import ReconnectingWebSocket from "reconnecting-websocket";
...
   loadAssets: () => Promise<Asset[]>;
   sifApiUrl: string;
-  nativeAsset: Asset;
+  client?: SifUnSignedClient;
 };
 
 type IClpService = {
   getPools: () => Promise<Pool[]>;
   onPoolsUpdated: (handler: PoolHandlerFn) => void;
+  getPoolsByLiquidityProvider: (address: string) => Promise<Pool[]>;
```
59023g, line 24: What's a Liquidity Provider? `eth` or `sif` address?

ryardley: sif

59023g: I see

ryardley:
So I am just thinking this connects up the backend API but it might not be what @novelja needs exactly. We may want to surface the lp pools to the vue app via an action query? Would that work for you @novelja?

Something like:

```ts
onMount(() => {
  pools.value = actions.clp.getLiquidityProviderPools();
}) 
```

ryardley:
So I added the action so @novelja can hook it up to the UI as above

59023g (approved):

PR #359, opened 2020-12-09
Feature/Add admin account to genesis via cli

Jedi2002:
Added command to add clp admins to genesis.
Changed type of clp admins to accAddress from valAddress 
Changed variable names to more appropriatre terms ValidatorWhitelist -> CLPWhitelist


utx0 (commented):
Hey @Jedi2002 I'm not 100% the point of this PR. Looks like you are adding a cli command to the noded that will enable admin accounts to get added into the whitelist? Issue I see is that this will enable any node operator to run this command and add to the list right?

Jedi2002:
> Hey @Jedi2002 I'm not 100% the point of this PR. Looks like you are adding a cli command to the noded that will enable admin accounts to get added into the whitelist? Issue I see is that this will enable any node operator to run this command and add to the list right?

Not sure if I understand the question , 
So the command would the accounts to the genesis , and we would be the ones creating the genesis file ,and starting the blockchain . If a user modifyies the genesis file and tries the join the network , he would be rejected .

intl-man:
> > Hey @Jedi2002 I'm not 100% the point of this PR. Looks like you are adding a cli command to the noded that will enable admin accounts to get added into the whitelist? Issue I see is that this will enable any node operator to run this command and add to the list right?
> 
> Not sure if I understand the question ,
> So the command would add the accounts to the genesis , and we would be the ones creating the genesis file ,and starting the blockchain . If a user modifyies the genesis file and tries the join the network , he would be rejected by the consensus

@Jedi2002 - is it possible to add a test for this case? It's not so much the genesis file per-se, because the transaction will still be broadcasted to the network and accepted, no? Maybe I'm missing something here (entirely possible).

Jedi2002:
> > > Hey @Jedi2002 I'm not 100% the point of this PR. Looks like you are adding a cli command to the noded that will enable admin accounts to get added into the whitelist? Issue I see is that this will enable any node operator to run this command and add to the list right?
> > 
> > Not sure if I understand the question ,
> > So the command would add the accounts to the genesis , and we would be the ones creating the genesis file ,and starting the blockchain . If a user modifyies the genesis file and tries the join the network , he would be rejected by the consensus
> 
> @Jedi2002 - is it possible to add a test for this case? It's not so much the genesis file per-se, because the transaction will still be broadcasted to the network and accepted, no? Maybe I'm missing something here (entirely possible).

Yeah we can definitely test.
Which transaction are you talking about though?

utx0 (approved):
Great work man, thanks

PR #361, opened 2020-12-09
[Frontend] Peggy

ryardley:
We need a mock peggy interface to move forward with peggy on the frontend which we sort out the details of the interaction.

Sofar this defines a set of events:

```
TxEventEthConfCountChanged
TxEventSifConfCountChanged
TxEventEthTxInitiated
TxEventEthTxConfirmed
TxEventSifTxInitiated
TxEventSifTxConfirmed
TxEventError;
TxEventComplete;
```

These events will be pulled from various sources from a combination of listening to the ethereum blockchain, the ebrelayer and sifnode.

This will be complex so it is best to place this complexity behind an abstraction. 

peggy-structure-2

This PR is an attempt defines the interface that the TxEventEmitter uses so we can start to use it to connect to our Vue app while the other peggy stuff is still being worked on.

What are thoughts on: 
* Fluent interface? Could pass a config object with all handlers attached - doesn't seem as flexible to me
* Hard coding event types as API vs types? - Reason I was thinking to do it this way was because we get better typesafety but we could use event names as string literal union types that might mean less code but less clear intent
.

### DIFF
--- a/ui/core/package.json
+++ b/ui/core/package.json
@@ -22,6 +22,7 @@
     "bip39": "^3.0.2",
     "coingecko-api": "^1.0.10",
     "decimal.js-light": "^2.5.1",
+    "eventemitter2": "^6.4.3",
     "ganache-time-traveler": "^1.0.15",
     "jsbi": "^3.1.4",
     "reconnecting-websocket": "^4.4.0",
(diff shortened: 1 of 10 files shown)


ryardley:
I think we should possibly hold off on merging this until we see the PRD

ryardley:
So it looks like we should just get this in and keep chugging on Peggy interface without feedback/tx-corelation.

ryardley:
It makes sense perhaps to build all the peggy stuff on top of this so I have pulled this back to be a draft.

Issue #308, opened 2020-11-24
[ChainOps] Deployments via Terraform Cloud failing


intl-man:
**Description:**

When deploying a new cluster via Terraform Cloud, it fails with the following error:

```
Error: ResourceCreate: GroupVersionKind 'rbac.authorization.k8s.io/v1, Kind=ClusterRoleBinding' Get "https://E2EFAE75F73321626D55A1475F63EF0F.gr7.us-west-2.eks.amazonaws.com/api?timeout=32s": getting credentials: exec: exec: "aws-iam-authenticator": executable file not found in $PATH

  on .terraform/modules/sifchain/deploy/terraform/providers/aws/main.tf line 210, in resource "kustomization_resource" "ebs_csi_driver":
 210: resource "kustomization_resource" "ebs_csi_driver" {
```

this would appear to indicate that the Terraform runner hasn't installed `aws-iam-authenticator`.

**Version:**

```
Terraform: v0.13.5
Helm: version.BuildInfo{Version:"v3.3.4", GitCommit:"a61ce5633af99708171414353ed49547cf05013d", GitTreeState:"clean", GoVersion:"go1.14.9"}
```

**Expected Behaviour:**

Deploy a new cluster, onto AWS, without error.

**Other Information:**

N/A

utx0:
Are you sure you arent missing the `aws-iam-authenticator` cli tool?

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__This issue now has a funding of 250.0 USDC (250.0 USD @ $1.0/USDC)  attached to it.__

 * If you would like to work on this issue you can 'start work' [on the Gitcoin Issue Details page](https://gitcoin.co/issue/Sifchain/sifnode/308/100024344).
* Want to chip in? Add your own contribution [here](https://gitcoin.co/issue/Sifchain/sifnode/308/100024344).
* Questions? Checkout <a href='https://gitcoin.co/help'>Gitcoin Help</a> or the <a href='https://chat.gitcoin.co/'>Gitcoin Chat</a>
* $1,276,681.60 more funded OSS Work available on the [Gitcoin Issue Explorer](https://gitcoin.co/explorer)

gitcoinbot:
[...]

PR #362, opened 2020-12-09
Fixes Sifchain/sifnode#308

lancefrench:
Adds a custom resource to install `aws-iam-authenticator` for Terraform Cloud runs.  This is more a proof of concept hack than a robust fix.  That is, depending on the environment, this could cause issues for folks running terraform locally.  (E.g. executing from Windows, adding a linux64 binary to the PATH of someone's ARM Mac, etc.)  

We may want to leave this commented out unless Terraform Cloud state is enabled.  If that is the case, this can be moved to `cluster.tf.tpl` without issue.


jkbrooks:
@lancefrench thanks a lot, someone will take a look at this fix.  If there are any updates you can make to polish this and make it more likely to be merged in, please do make them over the next 48 hours before we decide.

utx0:
Sorry @lancefrench we havent yet had the space to do an e2e test on this and in the mean time we have dropped support for terraform cloud as this really is just a nice to have at this point. Once we get our release out the door we will get this tested and merged in due course. Thanks for the PR. We do value it.

intl-man:
Closing as this PR is now stale and we've made a lot of changes since, to our deployment pipelines. @lancefrench - thanks for the PR regardless, it is appreciated.

PR #366, opened 2020-12-10
[Frontend] Change rwn -> rowan

ryardley:
https://app.asana.com/0/1198643968827245/1199548124188574


59023g (approved):
My only comment: 
Is there a way to set this once?

ryardley:
Yeah this needs to be cleaned up I will add a task in asana to backlog.

PR #368, opened 2020-12-10
[Frontend] Chore/fix FE failing tests

ryardley:
This fixes the frontend tests broken by the removal of `--sourceChain` and `--ticker`. There will be some other ramifications deeper in the code that are possibly not being tested for but this gets our testsuite running.


ryardley:
Closing this as the fix went in via rwn - rowan change

PR #369, opened 2020-12-10
Always run UI tests for backend changes.

khdegraaf:
To verify if any backend changes break UI.

### DIFF
--- a/.github/workflows/node.yml
+++ b/.github/workflows/node.yml
@@ -1,8 +1,6 @@
 name: Node.js CI
 
-on:
-  pull_request:
-    paths: ['ui/**', '**.go']
+on: [pull_request]
 
 jobs:
   build:
--- a/.github/workflows/node.yml
+++ b/.github/workflows/node.yml
@@ -2,7 +2,7 @@ name: Node.js CI
 
 on:
   pull_request:
-    paths: ['ui/**']
+    paths: ['ui/**', '**.go']
 
 jobs:
   build:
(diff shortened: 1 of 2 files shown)


.github/workflows/node.yml
```diff
@@ -2,7 +2,7 @@ name: Node.js CI
 
 on:
   pull_request:
-    paths: ['ui/**']
+    paths: ['ui/**', '**.go']
```
banshee, line 5: The title of the PR says always run for backend changes, but this says just run very specifically for .go files.  What's the goal?

khdegraaf: Our backend changes are triggered by **.go.  So this is the same.  If this is missing important stuff, so is our backend trigger.  Do you have an instance where you think this is wrong?

banshee: Changes to go.mod?

banshee: Or maybe the Makefile that builds the go stuff could have a subtle change in the way things are built?

banshee: The action ends up calling 

```run: yarn chain:build && yarn build```

So it builds and uses just about everything.  We can't limit the file set without a lot of work and maintenance.

khdegraaf: I removed the paths dependencies so we always get this.

banshee (approved):

ryardley (approved):
🚀

PR #384, opened 2020-12-14
[Frontend] Feature/Chainconfig

ryardley:
- [x] Fix bug in interface where swap was not working because of Asset loading problem due to `symbol` vs `ticker`
- [x] Chain config loaded as [json file](https://github.com/Sifchain/sifnode/blob/3219077f2e0e21f7932b81f605f556cfd5047ec8/ui/core/config.localnet.json) 
- [x] Driven from env var `VUE_APP_DEPLOYMENT_TAG=localnet` `VUE_APP_DEPLOYMENT_TAG=testnet`
- [x] Extra docs on VSCode workspace
- [x] Fixed up annoying eslint rules in `.vue` files
- [x] Remove `getFakeTokens`
- [x] Remove TokenService
- [x] Use real ERC-20 data for testnet & mainnet config


ui/app/package.json
```diff
@@ -51,7 +51,11 @@
     "parserOptions": {
       "ecmaVersion": 2020
     },
-    "rules": {},
+    "rules": {
+      "@typescript-eslint/no-unused-vars": "off",
+      "@typescript-eslint/no-explicit-any": "off",
+      "@typescript-eslint/explicit-module-boundary-types": "off"
```
59023g, line 57: Why these changes?

ui/core/src/api/EthereumService/utils/getMetamaskProvider.ts
```diff
@@ -12,7 +12,7 @@ type WindowWithPossibleMetaMask = typeof window & {
 };
 
 // Detect mossible metamask provider from browser
-export const getWeb3Provider = async (): Promise<provider> => {
+export const getMetamaskProvider = async (): Promise<provider> => {
```
59023g, line 15: Nice

ui/core/src/api/index.ts
```diff
@@ -14,21 +11,36 @@ export type WithApi<T extends keyof Api = keyof Api> = {
...
+  const configMap: ConfigMap = {
+    localnet: parseConfig(localnetconfig as ChainConfig),
+    testnet: parseConfig(testnetconfig as ChainConfig),
+  };
+
+  return configMap[tag.toLowerCase()];
+}
+
+export function createApi(tag?: string) {
+  const context = getConfig(tag);
```
59023g, line 35: Nice

ui/core/src/api/utils/parseConfig.ts
```diff
@@ -0,0 +1,63 @@
...
+}
+
+function parseAsset(a: unknown) {
+  if (isTokenConfig(a)) {
+    return Token(a);
+  }
+  return Coin(a as CoinConfig);
+}
+
+export type ChainConfig = {
```
59023g, line 34: I could see us wanting ChainConfig separate from providers, assets, etc

ui/core/src/store/poolFinder.ts
```diff
@@ -0,0 +1,20 @@
+import { Ref, toRefs } from "@vue/reactivity";
+import { Store } from ".";
+import { Asset, Pool } from "../entities";
+
+type PoolFinderFn = (
+  s: Store
+) => (a: Asset | string, b: Asset | string) => Ref<Pool> | null;
+
+export const createPoolFinder: PoolFinderFn = (s: Store) => (
```
59023g, line 9: What does this find? A quick comment would be helpful imo

59023g (approved):
I think `poolFinder` comment would be helpful

ui/app/package.json
ryardley: Because they were delivering weird warnings in vetur about types not being used even though they were being used. This has to do with veturs poor TS support.

ui/core/src/api/utils/parseConfig.ts
ryardley: I woke up and had exactly the same thought. I was thinking I could knock that up this morning before the meeting but that we should get this in so develop isn’t blocked

ui/core/src/store/poolFinder.ts
ryardley: It finds pools by asset pair in the store. I wanted to try pull it out of the setup code but because of the reactive store putting it on the store is not practical in terms of types as methods is not reactive data.

PR #388, opened 2020-12-14
Fix env issues

ElliotFriedman:


utx0 (changes requested):
Be great if we could stop using the `.env.example` files as its not clear on how and where they end up getting used.

banshee:
> Be great if we could stop using the `.env.example` files as its not clear on how and where they end up getting used.

FYI, the integration tests have their own file test/integration/.env.ciExample.  They don't use smart-contracts/.env.example at all any more.

PR #396, opened 2020-12-15
[Peggy] Test improvments and add test for ceth => eth

banshee:


ElliotFriedman (changes requested):
There are quite a few changes in here, and I'm sure they will improve our test suite. Could you please include a readme for how to run these tests so that we have a document we can point people to when they want to get rolling with peggy running these tests?

utx0 (commented):
If I am reading this correct it looks like we have reduced the amount of test coverage here by removing `test_case_3` and `test_case_4`, whats the reasoning here?

banshee:
> Could you please include a readme

The PR has changes to test/integration/README.md, and that should cover how to run them - is there something I've left out?  (I run these often often enough that I could easily have left out something that's obvious to me but not obvious to the user...)

banshee:
> whats the reasoning here?

They've never run; the code was still there, but nothing invoked them.  I have an outstanding task to update them add them back in, but I thought that adding the additional ceth => eth was enough of a test upgrade that it was worth doing a PR now and then adding the other tests after.

utx0 (approved):

PR #398, opened 2020-12-15
added asset details to query liquidity provider response

Jedi2002:
image

### DIFF
--- a/x/clp/handler.go
+++ b/x/clp/handler.go
@@ -53,7 +53,7 @@ func handleMsgDecommissionPool(ctx sdk.Context, keeper Keeper, msg MsgDecommissi
 	// iterate over Lp list and refund them there tokens
 	// Return both RWN and EXTERNAL ASSET
 	for _, lp := range lpList {
-		withdrawNativeAsset, withdrawExternalAsset, _, _ := clpkeeper.CalcualteAllAssetsforLP(pool, lp)
+		withdrawNativeAsset, withdrawExternalAsset, _, _ := clpkeeper.CalculateAllAssetsForLP(pool, lp)
 		poolUnits = poolUnits.Sub(lp.LiquidityProviderUnits)
 		nativeAssetBalance = nativeAssetBalance.Sub(withdrawNativeAsset)
 		externalAssetBalance = externalAssetBalance.Sub(withdrawExternalAsset)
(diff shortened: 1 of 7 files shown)


x/clp/handler.go
```diff
@@ -2,6 +2,7 @@ package clp
 
 import (
 	"fmt"
+	clpkeeper "github.com/Sifchain/sifnode/x/clp/keeper"
```
AustinoBombino, line 5: I'm a bit hazy on the inclusion of clpkeeper here. Do we not already have the keeper passed as a parameter to the handler?

Jedi2002: I moved calculations.go to the keeper package .
We have alwasys used the keeper package in handler.go ,just that I had created alias for the functionss ( see alias.go) ,  alias.go and handler.go are both in the same package .

Jedi2002: I can alias these funtions as well 
keeper.CalculatePoolUnits
keeper.SwapOne

but I am not sure if it makes sense for us to expose them to be used through a direct clp import .

AustinoBombino: If the calculations.go functions are now in the keeper package then can't they be called from the keeper passed to the handler?
 ie: 
 `func handleMsgCreatePool(ctx sdk.Context, keeper Keeper, msg MsgCreatePool)`
 
 I'm not trying to suggest adding them to the alias, I'm just confused about the two instances of the keeper. Is the clpKeeper different from the keeper passed as a parameter above?

x/clp/keeper/calculations.go
```diff
@@ -180,3 +180,11 @@ func calcSwapResult(X, x, Y sdk.Uint) sdk.Uint {
 	denom := d.Mul(d)
 	return (x.Mul(X).Mul(Y)).Quo(denom)
 }
+
+func CalcualteAllAssetsforLP(pool types.Pool, lp types.LiquidityProvider) (sdk.Uint, sdk.Uint, sdk.Uint, sdk.Uint) {
```
AustinoBombino, line 184: Just a typo here: Calcualte -> Calculate

x/clp/handler.go
Jedi2002: The calculations.go were moved to the keeper package , but they not part part of the keeper struct .

AustinoBombino: Okay gotcha, that answers my question. Sorry for the confusion.

utx0 (approved):

PR #412, opened 2020-12-18
User/james/feature/wait blocks confirm

banshee:
Junius, this branch merges in some test fixes to feature/wait-blocks-confirm and the integration tests pass.


banshee:
We did this by hand instead

PR #378, opened 2020-12-13
Feature permissioned validators

juniuszhou:
create a whitelist for validators, only validator in list can send prophecy to sifnode.

PR #416, opened 2020-12-22
Add support for calling sifnoded add-genesis-validators to integration tests

banshee:
Support for testing that no balance changes should happen if you don't whitelist validators


banshee:
This doesn't enable the tests themselves since they need #378 to work.

banshee:
This also includes the improvements for doing the txhash log lookups

ElliotFriedman (approved):

PR #426, opened 2020-12-29
Feature peggy fee

juniuszhou:
Sifchain use must pay ceth to cover the cost of lock native token or burn pegged token in ethereum.


cmd/ebrelayer/relayer/cosmos.go
```diff
@@ -30,17 +38,19 @@ type CosmosSub struct {
...
 // NewCosmosSub initializes a new CosmosSub
 func NewCosmosSub(tmProvider, ethProvider string, registryContractAddress common.Address,
-	privateKey *ecdsa.PrivateKey, logger tmLog.Logger) CosmosSub {
+	privateKey *ecdsa.PrivateKey, cosmosContext *types.CosmosContext, logger tmLog.Logger) CosmosSub {
 	return CosmosSub{
 		TmProvider:              tmProvider,
 		EthProvider:             ethProvider,
 		RegistryContractAddress: registryContractAddress,
 		PrivateKey:              privateKey,
+		CosmosContext:           cosmosContext,
```
ElliotFriedman, line 53: Why do we need a cosmos context to initialize a new cosmos subscription?

ElliotFriedman (changes requested):

cmd/ebrelayer/relayer/cosmos.go
```diff
@@ -6,20 +6,28 @@ import (
...
 	tmTypes "github.com/tendermint/tendermint/types"
 
-	"github.com/Sifchain/sifnode/cmd/ebrelayer/txs"
-	"github.com/Sifchain/sifnode/cmd/ebrelayer/types"
+	ethbridge "github.com/Sifchain/sifnode/x/ethbridge/types"
+)
+
+const (
+	// ReturnCethThreshold If more than 0.1 eth remaining after transaction
+	ReturnCethThreshold = 10000000000000000
```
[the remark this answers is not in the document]
juniuszhou: Ok, I will remove the threshold.

cmd/ebrelayer/txs/relayToEthereum.go
```diff
@@ -22,21 +25,41 @@ const (
...
-		return err
+		return 0, err
+	}
+
+	// Estimate the cost for the transaction
+	if callMsg != nil {
+		// Update gas price
+		callMsg.GasPrice = auth.GasPrice
+	}
+	estimateGas, err := client.EstimateGas(context.Background(), *callMsg)
```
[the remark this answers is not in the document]
juniuszhou: The estimateGas not work well and alwasy get the evm revert. I will disable it and gas used in last call as reference.

cmd/ebrelayer/txs/relayToEthereum.go
```diff
@@ -22,21 +25,41 @@ const (
...
+
+	if err != nil {
+		return 0, err
+	}
+
+	// If ceth amount is lower than estimated gas
+	if cethAmount.Cmp(big.NewInt(int64(estimateGas))) > 0 {
+		return 0, errors.New("ceth paid not enough")
```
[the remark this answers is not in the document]
juniuszhou: Accepted.

cmd/ebrelayer/relayer/cosmos.go
juniuszhou: Cosmos context wraps up all information for send tx to cosmos, it will be used by both ethereum sub and cosmos sub.

x/ethbridge/types/msgs.go
```diff
@@ -15,11 +15,25 @@ import (
...
+// MessageTypeStringToEnum return MessageType according to input string
+func MessageTypeStringToEnum(messageType string) MessageType {
+	switch messageType {
+	case "submit":
+		return MsgSubmit
+	case "revert":
+		return MsgSubmit
+	case "return":
+		return MsgSubmit
+	}
```
ElliotFriedman, line 33: In call cases except the case that no cases are met, the same type is returned. I do not understand this.

ElliotFriedman (changes requested):

x/ethbridge/types/msgs.go
juniuszhou: It is a bug, will fix it.

cmd/ebrelayer/txs/relayToEthereum.go
```diff
@@ -22,47 +23,57 @@ const (
...
 	// Initialize client service, validator's tx auth, and target contract address
 	client, auth, target, err := initRelayConfig(provider, contractAddress, event, key)
 	if err != nil {
-		return err
+		return 0, err
+	}
+
+	// If ceth amount is lower than estimated gas
+	if cethAmount.Cmp(big.NewInt(int64(gasUsedLastTx))) < 0 {
+		return 0, errors.New("not enough ceth to cover the gas costs")
```
ElliotFriedman, line 42: Instead of looking at the gas used last time, I think we should instead look at the current gas price because ethereum gas prices vary so much depending on network demand.

I think we should scope out how much gas these 3 transactions take in total, use that as a constant, then get the current gas price of the network, and use that to calculate how much ceth needs to be burned.

ElliotFriedman (changes requested):

cmd/ebrelayer/txs/relayToEthereum.go
juniuszhou: Just use the gas used last time as reference and we can't calculate the total cost since other two transaction happened in other relayer nodes. So the tx fee ceth we charged is not precise but very close.

ElliotFriedman: We do have an accurate reading of how much gas is used for each type of transaction, so let's use that here. I don't like the idea of being imprecise as it could lead to deficits especially in the early days of peggy when there is a few transactions going across the bridge to ethereum. This would lead to incorrect amounts of gas being paid especially because ethereum gas prices swing so rapidly.

When we mint tokens on the ethereum side, it costs us 282031 gas in total for 3/4 validators to send this tx.
When we unlock or send eth on the etheruem side, it costs us 248692 gas. Based on this, we can check the tx type, and charge the gas accordingly.

x/ethbridge/keeper/keeper.go
```diff
@@ -100,11 +100,32 @@ func (k Keeper) ProcessBurn(ctx sdk.Context, cosmosSender sdk.AccAddress, amount
 	return nil
 }
 
+// ProcessUnburn processes the revert burn of bridged coins from the given sender
+func (k Keeper) ProcessUnburn(ctx sdk.Context, cosmosSender sdk.AccAddress, amount sdk.Coins) error {
```
ElliotFriedman, line 104: @juniuszhou What happens if we are running multiple relayer instances? Will this double or quadruple refund a user?

ElliotFriedman (changes requested):

x/ethbridge/client/cli/tx.go
```diff
@@ -164,9 +179,9 @@ func GetCmdBurn(cdc *codec.Codec) *cobra.Command {
 func GetCmdLock(cdc *codec.Codec) *cobra.Command {
 	//nolint:lll
 	return &cobra.Command{
-		Use:   "lock [cosmos-sender-address] [ethereum-receiver-address] [amount] [symbol] --ethereum-chain-id [ethereum-chain-id]",
+		Use:   "lock [cosmos-sender-address] [ethereum-receiver-address] [amount] [symbol] [ceth-amount] [message-type] --ethereum-chain-id [ethereum-chain-id]",
```
banshee, line 182: We need more help text around ceth-amount.  Is it the exact amount that's charged, or is it a maximum amount?

banshee (changes requested):

cmd/ebrelayer/relayer/cosmos.go
```diff
@@ -6,20 +6,28 @@ import (
...
 	tmTypes "github.com/tendermint/tendermint/types"
 
-	"github.com/Sifchain/sifnode/cmd/ebrelayer/txs"
-	"github.com/Sifchain/sifnode/cmd/ebrelayer/types"
+	ethbridge "github.com/Sifchain/sifnode/x/ethbridge/types"
+)
+
+const (
+	// RelayerNodeNumber defined for estimate total fee of Ethereum transaction
+	RelayerNodeNumber = 3
```
banshee, line 30: Why is this hardcoded to 3?

cmd/ebrelayer/relayer/cosmos.go
```diff
@@ -127,20 +139,44 @@ func getOracleClaimType(eventType string) types.Event {
...
+	cethAmount := cosmosMsg.CethAmount.BigInt()
+
+	gasUsed, err := txs.RelayProphecyClaimToEthereum(sub.EthProvider, sub.RegistryContractAddress,
+		claimType, prophecyClaim, sub.PrivateKey, cethAmount)
+
+	// If failed to send prophecy claim to Ethereum
 	if err != nil {
 		fmt.Println(err)
+
+		if err.Error() == "not enough ceth to cover the gas costs" {
```
banshee, line 165: This should be a constant - it's used in multiple places and we care about exact string equality

x/ethbridge/client/cli/tx.go
juniuszhou: it is maximum amount, I will update the usage.

cmd/ebrelayer/relayer/cosmos.go
juniuszhou: we never know how many validators online and get the exact total cost of transaction fee in Ethereum. so it is hardcoded for our testnet now.

cmd/ebrelayer/relayer/cosmos.go
juniuszhou: I will define a constant

cmd/ebrelayer/txs/relayToEthereum.go
```diff
@@ -20,49 +21,66 @@ import (
 const (
 	// GasLimit the gas limit in Gwei used for transactions sent with TransactOpts
 	GasLimit = uint64(3000000)
+	// GasForMint mint a cosmos native token in Ethereum
+	GasForMint = int64(282031)
```
ElliotFriedman, line 25: Why not just make this a big.int when creating it instead of doing it like this and casting it later?

ElliotFriedman (changes requested):

cmd/ebrelayer/txs/relayToEthereum.go
juniuszhou: grammer limit.

cmd/ebrelayer/relayer/cosmos.go
banshee: Doesn't the testnet have 4?

juniuszhou: As I know, three validators are active, one is backup.

hackingsub:
For burning ceth, when I try to put a fee higher than my balance, fail. (OK)

Otherwise, If I try to burn all Rowans using all ceth as a fee, the funds doesn't return.

x/ethbridge/types/msgs.go
```diff
@@ -12,28 +12,58 @@ import (
...
+	symbol string, cethAmount sdk.Int, messageType MessageType) MsgLock {
 	return MsgLock{
-		EthereumChainID:  ethereumChainID,
-		CosmosSender:     cosmosSender,
-		EthereumReceiver: ethereumReceiver,
-		Amount:           amount,
-		Symbol:           symbol,
+		EthereumChainID:      ethereumChainID,
+		CosmosSender:         cosmosSender,
+		CosmosSenderSequence: 0,
```
ElliotFriedman, line 55: We can't have the cosmos sender sequence set to 0. This value must be passed in as a param. I don't see anywhere that it is set later to the correct sequence.

ElliotFriedman (changes requested):

x/ethbridge/types/msgs.go
```diff
@@ -86,26 +124,35 @@ func (msg MsgLock) GetSigners() []sdk.AccAddress {
...
+	ethereumReceiver EthereumAddress, amount sdk.Int, symbol string, cethAmount sdk.Int, messageType MessageType) MsgBurn {
 	return MsgBurn{
-		EthereumChainID:  ethereumChainID,
-		CosmosSender:     cosmosSender,
-		EthereumReceiver: ethereumReceiver,
-		Amount:           amount,
-		Symbol:           symbol,
+		EthereumChainID:      ethereumChainID,
+		CosmosSender:         cosmosSender,
+		CosmosSenderSequence: 0,
```
ElliotFriedman, line 144: This cannot be set to 0. We should pass this in as the correct cosmos account sequence.

ElliotFriedman (changes requested):

ElliotFriedman:
We won't be moving forward with this approach for a dynamic fee mechanism.

PR #433, opened 2020-12-30
[Peggy] Integration tests need to use a slightly larger amount to work around javascript issues

banshee:
Rounding issues in the javascript testing libraries mean that it's problematic to just use 1 wei.


utx0 (approved):
I see no issue with this.

PR #440, opened 2021-01-04
[ChainOps] Terraform tweaks.

intl-man:
Increase the default k8s compute instance from t2.medium to t2.xlarge.

### DIFF
--- a/deploy/terraform/providers/aws/variables.tf
+++ b/deploy/terraform/providers/aws/variables.tf
@@ -46,7 +46,7 @@ variable "min_capacity" {
 }
 
 variable "instance_type" {
-  default = "t2.medium"
+  default = "t2.xlarge"
 }
 
 variable "ami_type" {


deploy/terraform/providers/aws/variables.tf
```diff
@@ -46,7 +46,7 @@ variable "min_capacity" {
 }
 
 variable "instance_type" {
-  default = "t2.medium"
+  default = "t2.xlarge"
```
utx0, line 49: why the need to go bigger bro?

intl-man: Primarily for the ETH node. Had issues with a medium where pods were getting evicted.

utx0: ok got it.

utx0 (approved):

PR #442, opened 2021-01-05
Cut down on gas by +66%

ElliotFriedman:


ElliotFriedman:
run: 1
tx:  399966
tx2:  151915
tx3:  217354
~~~~~~~~~~~~
Total: 769235

run: 2
tx:  368936
tx2:  103245
tx3:  151044
~~~~~~~~~~~~
Total: 623225

run: 2

tx:  355313
tx2:  89622
tx3:  137421
~~~~~~~~~~~~
Total: 582356


Current run is 582356

ElliotFriedman:
With this change, the only thing we store on chain is the approvals of a given prophecyID. When you pass the params for a prophecy, we use those to generate a prophecy id, and instead of storing the data about that prophecyID, we just store the prophecyID and the current amount of approval on it. The rest gets passed in. In my eyes, this is just as secure, because if you could generate input to replicate the same keccak256 hash of another tx, it would still mess with the system, albeit in a different way. However, the current system version only relies on permissioned validators so I am not concerned about preimage attacks (also keccak256 is known to be secure).


Once the prophecy is completed, then the approval is set to true and we will not allow any more sign offs of that prophecy to prevent double spends.
Also, another big architectural difference is now the CosmosBridge, Valset, and Oracle contract are combined into a single contract instead of being their own independently deployed contracts. This cut down on the amount of calls and delegate calls and definitely gave us some gas savings.

smart-contracts/migrations/2_next.js
```diff
@@ -71,70 +70,54 @@ module.exports = function(deployer, network, accounts) {
...
     await deployProxy(
       BridgeRegistry,
       [
         CosmosBridge.address,
         BridgeBank.address,
-        Oracle.address,
-        Valset.address
+        CosmosBridge.address,
+        CosmosBridge.address
       ],
```
juniuszhou, line 113: Input the same address three times, BridgeRegistry cloud be refactored.

smart-contracts/contracts/Oracle.sol
```diff
@@ -125,52 +91,21 @@ contract Oracle is OracleStorage {
...
-            if (Valset(valset).isActiveValidator(signer)) {
-                signedPower = signedPower.add(Valset(valset).getValidatorPower(signer));
-            }
-        }
+        uint256 totalPower = totalPower;
+
+        signedPower = oracleClaimValidators[_prophecyID];
 
         // Prophecy must reach total signed power % threshold in order to pass consensus
         uint256 prophecyPowerThreshold = totalPower.mul(consensusThreshold);
```
juniuszhou, line 111: we can save it in contract, don't need compute it each time.

smart-contracts/migrations/2_next.js
ElliotFriedman: Agreed.

smart-contracts/contracts/Oracle.sol
ElliotFriedman: We do have to compute if we are over the threshold when we call this function. If we store it in the contract, that is more storage and it costs more gas.

ElliotFriedman: We do store the power that has signed off in the smart contract.

ElliotFriedman: Right here: https://github.com/Sifchain/sifnode/pull/442/files#diff-d271781e15da4ca1fd42b79c262e4e6f33623521edb6a96b361dd905335011dcR80

juniuszhou (approved):

PR #451, opened 2021-01-07
[Peggy] Change integration tests to be able to work with test chains

banshee:
These are improvements to move away from using environment variables and to decouple the test system from the way the integration environment sets up the test chains.  We'd like to be able to run a subset of the tests against any chain.


test/integration/peggy-basic-test-docker.py
```diff
@@ -84,27 +90,6 @@ def lock_rowan(user, amount):
     return get_shell_output(command_line)
 
 
-def test_case_1():
```
banshee, line 87: Replaced by test_new_account.py

banshee:
Turned this into a draft, since we want to wait for the fee work to go in first.

banshee:
This was included in PR #474

PR #456, opened 2021-01-08
[Chore] Refactor frontend test config

ryardley:
As part of helping out @ElliotFriedman I discovered a few places where the test config uses hard coded values. This PR substitutes them for values calculated from built assets.


ryardley:
Looks like we were pointed to a custom npm registry mirror which was failing and timing out so I have pointed us to the default yarn mirror and removed the package-lock.json.

ElliotFriedman (approved):

PR #457, opened 2021-01-08
[Peggy] Fix unauthorized: signature verification failed bug

ryardley:
This fixes this bug: https://gist.github.com/ryardley/d81e475cad2a08d2ddfddc6bbc9496ec

So after hours of debugging and digging around the cosmos sdk - I discovered the issue was that ethbridge uses the wrong JSON encoding. 

The test is in UI land because we don't have unit testing of REST endpoints setup yet. I figure we can move it/redo it once we know what we are doing there.


ryardley:
Unfortunately there looks to be a test bed issue where I can't make node child_process exec calls to the smart-contracts test helpers. I don't really have time to debug this as I have checked and this fix solves the problem and makes my tests run locally after recompilation.

ElliotFriedman (approved):

PR #461, opened 2021-01-10
[UI] Feature/Keplr Wallet

59023g:
- [x] connect
- [x] get address
- [x] get balances
- [x] update ui with both
- [x] sign then broadcast tx
- [x] pool
- [x] panel design
- [x] swap
- [x] peg
- [x] disconnect
- [x] test


ui/core/src/actions/wallet.ts
```diff
@@ -30,6 +36,29 @@ export default ({
...
     },
+    async connectToWallet() {
+      try  {
+        // TODO type
+        const address: any = await api.SifService.connect();
+        notify({
+          type: "success", 
+          message: "Sif Account connected", 
+          detail: address 
+        });
```
ryardley, line 47: Prob don't need this notify as there is one below listening for changes in connect.

ui/core/src/api/SifService/SifService.ts
```diff
@@ -117,6 +124,37 @@ export default function createSifService({
 
     async connect() {
       // connect to Keplr
+      console.log('connect service', keplrChainConfig, keplrProvider)
+      if (!keplrProvider) {
+        throw {message: "Keplr Not Found", detail: "Check if extension enabled for this URL"}
+      }
```
ryardley, line 130: We probably don't want to do this. We can merge it and change later if you like but we should really pull all notification code out of the service layer.

ryardley (approved):
🚀

PR #470, opened 2021-01-12
Feature/Demo pt 2

ryardley:
- [x] Merge Keplr branch
- [x] Confirmation screen after add liquidity is missing token images
- [x] Need to check values in confirmation screen that they are not hard coded
- [x] Token labels must be cXYZ everywhere


59023g (approved):
Ace

PR #471, opened 2021-01-12
[Demo] feature/keplr-wallet

59023g:


ui/core/src/actions/wallet.ts
```diff
@@ -30,6 +36,29 @@ export default ({
...
     },
+    async connectToWallet() {
+      try  {
+        // TODO type
+        const address: any = await api.SifService.connect();
+        notify({
+          type: "success", 
+          message: "Sif Account connected", 
+          detail: address 
+        });
```
ryardley, line 47: notify is called for this below too - might want to get rid of this one maybe as it will result in two notifications?

ui/core/src/actions/wallet.ts
```diff
@@ -30,6 +36,29 @@ export default ({
...
+        store.wallet.sif.isConnected = true;
+        store.wallet.sif.address = address;
```
ryardley, line 49: Fair enough to update this here.

ui/core/src/api/SifService/SifService.ts
```diff
@@ -37,10 +39,15 @@ export default function createSifService({
...
   sifWsUrl,
+  keplrChainConfig,
   assets,
 }: SifServiceContext): ISifService {
   const {} = sifAddrPrefix;
-
+  const keplrProvider = getKeplrProvider()
+  // createSifService would need to be invoked at actions level for this to be cleaner imo.. maybe some sort of app initialize 
+  if(!keplrProvider) {
+    notify({type: "error", message: "Keplr not found.", detail: "Check if extension enabled for this URL."})
```
ryardley, line 49: We probably don't want to do this here. Instead you can simply create an onError handler for the service and send it up to the usecase layer and that can make the connection to notifications or notify ends up connected to everything and gets hard to test and refactor.

```ts
// In an action
api.SifService.onError((error) => {
  if(error.type === 'keplr_error') {
     notify({ ...error, type: "error"})
  }
})
```

ui/core/src/api/SifService/getKeplrProvider.ts
```diff
@@ -0,0 +1,22 @@
...
+
+// Detect mossible keplr provider from browser
+export default (): provider => {
+  const win = window as WindowWithPossibleKeplr;
+
+  if (!win) return null;
+
+  if (win.keplr && win.getOfflineSigner) {
+    // assign offline signer (they use __proto__ for some reason), so this is not as pretty as i'd like)
+    win.keplr.__proto__.getOfflineSigner = win.getOfflineSigner
```
ryardley, line 17: oh wow ha. old school!

ryardley (approved):
Only real gripe I have is the notification in the service which we will need to change at some point and I suspect we will get double notifications on connect. Apart from that looks good.

PR #475, opened 2021-01-13
[ChainOps] Upgrade docs.

intl-man:


docs/chainOps/Upgrades.md
```diff
@@ -0,0 +1,74 @@
...
+    --description <description>
+```
+
+Where:
+
+| Parameter | Description |
+|-----------|-------------|
+| `<upgrade_name>` | Name of the upgrade. This must match the upgrade name being used by the upgrade handler in the new binary. |
+| `<from>` | The moniker of the validator proposing the upgrade. |
+| `<deposit>` | The deposit/fee for proposing the upgrade. |
```
Jedi2002, line 33: I think we can mention the deposit required  is configurable thrrough the genesis

docs/chainOps/Upgrades.md
```diff
@@ -0,0 +1,74 @@
...
+    --title 'Brave new world' \
+    --description 'A special new upgrade'
+```
+
+## Voting on an upgrade proposal
+
+To vote on a proposal, simply run:
+
+```
+sifnodecli tx gov vote 1 yes \
```
Jedi2002, line 56: Also the voting period is genensis configuration , the height mentioned when creating the request using --upgrade-height should greater that the voting period . 
Else the vote might be finalised after the upgrade height has been crossed

Jedi2002 (approved):

Issue #328, opened 2020-12-01
[Peggy] Dropped Transactions


ElliotFriedman:
A current hypothetical problem that we face in peggy is that if a validator goes offline and misses a transaction, they will not listen to the events that happened in the past and submit an oracle claim based on it. We need a solution where we can replay previous transactions, listen to their events, and submit transactions based on this.

This solution should work for both ethereum and cosmos transactions in case a listener drops tx's on either side.

utx0:
Yeah the current solution only works in real time. However why would you want to reply events? Really if a validator goes off line we should have one a BFT type system so it doesnt matter and two the validator should get slashed and effert their block rewards.

ElliotFriedman:
This was brought up in a discussion with @musnit where we were discussing what happens in a scenario where multiple validators are offline.

utx0:
yeah i get that. However I think it would be easier to just make the system more BFT and be totally ok with faults and avoid complex retry logic.

musnit:
it is still valuable to have retry/catchup logic even if the system is BFT.

also, we can't launch production without handling dropped connections. making the system BFT or collateralized/insured is way more work than just adding a retry/catchup - i don't really get where you're coming from @utx0

musnit:
@ElliotFriedman @utx0 also this refers specifically just to the relayer microservice going offline/crashing. we should really stop using the word validator to refer to anything relay related

ElliotFriedman:
In my eyes, it would make more sense to just build in a mint function to cosmos where we have the admin key so that if a relayer drops and misses a TX, then we can just step in and mint tokens for that user. If we build a whole solution around retrying, it will require human input to figure out the TX that was dropped.

musnit:
> In my eyes, it would make more sense to just build in a mint function to cosmos where we have the admin key so that if a relayer drops and misses a TX, then we can just step in and mint tokens for that user. If we build a whole solution around retrying, it will require human input to figure out the TX that was dropped.
> 
> In either case, we have to do human input, having an admin key and being able to mint seems simpler as we control all of the validators submitting oracle claims on sifchain.

don't think this would work due to legal reasons.

in terms of retrying/catchup, without any human input, i suggest something like:

`i think reliability should be fairly straightforward, basically we just need the relayer to do something like (example for direction cosmos -> ethereum):
start process
query ethereum to check most recent nonce from cosmos that has been confirmed on ethereum
[...]

PR #481, opened 2021-01-13
Feature tx replay

juniuszhou:
If the ebrelayer offline for some time, we need process all missed events. the new subcommands get the events from blocks and process these events one by one.


ElliotFriedman:
We are going to go with solution #1 as defined here: https://github.com/Sifchain/sifnode/issues/328#issuecomment-754794704

We will need scripts that can be fed transactions on either chain and replay them.

We will not be using an architecture where we read from block x to block y as this could result in giving someone more than they should receive if 1 transaction in the block was correct and another was not.

ElliotFriedman:
Hi Junius, please change the structure of this replay mechanism from replaying from block x to block y to a structure where you pass the transaction id's from a certain chain, and we relay those transactions to the other chain.

ElliotFriedman (changes requested):
Hi Junius, please change the structure of this replay mechanism from replaying from block x to block y to a structure where you pass the transaction id's from a certain chain, and we relay those transactions to the other chain.

juniuszhou:
> Hi Junius, please change the structure of this replay mechanism from replaying from block x to block y to a structure where you pass the transaction id's from a certain chain, and we relay those transactions to the other chain.
That is a simple solution for replay script. If it is fine from requirement point of view, I'd like to change my replay mechanism. current solution is complicated.

cmd/ebrelayer/main.go
```diff
@@ -189,6 +194,32 @@ func initConfig(cmd *cobra.Command) error {
 	return viper.BindPFlag(flags.FlagChainID, cmd.PersistentFlags().Lookup(flags.FlagChainID))
 }
 
+func replayEthereumCmd() *cobra.Command {
+	//nolint:lll
+	replayEthereumCmd := &cobra.Command{
+		Use:     "replayEthereum [tendermintNode] [web3Provider] [bridgeRegistryContractAddress] [validatorMoniker] [validatorMnemonic] [fromBlock] [toBlock]",
+		Short:   "replay missed ethereum events",
+		Args:    cobra.ExactArgs(9),
```
banshee, line 202: You have comments for 7 arguments ([tendermintNode] [web3Provider] [bridgeRegistryContractAddress] [validatorMoniker] [validatorMnemonic] [fromBlock] [toBlock]), and you're requiring 9.

banshee (changes requested):

cmd/ebrelayer/main.go
juniuszhou: will add the two arguments sifFromBlock and sifEndBlock

banshee: Are we going to add any documentation about when to use this?  Do you have to go through the logs to figure out which blocks have been missed?  Can you run it more than once with the same start/end blocks?

cmd/ebrelayer/main.go
```diff
@@ -189,6 +194,32 @@ func initConfig(cmd *cobra.Command) error {
...
+		Example: "replayEthereum tcp://localhost:26657 ws://localhost:7545/ 0x30753E4A8aad7F8597332E813735Def5dD395028 validator mnemonic --chain-id=peggy",
+		RunE:    RunReplayEthereumCmd,
+	}
+
+	return replayEthereumCmd
+}
+
+func replayCosmosCmd() *cobra.Command {
```
banshee, line 210: Why are there two commands?  Seems like if you have to tell replayEthereumCmd the start/end blocks for both chains, then replayEthereumCmd could do everything.

juniuszhou: We can combine the two commands into single one. It is clearer if we have seperate commands. Then you know if the message missed in Ethereum or Sifchain processed by replay script.

cmd/ebrelayer/main.go
juniuszhou: please see replay-instruction.md in docs/relayer. You can run it again, will generate some duplicated transactions. No other side effect.

cmd/ebrelayer/main.go
banshee: No need to combine them.  It just wasn't clear to me why you needed starting and ending blocks for both chains.

khdegraaf:
This statistic is inaccurate, and you can ignore the failure as it isn't a required check.  I have a PR to try to fix or at least improve this (very difficult to test or verify without updating config in Github), but it hasn't been approved yet by @59023g .  If you update the branch by merging it with develop, the stats may update correctly.

ElliotFriedman (dismissed):

banshee (approved):

PR #497, opened 2021-01-15
Fix up fees

ryardley:


ryardley:
Should be good to go @ElliotFriedman

PR #513, opened 2021-01-17
Add gcc to list of prerequisites

sauravtom:


utx0 (commented):
Gcc is not a dependency

PR #517, opened 2021-01-18
Update assets.ethereum.localnet.json with ropsten token addresses

sauravtom:


ryardley:
We should make a new file for ropsten: `ui/core/src/assets.ethereum.ropsten.json` and we can work out a way to integrate it later

ryardley (changes requested):
We should make a new file for ropsten: ui/core/src/assets.ethereum.ropsten.json and we can work out a way to integrate it later

ui/core/src/assets.ethereum.localnet.json
```diff
@@ -1,51 +1,121 @@
 {
   "assets": [
     {
-      "symbol": "erowan",
+      "address": "0x2d893f36d5dd233a9a8e2f6eebb807e6da331b7f",
+      "decimals": 6,
+      "imageUrl": "https://assets.coingecko.com/coins/images/325/small/Tether-logo.png?1598003707",
+      "name": "Tether",
+      "network": "ropsten",
```
ryardley, line 8: network should be "ethereum"

ryardley (changes requested):

banshee:
This is obsolete and has been maintained in other ways since this PR

PR #520, opened 2021-01-18
Novelja/UI bugs

novelja:


ui/app/src/views/CreatePoolPage.vue
```diff
@@ -132,11 +134,18 @@ export default defineComponent({
             return "Please enter an amount";
           case PoolState.INSUFFICIENT_FUNDS:
             return "Insufficient Funds";
+          case PoolState.ZERO_ROWAN:
```
ryardley, line 137: Maybe I am missing context here but I am not sure this logic is correct.

We should make the distinction between adding and creating a pool by looking at the pools that exist and if a pool exists we should be adding liquidity or if the pool doesn't exist we are creating a pool.

ryardley (changes requested):

novelja:
I got this one wrong... smaller part of this included in https://github.com/Sifchain/sifnode/pull/552
I'll create a fresh branch for Add Liquidity/Create Pair issue...

PR #521, opened 2021-01-18
[Frontend] Refactor formulae

ryardley:
This is mainly for @iverc's benefit as a starting point when implementing swap fees.

* Move swap formula to formulae file
* Expose swap and tradeslip formulae as exports on `formulae.ts`
* Note formulae are untested


ryardley:
Might close this now @iverc has pulled these changes

PR #531, opened 2021-01-19
New formulas for swap fees calculation, Show swap fees in the UI

iverc:
- [X] New formulas for liquidity provider fee (single and double swaps)
- [X] New formulas for price slippage (single and double swaps)
- [X] Automated and manual testing


test/test-tables/sample_liquidity_fee.json
```diff
@@ -1,1072 +1,1072 @@
...
             "X": "1000000",
             "Y": "5000000",
-            "expected": "4901480.247"
+            "expected": "4901480.247034604450544064"
         },
         {
             "x": "1",
             "X": "1000000",
             "Y": "5000000",
-            "expected": "4.99999E-06"
```
ryardley, line 55: Ahh yup yikes

iverc: 🧪

ui/app/src/hooks/useCurrencyFieldState.ts
```diff
@@ -6,13 +6,17 @@ const globalState = {
   fromAmount: ref<string>("0"),
   toSymbol: ref<string | null>(null),
   toAmount: ref<string>("0"),
+  priceImpact: ref<string | null>(null),
+  providerFee: ref<string | null>(null),
```
ryardley, line 10: Is this related to the field? Or do you think we should make this some kind of global state and call it something else? I am not opposed to that just think we should probably call it something else.

ryardley (approved):
This looks pretty good to me

PR #537, opened 2021-01-20
Enable ERC-20 / cERC-20 pegging

ryardley:
This PR enables ERC-20 / cERC-20 pegging. 

* Needed to approve the bridgeBank contract to be allowed to spend the contract
* Removed stipulating fees on contracts which avoids a bug where it costs hundreds of dollars to peg to sifchain.
* Instead of subscribing to newBlockHeaders we simply poll every second - this is a little more inefficient but I am having trouble getting web3 and metamask to play nicely I suspect this is because metamask provider is not a websocket provider


ElliotFriedman (approved):
Looks good to me.

ui/core/src/api/EthbridgeService/tokenContract.ts
```diff
@@ -0,0 +1,38 @@
+import Web3 from "web3";
+import { AbiItem } from "web3-utils";
+
+// We should add other ABIs as they are required here
```
59023g, line 4: Are these in the parent dir? If so, a  `fetch()` would be most appropriate

59023g (approved):

ui/core/src/api/EthbridgeService/tokenContract.ts
ryardley: Before I was mistakenly under the impression that we needed to approve on the bridgebank token which in my head might have been a custom token with custom methods. Since then I found that we actually just need to approve on the erc-20 token associated with the burn or lock and that `approve()` was a general method found as part of the erc-20 spec. This is why I thought it made sense to make this more general and about erc-20 tokens.

I can see some download benefits to using fetch but I think the data we are talking about is pretty small so it probably doesn't matter. We could host the FakeTokenAbis as an asset say `./assets/AliceToken.json` then use fetch to pull it down at runtime. But that doesn't make sense necessarily because AliceToken is not a real token and not part of our mainnet deployment so there is a bit of a hackery type mismatch there. All the tools I see for getting ABIs from the internet are using block explorers and third party tools for published tokens so again hackery. Seems like the cleanest way is to store a general set of abi functions mapped to open zeppelin contracts and only use what we need when we need it which should keep things slim.   

I dunno seems easier to just hard code it here in a neat way we can modularize later perhaps?

59023g: 🏖

ui/core/src/api/EthbridgeService/EthbridgeService.ts
```diff
@@ -37,23 +36,32 @@ export default function createEthbridgeService({
...
 
-  async function approveBridgeBank(account: string, amount: AssetAmount) {
+  async function approveBridgeBankSpend(account: string, amount: AssetAmount) {
+    // This will popup an approval request in metamask
     const web3 = await ensureWeb3();
-    const bridgeTokenContract = await getBridgeTokenContract(
+    const tokenContract = await getTokenContract(
       web3,
-      bridgetokenContractAddress
+      (amount.asset as Token).address
```
59023g, line 44: I found an issue here. What's this `address` supposed to be?

59023g: I put the bridgeTokenContractAddress in... It asks me for permission to spend `erowan` 🤔 
Screen Shot 2021-01-21 at 2 01 12 PM

ryardley: Yep we need to do this by design. @ElliotFriedman knows more about this

ryardley: So we need to allow the bridgebank contract to spend our tokens on our behalf.

ryardley: erowan is the bridgeToken incase that was not clear.

59023g: 🤔

ElliotFriedman: Looks like that tx is authorizing the bridgebank to spend .00001 erowan on your behalf.

PR #538, opened 2021-01-20
[Peggy] Integration test DEBUG logging level is now less verbose

banshee:
Don't need to log so much when we do peg/unpeg


utx0 (approved):

banshee:
These changes were pulled in in PR #544

PR #543, opened 2021-01-21
clp test tables

AustinoBombino:
Test table samples for swaps, pool units, liquidity fees, trade slip, and standard slip.


utx0 (approved):
Nice one.

PR #552, opened 2021-01-21
Novelja/UI bugs rewind

novelja:
This is branched off of `novelja/ui-bugs` but rewinded to not include Create Pool vs Add Liquidity problem, as it was addressed in a wrong way. I'll create a clean branch for that. This addresses:

- few minor UI related issues (remove buttons, links, links open in new tab, etc...)
- bubbling up the `transactionHash` into confirm screens and linking blockexplorer transaction links.


ryardley (approved):
Looks good to me

PR #553, opened 2021-01-21
allow adding 0 rowan to preExistingPool

novelja:
This should address:
- user MUST add rowan if the pool doesn't exist (brand new pair)
- allow user to add only external asset when adding liquidity to pre existing pool.


ryardley:
So I had a quick look at this and it looks like we haven't resetting the mocks between tests when we should have:

Try adding this to the test. 

```
afterEach(() => {
  poolFinder.mockReset();
});
```

When I looked at this it appeared that this fails our tests and means that zero values in both fields delivers a share of pool percent of `100.00%` so you should probably try and fix this too.

ryardley:
I need to fix this for another ticket. I have opened this: https://github.com/Sifchain/sifnode/pull/571

ui/core/src/hooks/addLiquidityCalculator.ts
```diff
@@ -112,31 +112,52 @@ export function usePoolCalculator(input: {
...
+    if (!aAmount || aAmount.equalTo("0"))
+      return PoolState.ZERO_AMOUNTS;
+
+    if (!bAmount || bAmount.equalTo("0"))
+      // if rowan is 0 or empty ...
+      // allow if the pool exists
+      // invalid if the pool doesn't exist - ""
+      return preExistingPool.value ? PoolState.VALID_INPUT : PoolState.ZERO_AMOUNTS;
 
     return PoolState.VALID_INPUT;
```
ryardley, line 162: nit: this might read a little better if every if statement was the expression that resulted in a single state so there should only be one expression that leads to the state ZERO_AMOUNTS

59023g (approved):
🚀

PR #570, opened 2021-01-25
[ui] feature/remove-liquidity

59023g:
- Basic decimal fixes, 
- `href` fix, 
- `useRemoveLiquidityCalculator()` fix where withdraw values were not being calculated, even though was invoked by app. Basically removed `computed` from `core` file. It could use further refactor but I don't think reactivity should really be in there
- removed actions panel 'connected to connected'
- pool router fix where 'back' takes you to `singlePool` which now can load as standalone page (`/pool/${externalAsset}`)
- Some minor visual enhancements
Screen Shot 2021-01-27 at 12 44 13 AM


ui/app/src/router/index.ts
```diff
@@ -22,6 +23,11 @@ const routes: Array<RouteRecordRaw> = [
     name: "Pool",
     component: Pool,
   },
+  {
+    path: "/pool/:externalAsset",
```
ryardley, line 27: Awesome!

ryardley (approved):
Looks good to me.

PR #572, opened 2021-01-26
[Peggy] Add a test that puts multiple transactions in one ethereum block

banshee:
Make sure we can deal with more than one transaction in an ethereum block


banshee:
Closing this since it was part of a later PR that's already merged

PR #573, opened 2021-01-26
[Frontend] Display liquidity provider percent based on total liquidity provider units

ryardley:
Before the liquidity provider pool share displayed on the add liquidity page was being calculated based off the pool units generated by the add liquidity transaction only not any preexisting pool units the liquidity provider had in the pool. This makes it so the share is calculated off the total amounts the liquidity provider has contributed to the pool.


novelja (approved):
Looks fine to me!

iverc (approved):

PR #578, opened 2021-01-26
[Frontend] Fix max button

ryardley:
When Max button was built there appeared to be some issues with sending transactions that maxed out accounts and we needed to supply amounts just shy of the max. To compensate the max button filled out balance close to the max amount but just shy of it as a temporary measure. It now appears that we can remove this as our product has stabilised. 

* Boundary tests
* Added new `lessThanOrEqual` and `greaterThanOrEqual` methods on `AssetAmount` and `Fraction`


59023g (approved):
🏖️

PR #580, opened 2021-01-26
🚀 [Frontend] Swap UI improvements 🚀

iverc:


ui/core/src/api/SifService/SifService.ts
```diff
@@ -298,7 +298,10 @@ export default function createSifService({
 
         return txHash;
       } catch (err) {
-        console.error(err);
+        if (err.toString().includes("Request rejected")) {
```
59023g, line 301: Nice

59023g (approved):
Awesome

ui/core/src/actions/clp/index.ts
```diff
@@ -77,7 +77,7 @@ export default ({
   }
 
   const actions = {
-    async swap(sentAmount: AssetAmount, receivedAsset: Asset) {
+    async swap(sentAmount: AssetAmount, receivedAsset: AssetAmount) {
```
ryardley, line 80: Would it be better to call this `minReceivedAmount:AssetAmount` then it is clearer as to the intention here?

ui/app/src/components/slippagePanel/Index.vue
```diff
@@ -0,0 +1,153 @@
+
```
ryardley, line 1: been trying to throw reusable ui components that don't depend on shared state in `shared` which we can later rename `ui`

ui/app/src/hooks/useCurrencyFieldState.ts
```diff
@@ -8,6 +8,7 @@ const globalState = {
   toAmount: ref<string>("0"),
   priceImpact: ref<string | null>(null),
   providerFee: ref<string | null>(null),
+  slippage: ref<string>("0"),
```
ryardley, line 11: this is starting to hold more than the state for the currency field and probably could be renamed?

ui/core/src/api/ClpService/ClpService.ts
```diff
@@ -116,16 +117,17 @@ export default function createClpService({
...
+          symbol: params.receivedAsset.asset.symbol,
+          ticker: params.receivedAsset.asset.symbol,
         },
         sent_amount: params.sentAmount.toBaseUnits().toString(),
         sent_asset: {
           source_chain: params.sentAmount.asset.network as string,
           symbol: params.sentAmount.asset.symbol,
           ticker: params.sentAmount.asset.symbol,
         },
+        min_receiving_amount: test.toString(),
```
ryardley, line 130: test failure?

ui/core/src/api/ClpService/ClpService.ts
```diff
@@ -24,7 +25,7 @@ type IClpService = {
   getPoolSymbolsByLiquidityProvider: (address: string) => Promise<string[]>;
   swap: (params: {
     fromAddress: string;
-    receivedAsset: Asset;
+    receivedAsset: AssetAmount;
```
ryardley, line 28: Would it be better to call this `minReceivedAmount` to signal intent?

ryardley (changes requested):
Looks good. Aside from debugging test failure. My main concern is swap signature change.

ui/app/src/hooks/useCurrencyFieldState.ts
ryardley: So I removed some of this state in another PR as if we need to sync state I am thinking it might be better to do so with the store as our view model?

ui/app/src/components/confirmationDialog/AnimatedConfirmation.vue
```diff
@@ -17,11 +17,37 @@
               <p class="sub">Confirm this transaction in your wallet</p>
             </div>
           </transition>
+          <transition name="swipe">
```
59023g, line 20: cc @novelja

ui/app/src/components/slippagePanel/Index.vue
```diff
@@ -0,0 +1,155 @@
...
+  emits: [
+    "update:slippage",
+  ],
+  setup(props, context) {
+    const localSlippage = computed({
+      get: () => props.slippage,
+      set: (amount) => handleUpdateSlippage(amount),
+    });
+    function handleUpdateSlippage(amount: string) {
+      context.emit("update:slippage", amount);
```
59023g, line 59: Nice.

nit: I noticed in the screen recording that the display decimal point used a comma, not period to separate

59023g (dismissed):
💯

ui/app/src/components/slippagePanel/Index.vue
iverc: @59023g yes, same for other inputs. Should we use a dot for all of them?

59023g: I would. But not a dealbreaker

59023g (approved):

ryardley (approved):
Nice work! 🚀

ui/app/src/views/SwapPage.vue
```diff
@@ -96,16 +100,27 @@ export default defineComponent({
...
+        } else {
+          transactionState.value = "confirmed";
+        }
+        clearAmounts();
+      } catch (e) {
+        if (e.toString().includes("Request rejected")) {
+          transactionState.value = "rejected";
+        } else {
+          transactionState.value = "failed";
+        }
```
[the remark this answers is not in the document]
iverc: @ryardley Good point. I think we throw errors in both cases, and show different UI for each case (rejected is when the user rejects transaction in wallet popup, either intentionally or not, and transaction was not even sent), and failed is when we send transaction, but it does not go through. For some elements we just don't care which one happenned, so use something like `if rejected or failed => show error svg`

59023g (dismissed):

59023g (dismissed):

ryardley (dismissed):

ryardley (approved):

ryardley (approved):

PR #581, opened 2021-01-26
Bug: Buttons - Look Disabled, but Can Still Click

novelja:
Disabled button on `ActionsPanel` cpt level, if `!nextStepAllowed` so this should now work for all scenarios where actions panel is used


59023g (dismissed):

ui/app/src/components/actionsPanel/ActionsPanel.vue
```diff
@@ -22,7 +22,9 @@ export default defineComponent({
   setup(_, { emit }) {
     return {
       handleNextStepClicked() {
-        emit("nextstepclick");
+        if (_.nextStepAllowed) {
```
ryardley, line 25: This is not the best way to solve this. I think it has to do with the SifButton component. What is happening is that we are not delegating event handlers to the button and link components within SifButton.

ryardley (changes requested):

ui/app/src/components/actionsPanel/ActionsPanel.vue
novelja: I think this is now addressed by removing the wrapping `span` so that click handler is attached directly to the `button` element.

ryardley (approved):
😎

PR #583, opened 2021-01-27
Feature/more UI coverage2

khdegraaf:
test coverage


khdegraaf:
The code coverage isn't really down.  The configuration changes should maybe help with more consistent stats and not losing UI coverage.  Can't tell until it is merged in though for sure.

.github/workflows/go.yml
```diff
@@ -29,7 +29,7 @@ jobs:
         go get -v -t -d ./...
 
     - name: Build
-      run: go build -v ./...
```
ryardley, line 32: Not sure I understand what this change does and why make install is better.

.github/workflows/node.yml
```diff
@@ -27,9 +27,6 @@ jobs:
         run: |
           go get -v -t -d ./...
 
-      - name: Build
-        run: go build -v ./...
```
ryardley, line 31: Does this reuse the preexisting binary that is cached?

ryardley (commented):
I would like to review this but I am not sure exactly what it is doing so hopefully you don't mind I ask some potentially silly or basic questions?

.github/workflows/node.yml
khdegraaf: No, yarn chain:build ends up do a `make build`, so this is unnecessary and broken.

.github/workflows/go.yml
khdegraaf: the go build command literally doesn't work.  Make install has lots of detailed options and works.

.github/workflows/node.yml
ryardley: Ahh that's right it does too.

59023g (changes requested):
I'd like to hold off on this until at least after the release. There have been blocking inconsistencies with the existing code coverage action, so I'm hesitant to change it more. Especially now, and without further dialogue amongst stakeholders (cc @ryardley).

khdegraaf:
This change might fix those inconsistencies.  I don't believe code coverage can block merges either.  It is an optional check, not a required one.  I suspect this failed temporarily, but you were blocked due to something else.

codecov.yml
```diff
@@ -47,6 +47,8 @@ ignore:
   - "*.md"
   - "*.rst"
   - "**/test_common.go"
+  - "cmd/ebrelayer"
```
utx0, line 50: whats having these files in here do? They arent ignored I hope?

khdegraaf:
Merged elsewhere.

PR #585, opened 2021-01-27
Fix/min event

Jedi2002:


Jedi2002:
@iverc  , I am closing this, let me know if you think otherwise

iverc:
@Jedi2002 yes, the correct version is merged already. Here it returns both response and error, but it looks like if the error is not `nil` it disregards the response even if it is present. Please correct me if I'm wrong, I think that it also does not send events if the error is present, does it?

PR #591, opened 2021-01-27
Bug/guard pools in swap

ryardley:
This PR guards against the following error:

image

* Fixes issue in `CompositePool` where `contains()` method was unusable as was returning incorrect result.
* Guards against pools trying to swap assets they don't contain until eventual consistency is reached.
* Caches contains lookup in computed value
* Make it so swapCalculator simply returns `priceImpact` and `providerFee` to avoid duplication and as they are really only computed properties based on inputs - this will be easier to refactor to a render function later too.
* Add store accessor to debug store in browser

image


59023g (dismissed):
Looks good. I propose we wait a few hours until @iverc puts his work in before this is merged. Just to keep his runway clear. Looks like there's some overlap

ryardley:
> Looks good. I propose we wait a few hours until @iverc puts his work in before this is merged. Just to keep his runway clear. Looks like there's some overlap

Been thinking the same myself

59023g (approved):
Looks good. Curious of @iverc thoughts here

ui/app/src/hooks/useCore.ts
```diff
@@ -17,6 +17,13 @@ const store = createStore();
 const actions = createActions({ store, api });
 const poolFinder = createPoolFinder(store);
 
+// expose store on the window so it is easy to inspect
+Object.defineProperty(window, "store", {
```
iverc, line 21: Should we remove this part when done testing?

ryardley: I think it is useful to be able to inspect this generally and will help with isolating some of these UI bugs. There should be no sensitive information here and you cannot write to the store.

59023g: I agree. Let's keep for now

ryardley: By the way I tried exposing it for Vue devtools on the App component but that chugged big time and this seemed to be the better way to go.

iverc (approved):

59023g (approved):
💯

PR #592, opened 2021-01-27
Implement min received swap amount

iverc:
Respond with correct data if the received amount is less than min threshold


Jedi2002 (approved):

59023g:
I think this code is now included in https://github.com/Sifchain/sifnode/pull/580. Is that correct?

iverc:
@Jedi2002 right, but let's merge this one first so BE and FE changes are separated

PR #596, opened 2021-01-28
[UI] Show more appropriate fields when viewing a pool

thomasdavis:


ui/app/src/components/poolList/SinglePool.vue
```diff
@@ -54,33 +61,44 @@ export default defineComponent({
       const t = toToken.value;
       return t.imageUrl;
     });
-    const toValue = computed(() => accountPool?.value?.pool.amounts[0].toFixed(DECIMALS));
+
+    const toTotalValue = computed(() => accountPool?.value?.pool.amounts[0].toFixed(DECIMALS));
```
ryardley, line 65: I am wondering with this number formatting does it make sense to throw this all in a function now or wait until later?

thomasdavis: From our call this morning, I think someone wanted to rework the concept of to/from in most places, maybe it is better left until that refactor?

ryardley (approved):
Not sure if you wanna follow this with a cleanup PR with sorting out decimals in a function?

ui/app/src/components/poolList/SinglePool.vue
ryardley: Fair enough

ryardley: This file can be simplified pretty majorly

59023g:
👍

PR #601, opened 2021-01-28
Swap UI hotfixes

iverc:


ryardley (approved):
🎸

ui/app/src/components/shared/DetailsPanel.vue
```diff
@@ -91,8 +91,18 @@ export default defineComponent({
       }
     }
 
+    function showMinReceived(minReceived: string) {
+      const minimumReceived = parseFloat(minReceived);
+      if (minimumReceived < 100000) {
```
59023g, line 96: We'll want to abstract this logic out so any display value can be input here

59023g (approved):

ui/app/src/components/shared/DetailsPanel.vue
ryardley: Yep totally agree I think we need a set of core utility functions that handle this.

PR #603, opened 2021-01-28
[UI] Confirmation screens for peggy

novelja:
A try to make confirmation flow a bit more dry using props, slots, and adding transaction-specific content through wrapper component.

- abstracts the screen flow, and shared content/props (transition, icons, state)
- removes unnecessary passing of props 3 levels deep
- makes it easy to create a wrapper component that passes the required slots/props for each type of transaction


ui/app/src/views/CreatePoolPage.vue
```diff
@@ -108,16 +109,33 @@ export default defineComponent({
...
+      } catch (e) {
+        // TODO: Implement better error checks and status updates. -> check swap also
+        if (e.toString().includes("Request rejected")) {
+          transactionState.value = "rejected";
+          transactionStateMsg.value = "Please confirm the transaction in your wallet.";
+        } else {
+          transactionState.value = "failed";
+          transactionStateMsg.value = "Oops... Something went wrong. Please try again!";
+        }
+      }
```
ryardley, line 138: Yeah so having this ceremony everywhere is what I was worried about from the perspective of my comment on @iverc 's PR 🤔 @novelja do you feel like simplifying this from the vue layer pushing some of this marshalling at least down to core and applying it to swap too?

ryardley: I think we could do better with:

```ts
let tx = await actions.clp.addLiquidity(
  tokenBFieldAmount.value,
  tokenAFieldAmount.value
);
transactionHash.value = tx.hash
transactionState.value = tx.state
```

ryardley: Or there is also a more event emitter driven approach where instead of awaiting 
```ts
const tx = actions.clp.addLiquidity(
  tokenBFieldAmount.value,
  tokenAFieldAmount.value
);

tx.onTxHash((txHash:string) => {
  transactionHash.value  = txHash;
});

tx.onStateChanged((state:Tx.State) => {
  transactionState.value  = txHash
}); 

tx.onError((err) => { 
  tellUserSomethingRealBadHasHappened(err);
});
```

ryardley: These can be combined in a "promievent" pattern where you can do both so I think if we support option1 now we can add optinon2 later and that wont involve changing code

ui/app/src/components/confirmationDialog/CreatePoolConfirmationDialog.vue
```diff
@@ -0,0 +1,221 @@
...
+    const fromAsset = useAssetItem(fromSymbol);
+    const fromLabel = fromAsset.label;
+    const fromImage = computed(() => fromAsset.token.value ? fromAsset.token.value.imageUrl : '');
+    const fromAmount = computed(() => props.fromAmount);
+
+    const toSymbol = computed(() => props.toToken);
+    const toAsset = useAssetItem(toSymbol);
+    const toLabel = toAsset.label;
+    const toImage = computed(() => toAsset.token.value ? toAsset.token.value.imageUrl : '');
+    const toAmount = computed(() => props.toAmount);
```
iverc, line 45: @ryardley I might be wrong, but it looks like excessive reactivity here. we have reactive props outside and then also use models and events. I think we should debug how many times we reload components with one action to make sure that all is good

ryardley: yes props are already reactive.

ui/core/src/actions/clp/index.ts
```diff
@@ -94,6 +94,11 @@ export default ({
       nativeAssetAmount: AssetAmount,
       externalAssetAmount: AssetAmount
     ) {
+      const response = {
+        hash: <string>"",
+        state: <string>"",
+        stateMsg: <string>""
```
ryardley, line 100: There are a few issues with this - perhaps I can make some suggestions

ui/app/src/views/CreatePoolPage.vue
ryardley: What I understand you might be saying would be we write to the store from within core and then have our view read from that instead of track the transaction in the handler. That makes sense to me.

ui/app/src/components/shared/ConfirmationModal.vue
```diff
@@ -0,0 +1,89 @@
+<template>
+  <ModalView
```
ryardley, line 2: This looks awesome @novelja

ui/app/src/components/shared/ConfirmationModal.vue
```diff
@@ -0,0 +1,89 @@
...
+  | "selecting"
+  | "confirming"
+  | "signing"
+  | "confirmed"
+  | "rejected"
+  | "failed";
+
+export default defineComponent({
+  props: { 
+    isOpen: { type: Boolean, default: false },
```
ryardley, line 65: could this be redundant as it could be based on the state? Maybe you can pull that state logic in?

novelja: That's a good point! I'll correct it now!

ryardley (dismissed):
We still need to change over the swap confirmation but can do that later. This is a good cleanup.

ryardley (approved):

PR #614, opened 2021-01-30
[ui] Liquidity testing / insufficient funds bug / calculator fixed!! 🎉🚀

ryardley:
* Adding more unit tests shows calculator is correctly accounting for pre-existing liquidity
* Fixed a bug with calculation of state by refactoring so that there is only a single escape route per output state
* Updated replacer in window console to show values instead of internal representation.
* Fix bug in calculator where liquidity provider was not being passed in


ui/core/src/hooks/addLiquidityCalculator.ts
```diff
@@ -233,25 +233,26 @@ export function usePoolCalculator(input: {
...
+    const bAmountIsZeroOrFalsy = !bAmount || bAmount.equalTo("0");
+    const noPreexistingPool = !preExistingPool.value;
+    if ((noPreexistingPool && bAmountIsZeroOrFalsy) || aAmountIsZeroOrFalsy)
+      return PoolState.ZERO_AMOUNTS;
 
+    // Insufficient Funds
     if (fromBalanceOverdrawn.value || toBalanceOverdrawn.value)
       return PoolState.INSUFFICIENT_FUNDS;
 
+    // Valid yay!
```
59023g, line 255: :chart:

59023g (dismissed):
:win:

59023g (approved):
🥇

PR #618, opened 2021-01-30
Bug/metamask connection weirdity

ryardley:
Metamask connection weirdness

So we have been having an issue where the new metamask provider is actually connecting on load but we are not fetching accounts. This fixes that for now. We should work through our usecases for how to deal with wallets and refactor to a better Blockchain/Provider pattern to deal with this but this fixes this in the meantime.

We might want to autoconnect to wallets in the future for example.

This also reloads the page on network change as recommended by metamask docs

---

Secondly this fixes the console store debugger which I accidentally broke.

Dumps a JSON string to the console:

```js
store.dump()
```

Dumps JSON to another browser tab for easy copying:

```js
store.dumpTab()
```

### DIFF
--- a/ui/app/src/hooks/useCore.ts
+++ b/ui/app/src/hooks/useCore.ts
@@ -22,13 +22,14 @@ Object.defineProperty(window, "store", {
   get: function() {
     return JSON.parse(
       JSON.stringify(store, function replacer(_, value) {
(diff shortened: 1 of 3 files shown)


59023g (approved):
Good idea to reload

PR #619, opened 2021-01-31
[ui] bug/rm-liquidity + banner update

59023g:
- Smarter 'back' button from 'add liquidity' page with or without external asset selected
- Asymmetry fix where chain didn't like value we sent (somewhat a hack until `Fraction` can take floating point)
- More clear copy on what withdrawal means: "You Should Receive"
- Banner updated copy and header has white background again


ui/core/src/actions/clp/index.ts
```diff
@@ -139,6 +139,11 @@ export default ({
       wBasisPoints: string,
       asymmetry: string
     ) {
+
+      // Note: Chain needs -1 - 1, 
+      // UI needs to do -10000 - 10000 bc current Fraction floating point limitation
+      asymmetry =  (+asymmetry/10000).toString()
```
ryardley, line 145: Wow I didn't know you could do that! Is that JS type coercion?

59023g: haha yea

ryardley (approved):

PR #620, opened 2021-01-31
[Frontend] Prevent bug in swap for small numbers

ryardley:
* Move `minimumReceived` into swapCalculator.
* Use `AssetAmount` and base units so we arent sending unparsable strings to backend
* Add comments and some testing
* Remove some cruft from entities


59023g (approved):
:cool:

PR #621, opened 2021-01-31
[UI] Various bugfixes

iverc:


ui/app/src/components/slippagePanel/Index.vue
```diff
@@ -55,18 +55,31 @@ export default defineComponent({
       get: () => props.slippage,
       set: (amount) => handleUpdateSlippage(amount),
     });
+
     function handleUpdateSlippage(amount: string) {
       context.emit("update:slippage", amount);
     }
+
+    function isNumeric(s: any) {
+      return (s - 0) == s && (''+s).trim().length > 0;
```
ryardley, line 64: would `!isNaN(s)` work or are there other possibilities you are thinking about?

ui/app/src/components/slippagePanel/Index.vue
```diff
@@ -55,18 +55,31 @@ export default defineComponent({
...
+      try {
+        if (!isNumeric(localSlippage.value)) {
+          localSlippage.value = "1.0";
+        } else if (parseFloat(localSlippage.value) < 0) {
+          localSlippage.value = localSlippage.value.substring(1);
+        } else if (parseFloat(localSlippage.value) >= 99.0) {
+          localSlippage.value = "99.0"
+        } else if (!(localSlippage.value.startsWith(".") && localSlippage.value.length === 2)) {
+          localSlippage.value = parseFloat(localSlippage.value).toFixed(1);
+        }
```
ryardley, line 77: Yeah fair enough. We should probably set up a number backed field with a display value at some point we can use to avoid having to do this sort of stuff.

ryardley (approved):
Nice work getting subscriptions working!

ui/core/src/api/EthbridgeService/parseTxFailure.ts
```diff
@@ -4,7 +4,7 @@ import { ErrorCode, getErrorMessage } from "../../entities/Errors";
 // TODO: Should this go in a shared ethereum client mimicking sifchain?
 export function parseTxFailure({
   hash = "",
-  log,
+  log = "",
```
ryardley, line 7: 👍

ryardley (changes requested):
Ahh yes I remember this - you have run in to the issue I had around the subscriptions. They are not testing very well against the test ethereum provider. You might have to fix that up and use a different type of provider to get the tests to work.

ui/app/src/components/slippagePanel/Index.vue
iverc: @ryardley `!isNan()` does not work in some cases. This might look like overkill since we also validate the swap button, but better to be more cautions I think

ui/core/src/api/EthereumService/EthereumService.ts
```diff
@@ -65,33 +63,30 @@ export class EthereumService implements IWalletService {
...
-          provider.on("connect", () => {
-            this.state.connected = true;
-          });
-          provider.on("disconnect", () => {
-            this.state.connected = false;
-          });
-        }
+    if (Web3.givenProvider) {
+      this.provider = Web3.givenProvider;
+      this.web3 = new Web3(Web3.givenProvider || "ws://localhost:7545");
```
ryardley, line 68: This default looks like it will never happen?

ryardley (approved):
Simplified things to always have a web3 instance. Looks good.

ryardley (changes requested):
Actually this brings back the bug with metamask asking to connect when trying to disconnect

ui/core/src/api/EthereumService/EthereumService.ts
```diff
@@ -129,82 +127,47 @@ export class EthereumService implements IWalletService {
...
-  removeWeb3Subscription() {
-    this.blockSubscription?.unsubscribe();
-  }
-
   async disconnect() {
     if (isMetaMaskProvider(this.provider)) {
-      this.provider.disconnect &&
-        this.provider.disconnect(0, "Website disconnected wallet");
+      if (this.provider.request) {
+        await this.provider.request({
```
ryardley, line 148: This probably needs to be reverted. I found this in the docs too but it doesn't work and asks metamask to reconnect instead of disconnect.

iverc: @ryardley yes, there is no way to disconnect. But this approach pops up metamask forcing address selection, so you can actually connect a different address. Otherwise, the disconnect button makes it look like you disconnect, but the wallet is actually keeps being connected...

ryardley (approved):

PR #627, opened 2021-02-01
Fix remove liquidity

iverc:


ui/core/src/actions/clp/index.ts
```diff
@@ -139,10 +139,6 @@ export default ({
       asymmetry: string
     ) {
 
-      // Note: Chain needs -1 - 1, 
-      // UI needs to do -10000 - 10000 bc current Fraction floating point limitation
-      asymmetry =  (+asymmetry/10000).toString()
```
59023g, line 144: @iverc This reintroduces error where server request fails because sent out of bounds value:
Screen Shot 2021-02-01 at 2 17 58 PM

59023g (changes requested):

ui/core/src/actions/clp/index.ts
iverc: @59023g 
This is remove liquidity test on the backend:
It passes with 1000 assymetry
Screenshot 2021-02-01 at 23 40 56

iverc: @59023g Here is the check for asymmetry on the backend:
`	if m.Asymmetry.GTE(sdk.NewInt(10000)) || m.Asymmetry.LTE(sdk.NewInt(-10000)) {
		return sdkerrors.Wrap(ErrInvalidAsymmetry, m.Asymmetry.String())
	}
	return nil
`

The error message should be updated 
`
	ErrInvalidAsymmetry                = sdkerrors.Register(ModuleName, 9, "Asymmetry has to be 1,-1 or 0")

`

iverc: So yes, if you send < -10000 or > 10000, it will fail. Fixed error message on the backend side

ryardley (dismissed):

x/clp/types/msgs.go
```diff
@@ -127,7 +127,7 @@ func (m MsgRemoveLiquidity) ValidateBasic() error {
 	if !(m.WBasisPoints.IsPositive()) || m.WBasisPoints.GT(sdk.NewInt(MaxWbasis)) {
 		return sdkerrors.Wrap(ErrInvalidWBasis, m.WBasisPoints.String())
 	}
-	if m.Asymmetry.GTE(sdk.NewInt(10000)) || m.Asymmetry.LTE(sdk.NewInt(-10000)) {
+	if m.Asymmetry.GT(sdk.NewInt(10000)) || m.Asymmetry.LT(sdk.NewInt(-10000)) {
```
59023g, line 130: To confirm: will error if greater than 10,000 or less than -10000 ?

iverc: Yes, but will accept == 10000 and == -10000

59023g (approved):

PR #631, opened 2021-02-02
[ui] added a basic tooltip system

thomasdavis:


ui/app/src/components/shared/Tooltip.vue
```diff
@@ -0,0 +1,68 @@
...
+export default defineComponent({
+  props: {
+    message: {
+      type: String,
+    },
+  },
+
+  data: function() {
+  	return {
+  		opened: false
```
iverc, line 13: Dit auto-format it like that? could be as 1 line or have 2 spaces I think. @ryardley @59023g In general, I think we should have a default code style across the team so we don't restyle each other's code on every commit. It really makes commits look much bigger than they are and also makes it harder to review and easier to miss bugs.
What do you think about using prettier with default settings?

iverc (approved):

ui/app/src/components/shared/Tooltip.vue
thomasdavis: I am a super prettier fanboy, I would love if we can use it. 

I think they do a thorough job of explaining the benefits for the team -> https://prettier.io/docs/en/why-prettier.html

59023g: I know @ryardley had a linter in mind. Prettier sounds good to me I'll ping the group in Slack. We will find consensus

ui/app/src/components/shared/Tooltip.vue
```diff
@@ -0,0 +1,68 @@
...
+    };
+  },
+  methods: {
+    close() {
+      console.log("close tool tips");
```
59023g, line 18: would  remove this

59023g (approved):
looks good!

PR #635, opened 2021-02-02
Proof of concept block explorer functionality for Peggy

musnit:
Note: depends on https://github.com/Sifchain/sifnode/pull/636 being merged first

see README

includes a nodejs backend that syncs/indexes all ethereum and cosmos transactions and a frontend that shows a table with transactions, matching up corresponding locks+ relayed claims.

eg:
image


ElliotFriedman:
This would definitely be a nice to have. I don't think it makes sense to productionize and deploy.

PR #636, opened 2021-02-02
add events to ethbridge

musnit:


utx0 (approved):
Fixed matching event tests too.

utx0:
old  branch as merge target

PR #639, opened 2021-02-03
Fix disconnect button

iverc:


59023g (approved):
tested working in vercel deployment

PR #640, opened 2021-02-03
[Peggy] Queueless Architecture

ElliotFriedman:


cmd/ebrelayer/relayer/ethereum.go
```diff
@@ -176,54 +179,75 @@ func (sub EthereumSub) Start(completionEvent *sync.WaitGroup) {
...
 		case err := <-subHead.Err():
-			sub.Logger.Error(err.Error())
+			sub.Logger.Error("subHead failed: " , err.Error())
 			completionEvent.Add(1)
 			go sub.Start(completionEvent)
 			return
 		case newHead := <-heads:
 			sub.Logger.Info(fmt.Sprintf("New header %d with hash %v", newHead.Number, newHead.Hash()))
+			
+			trailingBlocks := big.NewInt(4)
```
musnit, line 197: '4' here should be moved up to a constant/config somewhere

cmd/ebrelayer/relayer/ethereum.go
```diff
@@ -176,54 +179,75 @@ func (sub EthereumSub) Start(completionEvent *sync.WaitGroup) {
 	}
 	defer subHead.Unsubscribe()
 
+	lastProcessedBlock := big.NewInt(0)
```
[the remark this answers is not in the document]
banshee: We also talked about writing out some local data on where the relayer is, so it could just read that on restart if it were available.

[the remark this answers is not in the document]
musnit: @banshee on-chain query would be more robust, and remove the risk of local becoming out of sync with onchain for whatever reason

cmd/ebrelayer/main.go
```diff
@@ -43,6 +43,9 @@ func init() {
 	// config.SetBech32PrefixForConsensusNode(sdk.Bech32PrefixConsAddr, sdk.Bech32PrefixConsPub)
 	// config.Seal()
 
+	// TODO Only for debugging. Remove me laterz
```
banshee, line 46: Needs to be removed

cmd/ebrelayer/relayer/cosmos.go
```diff
@@ -6,6 +6,7 @@ import (
 	"context"
 	"crypto/ecdsa"
 	"fmt"
+	"github.com/prometheus/common/log"
```
banshee, line 9: Are we picking up the right logger here?  Most other files just import "log"

cmd/ebrelayer/relayer/ethereum.go
```diff
@@ -161,9 +160,12 @@ func (sub EthereumSub) Start(completionEvent *sync.WaitGroup) {
 	signal.Notify(quit, syscall.SIGINT, syscall.SIGTERM)
 	defer close(quit)
 
-	// Start BridgeBank subscription, prepare contract ABI and LockLog event signature
-	bridgeBankAddress, subBridgeBank := sub.startContractEventSub(logs, client, txs.BridgeBank)
-	defer subBridgeBank.Unsubscribe()
+	// get the bridgebank address from the registry contract
+	bridgeBankAddress, err := txs.GetAddressFromBridgeRegistry(client, sub.RegistryContractAddress, txs.BridgeBank)
+	if err != nil {
```
banshee, line 165: Shouldn't this be fatal?  If we can't get the bridgeBankAddress, we'll still try to use it later on.

cmd/ebrelayer/relayer/ethereum.go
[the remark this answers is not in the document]
banshee: Totally agree that on-chain is better than local.  Local was just something we discussed and didn't implement, and has lots of obvious drawbacks (what if you loose the machine, for example).

cmd/ebrelayer/txs/relayToCosmos.go
```diff
@@ -3,50 +3,96 @@ package txs
...
+	txBytes, err := txBldr.BuildAndSign(moniker, password, messages)
 	if err != nil {
+		log.Printf("ebrelayer RelayToCosmos error 1 is %s\n", err.Error())
 		return err
 	}
-
+	
+	log.Println("built tx, now broadcasting")
 	// Broadcast to a Tendermint node
-	res, err := cliCtx.BroadcastTxSync(txBytes)
```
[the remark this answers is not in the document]
ElliotFriedman: I moved Async on suggestion from Tanmay. Even if we move back to sync, we should still wait because I think this broadcast will always resolve before the tx's are included in a block.

cmd/ebrelayer/relayer/cosmos.go
ElliotFriedman: That is a good question for @utx0

cmd/ebrelayer/relayer/ethereum.go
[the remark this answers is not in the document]
ElliotFriedman: I agree that this would be a really great feature to add in the future. For now, I think that instead of starting at block 0, it should just start at the block number that is passed in from an environment variable. This way, if we need to restart, it doesn't give people extra money on the sifchain side that they shouldn't have.

cmd/ebrelayer/relayer/ethereum.go
juniuszhou: Agree with James, we can exit here if not get the bridge bank address.

ElliotFriedman: Added fatal if bridgebank address is not found.

Peggy-Debugging.md
```diff
@@ -0,0 +1,31 @@
+# Current Peggy Issues
```
banshee, line 1: Should this file be removed?  (not a blocker)

cmd/ebrelayer/relayer/ethereum.go
banshee: FYI, that's just marking the log message as fatal, it doesn't exit.

setup.sh
```diff
@@ -0,0 +1,4 @@
+make clean install
```
banshee, line 1: Did you mean to check this file in at the root?

x/ethbridge/handler.go
```diff
@@ -40,16 +42,31 @@ func NewHandler(
...
 	ctx sdk.Context, cdc *codec.Codec, bridgeKeeper Keeper, msg MsgCreateEthBridgeClaim,
 ) (*sdk.Result, error) {
+	var mutex = &sync.RWMutex{}
+	mutex.Lock()
+	defer mutex.Unlock()
+	// require tx lock to be false
+	// set mutex lock to true
+	fmt.Println("Sifnode handleMsgCreateEthBridgeClaim 43")
+
+	spew.Dump("---------------- handleMsgCreateEthBridgeClaim ----------------------")
```
banshee, line 52: Did you want to leave these spews in?

intl-man: @banshee Nope, lets get rid of 'em.

Peggy-Debugging.md
ElliotFriedman: Probably

cmd/ebrelayer/relayer/cosmos.go
utx0: fixed

musnit:
@ElliotFriedman what happens with the cosmos -> eth direction on restart? Where does it start?

ElliotFriedman:
> @ElliotFriedman what happens with the cosmos -> eth direction on restart? Where does it start?

Cosmos -> eth direction only needs one confirmation, so it just starts listening on the current block when the relayer restarts.

musnit:
> > @ElliotFriedman what happens with the cosmos -> eth direction on restart? Where does it start?
> 
> Cosmos -> eth direction only needs one confirmation, so it just starts listening on the current block when the relayer restarts.

ah we'll want similar resume-ability for it too then (eg: if it crashes and misses a few blocks)

utx0 (commented):
> > > @ElliotFriedman what happens with the cosmos -> eth direction on restart? Where does it start?
> > 
> > 
> > Cosmos -> eth direction only needs one confirmation, so it just starts listening on the current block when the relayer restarts.
> 
> ah we'll want similar resume-ability for it too then (eg: if it crashes and misses a few blocks)

We are going to handle this case in a new PR so we can move forward with testing the main issues in our sandpit env. Ticket https://app.asana.com/0/1199697235740010/1199898505617719/f

utx0 (approved):

PR #641, opened 2021-02-03
🚧 [ui] Upgrade Vue

ryardley:
So this is causing a few weirdities.


ryardley:
What appears to be happening is that `@click.stop` has stopped working.

intl-man:
@ryardley Closing as the UI has since been decoupled.

PR #642, opened 2021-02-04
[ui] added a page title to all routes. added a before hook to router to trigger manual GA page views

thomasdavis:


thomasdavis:
This PR will currently run on enviroments. I think we should add the check to only run on production later. 

1) because production is not live yet
2) we can keep triggering activity locally to make sure google analytics is reporting everything correctly.

iverc (approved):
🆗

PR #645, opened 2021-02-04
Peggy limit block

juniuszhou:
based on the queueless peggy solution, just limit the blocks processed each time.


juniuszhou:
will create other pr based on develop branch. it is difficult to review.

PR #646, opened 2021-02-04
Feature/approve state for erc20

ryardley:
- [x] Create dynamic way to deliver confirmation dialogs
- [x] Manage page state with a clearer mental model using a state machine
- [x] Reduce boilerplate in PegAssetsPage via JSX 
- [x] Add approve panel and state 
- [x] Tidy up rubbish and apply confirmation dialogs to other pages

Please test this thoroughly as you review it I may have missed some details. 

* Swap has not been converted to the new confirmation screens but should be flagged for conversion
* In `PegAssetPage` confirmation dialog states are dynamic but are tied to pageState so if you try and define a state that doesn't exist your editor will complain.

---

### On JSX

Ok from the slack message we had there is a chance this might be controversial but I thought it might make more sense to refactor a single view to JSX in order to demonstrate the advantages of doing so. I am finding I am way more productive on complex components because of the advantages that JSX templates provide.

I think this is going to be a really huge productivity boost for the team which is why I am pushing this strongly. Looks at a few of the reasons why using JSX for complex state components should be written using JSX:

JSX mode
* In VSCode click on an element it takes you directly to where it was defined you don't need to search for variable names.
* Copy and paste template code to a new component you will get state errors you need to fix up directly in your browser so much easier to refactor
* You can take chunks of interface and save them off to variables
* If you make properties required you will know in your editor if you have forgotten a prop
* There is no boilerplate
* JSX components do not allow scoped styles - seems like a bug but is in fact a feature. We do not want to include styles in JSX templates. We should save styling for template components.
* If you screw up you know in your editor without looking at the output
* It is pretty clear after using it that the Composition API was pretty much designed for a JSX workflow.

Vue Templates
* You have to have boilerplate to use refs and state often repeating references to a property 3 times
* You have to search in the file for all the references to that element that exist
* To save off chunks of interface in Vue templates you have to create a separate component
* If you make properties required you have to compile the code to discover you have made a mistake and then find it in the code
* Vue Templates make it too easy to mix complex state driven components with display components
* Sometimes vue templates has a nicer syntax eg ternaries


ryardley:
This PR is probably too far out of date to be merged but the approach is good to transplant so leaving this open and branch here for now.

intl-man:
@ryardley Closing as the UI has since been decoupled.

PR #652, opened 2021-02-05
set limit for processed block.

juniuszhou:
set limit for processed block as 2, to avoid pull too much events from ethereum one time.


cmd/ebrelayer/relayer/ethereum.go
```diff
@@ -271,9 +280,9 @@ func (sub EthereumSub) Start(completionEvent *sync.WaitGroup) {
 				}
 				time.Sleep(transactionInterval)
 			}
-			// save the current ending block to the lastprocessed block to ensure we keep reading blocks sequentially
-			lastProcessedBlock = endingBlock
-			err = db.Put([]byte(ethLevelDBKey), lastProcessedBlock.Bytes(), nil)
+
+			// save the current ending block + 1 to the lastprocessed block to ensure we keep reading blocks sequentially and don't repeat blocks
+			err = db.Put([]byte(ethLevelDBKey), endingBlock.Add(endingBlock, big.NewInt(1)).Bytes(), nil)
```
banshee, line 285: can we move endingBlock.Add(endingBlock, big.NewInt(1)) onto its own line?  It's easy to miss that the call to db.Put is also changing the value of endingBlock.

cmd/ebrelayer/relayer/ethereum.go
```diff
@@ -228,6 +231,11 @@ func (sub EthereumSub) Start(completionEvent *sync.WaitGroup) {
 			// The user who starts this must provide a valid last processed block
 			if lastProcessedBlock.Cmp(big.NewInt(0)) == 0 {
 				lastProcessedBlock.Sub(endingBlock, big.NewInt(1))
+			} else if catchUpNeeded {
```
banshee, line 234: This isn't part of the current change, but it's related.  On line 217, we do:

`startingBigInt := newHead.Number
`

Problem is that we're now going to change the number that's inside newHead (since newHead.Number is a pointer to an int, not an int) - that could potentially cause later problems.  I think we need:

`startingBigInt := big.NewInt(newHead.Number.Int64())`
			
unless I'm reading how go does pointers incorrectly.

cmd/ebrelayer/relayer/ethereum.go
ElliotFriedman: Sure.

cmd/ebrelayer/relayer/ethereum.go
```diff
@@ -228,6 +231,11 @@ func (sub EthereumSub) Start(completionEvent *sync.WaitGroup) {
...
+				// if we need to catch up, then do that and let the relayer know that we don't need to catch up again
```
banshee, line 235: The comment doesn't match the code - it says "then do that" but all we're doing is setting the bool.

cmd/ebrelayer/relayer/ethereum.go
```diff
@@ -228,6 +231,11 @@ func (sub EthereumSub) Start(completionEvent *sync.WaitGroup) {
...
+				catchUpNeeded = false
```
juniuszhou, line 236: If catchUpNeeded, then just set is as false, how about the lastProcessedBlock, no update? If use the same block number, the all events in this block number will be processed again.

cmd/ebrelayer/relayer/ethereum.go
ElliotFriedman: New head gets reset any time this channel receives information so we don't need to worry about pointers.

cmd/ebrelayer/relayer/ethereum.go
ElliotFriedman: So it says not to mess with the lastProcessed block if it needs to catch up. Once that bool is set, the next time around, it will miss that statement and change the lastProcessedBlock value to the current block we are looking at on the line 2 lines down.

cmd/ebrelayer/relayer/ethereum.go
juniuszhou: @ElliotFriedman

ElliotFriedman: You set catchUpNeeded to false, remember the last block number for that run, then, the next time you see a block, you just grab data from 50 blocks ago.

ElliotFriedman: I have tested it with thousands of transactions locally and it works just fine.

banshee:
> set limit for processed block as 2, to avoid pull too much events from ethereum one time.

Is this comment still relevant?  Seems like we don't limit it to 2 any more in this PR.

ElliotFriedman:
> > set limit for processed block as 2, to avoid pull too much events from ethereum one time.
> 
> Is this comment still relevant? Seems like we don't limit it to 2 any more in this PR.

I realized that requirement shouldn't really be an issue as I believe we can broadcast many tx's at once to cosmos, and geth shouldn't crash if we ask it for 1000 events across a few blocks.

ElliotFriedman (approved):

PR #654, opened 2021-02-05
[ChainOps] Governance voting parameters.

intl-man:
Minor tweaks to allow us to set the governance voting period.


utx0 (approved):
Nice one

PR #659, opened 2021-02-06
[Peggy] Update integration tests to work against non-local chain

banshee:


test/integration/src/py/test_utilities.py
```diff
@@ -21,12 +26,13 @@ class EthereumToSifchainTransferRequest:
...
     smart_contracts_dir: str = ""
     ethereum_chain_id: str = "5777"
-    chain_id: str = "localnet"
+    chain_id: str = "localnet" #cosmos chain id
     manual_block_advance: bool = True
-    n_wait_blocks: int = 4
+    n_wait_blocks: int = n_wait_blocks
     bridgebank_address: str = ""
     bridgetoken_address: str = ""
     sifnodecli_node: str = "tcp://localhost:26657"
```
ElliotFriedman, line 34: This needs to be parameterized as well to get balances from sifchain on testnet

test/integration/src/py/test_rollback_chain.py
```diff
@@ -30,7 +30,8 @@ def test_rollback_chain(source_ethereum_address):
         ethereum_private_key_env_var="ETHEREUM_PRIVATE_KEY",
         bridgebank_address=get_required_env_var("BRIDGE_BANK_ADDRESS"),
         ethereum_network=(os.environ.get("ETHEREUM_NETWORK") or ""),
-        amount=amount
+        amount=amount,
```
ElliotFriedman, line 33: Maybe this is a failure of imagination on my part, but I don't know how we could test rolling things back on a testnet we don't have control over.

test/integration/src/py/conftest.py
```diff
@@ -79,23 +127,22 @@ def sifchain_fees():
 def source_ethereum_address(is_ropsten_testnet, smart_contracts_dir):
     """account with some starting eth that can be transferred out"""
     addr = test_utilities.get_optional_env_var("ETHEREUM_ADDRESS", "")
+    if addr:
+        logging.debug("using ETHEREUM_ADDRESS provided for source_ethereum_address")
+        return addr
     if is_ropsten_testnet:
+        # Ropsten requires ETHEREUM_ADDRESS
```
ElliotFriedman, line 134: Seems like ropsten won't return an address. Is this intentional?

test/integration/src/py/conftest.py
```diff
@@ -33,14 +42,43 @@ def ethereum_network():
...
+
+
+@pytest.fixture
+def rowan_source_key(is_ropsten_testnet):
+    """A sifchain address or key that has rowan and can send that rowan to other address"""
+    result = test_utilities.get_optional_env_var("ROWAN_SOURCE_KEY", None)
+    if result:
+        return result
+    if is_ropsten_testnet:
+        assert result
```
ElliotFriedman, line 79: We don't need to return this?

test/integration/src/py/test_utilities.py
banshee: Right now, we just have a constant n_wait_blocks set in test_utilities.py.  Since there's no way to get it at runtime (I think?) I'm not sure what else we should do.  This change just uses the constant that everyone else uses; it was incorrectly hardcoded to 4 in two places.

test/integration/src/py/test_rollback_chain.py
banshee: There's no code yet to automate running tests against another chain.  This PR is just to make it possible for a human to follow the instructions in the readme to run tests that they think are sensible against something other than a local ganache test setup.

test/integration/src/py/conftest.py
banshee: You have to specify the environment variable if you're using ropsten; this is just an assert that causes the tests to halt if you haven't set it.  It would have returned before you hit the assert if you have it set.  I updated the comment to hopefully make this more clear.

test/integration/src/py/conftest.py
banshee: No, this just asserts that you've set it manually.  Added a comment.

khdegraaf (approved):
LGTM

khdegraaf (dismissed):

khdegraaf (approved):

PR #663, opened 2021-02-08
[ui] guided liquidity pool calculations 

thomasdavis:


59023g (changes requested):
This is a new feature that I don't want to launch until after launch is stable. Please hold

ui/core/src/hooks/addLiquidityCalculator.ts
```diff
@@ -233,6 +236,22 @@ export function usePoolCalculator(input: {
...
 
+  effect(() => {
+    // if in guided mode
+    // calculate the price ratio of A / B
+    if (input.asyncPooling.value && input.lastFocusedTokenField.value !== null) {
+      if (bPerARatio === null || aPerBRatio === null) {
+        return null;
+      }
+      if (input.lastFocusedTokenField.value === 'A') {
+        input.tokenBAmount.value = (Number(input.tokenAAmount.value) * Number(bPerARatio.value.toFixed(8))).toString();
```
ryardley, line 247: Probably don't want to use `Number` we would be better off using `AssetAmount` instead.

59023g (approved):

PR #666, opened 2021-02-08
[ui] set google analytics to only run on mainnet

thomasdavis:


ui/app/public/index.html
```diff
@@ -9,10 +9,13 @@
     <!-- Global site tag (gtag.js) - Google Analytics -->
     <script async src="https://www.googletagmanager.com/gtag/js?id=G-MSGFSQQEJ3"></script>
     <script>
-      window.dataLayer = window.dataLayer || [];
-      function gtag(){dataLayer.push(arguments);}
-      gtag('js', new Date());
-      gtag('config', 'G-MSGFSQQEJ3');
+      if (window.location.href === "prod.sifchain.finance") {
```
59023g, line 12: Should this be `dex.sifchain.finance`?

thomasdavis: Indeed it should, one of those things I triple checked and still got wrong lol one moment

59023g (approved):
👼

Issue #669, opened 2021-02-09
[Peggy] Need to Remember Where to Restart on Cosmos


ElliotFriedman:
**Describe the bug**
The current architecture of peggy when receiving transactions from cosmos does not remember the last block or transactions that were processed. We need to build a mechanism such that we remember where we left off if the relayer crashes or suddenly goes down.

This should also take care of:
- [ ] Handle peggy restart and cosmos event picking up from where it left off.

intl-man:
Testing...seeing if this actually works!?

PR #671, opened 2021-02-09
[Peggy] Cosmos Per Block Solution

juniuszhou:
the pr is for the issue https://github.com/Sifchain/sifnode/issues/669. ebrelayer will record the processed block number in leveldb.


ElliotFriedman (changes requested):

cmd/ebrelayer/relayer/cosmos.go
```diff
@@ -93,33 +101,67 @@ func (sub CosmosSub) Start(completionEvent *sync.WaitGroup) {
...
+
+	data, err := sub.DB.Get([]byte(cosmosLevelDBKey), nil)
+	if err != nil {
+		log.Println("Error getting the last cosmos block from level db", err)
+		lastProcessedBlock = 0
+	} else {
+		lastProcessedBlock = int64(binary.BigEndian.Uint64(data))
+	}
+
+	for e := range results {
```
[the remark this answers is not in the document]
juniuszhou: the result of cosmos client subscribe is a channel. so loop is for each new block. and I find out there is no handler for quit signal, will add it.

cmd/ebrelayer/relayer/cosmos.go
```diff
@@ -94,33 +102,74 @@ func (sub CosmosSub) Start(completionEvent *sync.WaitGroup) {
...
+							}
+							sub.handleBurnLockMsg(cosmosMsg, claimType)
+						}
 					}
-					// Parse event data, then package it as a ProphecyClaim and relay to the Ethereum Network
-					sub.handleBurnLockMsg(cosmosMsg, claimType)
 				}
+
+				b := make([]byte, 8)
+				binary.BigEndian.PutUint64(b, uint64(blockNumber))
```
ElliotFriedman, line 163: I don't think we need to ensure it is big endian before saving it to the db. There should not be any endian mismatch between the machine that the data is saved on and the machine it is read from.

Unless there is some leveldb semantic that makes this required, let's remove this and have it function like it does on the ethereum.go file in this directory.

ElliotFriedman (changes requested):

cmd/ebrelayer/relayer/cosmos.go
juniuszhou: Use the big.Int for leveldb the same as ethereum.go

cmd/ebrelayer/relayer/cosmos.go
```diff
@@ -99,35 +106,75 @@ func (sub CosmosSub) Start(completionEvent *sync.WaitGroup) {
...
-			tx, ok := result.Data.(tmTypes.EventDataTx)
+		case <-quit:
+			log.Println("we receive the quit signal and exit")
+			return
+
+		case e := <-results:
+			data, ok := e.Data.(tmTypes.EventDataNewBlock)
 			if !ok {
 				sub.SugaredLogger.Errorw("sifchain client failed to extract event data from new tx.",
 					errorMessageKey, err.Error())
```
ElliotFriedman, line 129: err is defined somewhere else, so instead of having this logger spit out this error message that will likely be nil. Let's remove line 129. On line 128, let's have a message that says `sifchain client failed to extract event data from new block."` and then prints out the value in `tmTypes.EventDataNewBlock` so that we know why there was a failure and we can add failsafes in if this ever happens.

ElliotFriedman (changes requested):

cmd/ebrelayer/relayer/cosmos.go
juniuszhou: Accepted, will fix the issue.

banshee:
Added tests in #977

cmd/ebrelayer/relayer/cosmos.go
```diff
@@ -98,35 +106,75 @@ func (sub CosmosSub) Start(completionEvent *sync.WaitGroup) {
...
+			if lastProcessedBlock == 0 {
+				lastProcessedBlock = blockHeight
+			}
 			sub.SugaredLogger.Infow("new transaction witnessed in sifchain client.")
 
-			// Iterate over each event in the transaction
-			for _, event := range tx.Result.Events {
-				claimType := getOracleClaimType(event.GetType())
+			startBlockHeight := lastProcessedBlock + 1
+			fmt.Printf("cosmos process events from block %d to %d\n", startBlockHeight, blockHeight)
```
ElliotFriedman, line 140: Let's change this to a sugared logger call instead of a printf for consistency with `startingBlockHeight` and `currentBlockHeight` as keys.

ElliotFriedman (changes requested):

cmd/ebrelayer/relayer/cosmos.go
juniuszhou: should use sugared logger, will fix it.

ElliotFriedman (approved):

PR #674, opened 2021-02-09
[ui] bug/swap-reverse-invalid-amount

59023g:
Fixes issue where user able to click "swap" button with an invalid calculated swap amount. Screen Shot 2021-02-09 at 11 02 41 AM


iverc:
@59023g Can we use existing "ZERO_AMOUNTS" that will change text to "Please enter an amount"?

iverc (approved):
You may or may not address my comment above, both options will work.

59023g:
> You may or may not address my comment above, both options will work.

Ok. I will stick with what I have since we agreed on "Invalid Amount" and I think it's more accurate.

PR #675, opened 2021-02-09
Remove slither workflow - we don't use the automated version of this.

banshee:
And the upstream install changed so it's currently breaking.


khdegraaf (approved):
LGTM

PR #676, opened 2021-02-09
[ui] swap, arrow click fix, also, disable unpeg input 

59023g:
Fixes swap issue where input values did not correctly update on arrow click. Expected behavior: onclick, "from" value persists, but in "to" input, and new "from" value updates

Also, includes fix to dsiable unpeg recipient address input

https://user-images.githubusercontent.com/5235538/107454746-399be200-6b13-11eb-81a0-b2e7ef5666c6.mov


ui/app/src/views/SwapPage.vue
```diff
@@ -110,6 +110,8 @@ export default defineComponent({
     }
 
     function swapInputs() {
+      // we want the "from" value to persist, but now in the "to" input
+      selectedField.value = "to";
```
iverc, line 114: Should we also toggle the selected value, rather than just setting it to "to"?

iverc: @59023g Also, I think we are generally not doing it correctly. This should swap both amounts and tokens (check Uniswap for reference). Otherwise, it does not make sense to swap these fields

iverc (changes requested):

ui/app/src/views/SwapPage.vue
59023g: It does swap both amounts and tokens. Although the "from" amount updates. Which it was not before. This is the same behavior as Uniswap.

I will look into code suggestion

iverc (approved):
🚀

PR #686, opened 2021-02-10
[ui] bug/decimals

59023g:
This applies particularly to display values. Fixes future issue where app would crash if decimals of asset on pegAssetPage < 18. This would apply to main net tokens like `USDT` which have 6 decimals. 

Also, applies `formatNumber` in AssetList (Peg)

### DIFF
--- a/ui/app/src/components/shared/utils.ts
+++ b/ui/app/src/components/shared/utils.ts
@@ -16,6 +16,7 @@ export function formatPercentage(amount: string) {
 }
 // TODO: make this work for AssetAmounts and Fractions / Amounts
 export function formatNumber(displayNumber: string) {
+  if (!displayNumber) return "0"
   const amount = parseFloat(displayNumber);
   if (amount < 100000) {
     return amount.toFixed(5);
--- a/ui/app/src/views/PegAssetPage.vue
+++ b/ui/app/src/views/PegAssetPage.vue
@@ -110,7 +110,7 @@ export default defineComponent({
 
     const nextStepAllowed = computed(() => {
       const amountNum = new BigNumber(amount.value);
-      const balance = accountBalance.value?.toFixed(18) ?? "0.0";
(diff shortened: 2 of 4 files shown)


ui/app/src/components/shared/utils.ts
```diff
@@ -16,6 +16,7 @@ export function formatPercentage(amount: string) {
 }
 // TODO: make this work for AssetAmounts and Fractions / Amounts
 export function formatNumber(displayNumber: string) {
```
iverc, line 18: should it be `string | undefined` ?

ui/app/src/views/PegAssetPage.vue
```diff
@@ -110,7 +110,7 @@ export default defineComponent({
 
     const nextStepAllowed = computed(() => {
       const amountNum = new BigNumber(amount.value);
-      const balance = accountBalance.value?.toFixed(18) ?? "0.0";
+      const balance = accountBalance.value?.toFixed() ?? "0.0";
```
iverc, line 113: @59023g so here we let the Asset.decimals decide how many decimals to show, right? And this is only for displaying. Before we send it to the backend, we have to add missing decimals for assets with decimals < 18? (e.g. fill missing decimals with 0, essentially the opposite of `trimZeroes` that we have in `swapCalculator`).

iverc (approved):

ui/app/src/views/PegAssetPage.vue
59023g: `nextStepAllowed` returns a boolean, which decides if user can click the submit button to Peg or not. So this `balance` variable is limited in scope to `nextStepAllowed` as its used to decide truthiness. Does not go to display beyond said truthiness.
Screen Shot 2021-02-10 at 12 34 45 AM



As far as "Before we send it to the backend, we have to add missing decimals for assets with decimals < 18" - I don't know. I did test peg function while changing decimal places to < 18. It appeared to work.

Issue #670, opened 2021-02-09
[Peggy] Handling Errors with Blocks


ElliotFriedman:
**Describe the bug**
Currently, when we receive an error from a block, we skip over it and go to the next one. Instead, we need to retry until we get that block processed and then move on.

PR #694, opened 2021-02-10
[peggy] retry if any error from process block

juniuszhou:
the pr is for https://github.com/Sifchain/sifnode/issues/670. we need retry the block if any error instead of skip it.


cmd/ebrelayer/relayer/ethereum.go
```diff
@@ -234,14 +231,9 @@ func (sub EthereumSub) Start(completionEvent *sync.WaitGroup) {
 			// The user who starts this must provide a valid last processed block
 			if lastProcessedBlock.Cmp(big.NewInt(0)) == 0 {
 				lastProcessedBlock.Sub(endingBlock, big.NewInt(1))
-			} else if catchUpNeeded {
-				// if we need to catch up, then do that and let the relayer know that we don't need to catch up again
-				catchUpNeeded = false
-			} else {
-				lastProcessedBlock = endingBlock
```
juniuszhou, line 241: lastProcessedBlock only updated if the block processed by relayer successfully. we can't set it as as endingBlock everytime if we should handling errors with blocks as https://github.com/Sifchain/sifnode/issues/670.

cmd/ebrelayer/relayer/ethereum.go
```diff
@@ -234,14 +231,9 @@ func (sub EthereumSub) Start(completionEvent *sync.WaitGroup) {
 			// The user who starts this must provide a valid last processed block
 			if lastProcessedBlock.Cmp(big.NewInt(0)) == 0 {
 				lastProcessedBlock.Sub(endingBlock, big.NewInt(1))
```
ElliotFriedman, line 233: In this case, if you want to delete that other logic, if lastProcessedBlock is equal to 0, then you should just set last processed block to the ending block instead of ending block minus one.

ElliotFriedman (changes requested):

cmd/ebrelayer/relayer/ethereum.go
```diff
@@ -234,14 +231,9 @@ func (sub EthereumSub) Start(completionEvent *sync.WaitGroup) {
...
-			} else if catchUpNeeded {
```
ElliotFriedman, line 237: Have you tested these changes by killing the relayer, sending a bunch of transactions, and then restarting the relayer to see if it picks up on the transactions that it misses?

cmd/ebrelayer/relayer/ethereum.go
juniuszhou: agree, will update it.

cmd/ebrelayer/relayer/ethereum.go
juniuszhou: I tried the killing relayer case, and just checked at block level. Relayer can start from the lastProcessedBlock stored in level db. will try some transaction.

ElliotFriedman: So you killed it, then you advanced blocks and submitted transactions, then when you woke it back up, it processed all blocks that it missed when asleep? That needs to happen to approve this PR.

juniuszhou: the test case is verified. ebrelayer restart and processed transactions it missed during offline.

ElliotFriedman: That is good. We will need to funnel this through the pipeline going from swingset to sandpit to mainnet with a release.

PR #695, opened 2021-02-10
[ui] don't show peg buttons when asset amount is zero

thomasdavis:


ui/app/src/components/shared/AssetList.vue
```diff
@@ -4,7 +4,7 @@
       <AssetItem class="token" :symbol="item.asset.symbol" />
       <div class="amount">{{ formatNumber(item.amount) }}</div>
       <div class="action">
-        <slot v-if="!!item.amount" :asset="item"></slot>
+        <slot v-if="Number(item.amount.toFixed(6)) > 0" :asset="item"></slot>
```
59023g, line 7: This may break app if Token has less than 6 decimals. Will you test it by changing a token in `assets.ethereum.ropsten.json` to have 5 decimals?

59023g (changes requested):
Please remove `toFixed(6)` - Can replace it with empty param, will then default to whatever `asset.decimal` value is. There's an issue where toFixed crashes app when param is greater than `asset.decimals`. Which only occurs for ETH assets

thomasdavis:
@59023g Yeah that works better, tested and pushed.

59023g (approved):
Cool

PR #711, opened 2021-02-11
[Peggy] manual set ethereum nonce.

juniuszhou:
ebrelayer will store and set the ethereum nonce instead of get from client.


test/integration/src/py/test_random_currency_roundtrip.py
```diff
@@ -73,7 +73,7 @@ def test_transfer_tokens_with_some_currency(
         ethereum_network,
     )
 
-
+@pytest.mark.skip(reason="ethereum nonce set by relayer now")
```
banshee, line 76: I'll fix this test before we merge this code.

banshee (changes requested):
I need to fix the test.

banshee:
Instead of fixing it this way, we should think about fixing it by waiting for the transaction id.

banshee:
Should we be waiting for the transaction?

```
	for needFinalTransaction := true; needFinalTransaction; {
		// Get the transaction receipt
		receipt, err := client.TransactionReceipt(context.Background(), tx.Hash())
		if err != nil {
			log.Println(err)
			return err
		} else {
			switch receipt.Status {
			case 0:
				log.Println("Tx Status: 0 - Failed")
				// and we need to do something here
			case 1:
				log.Println("Tx Status: 1 - Successful")
			}
			needFinalTransaction = false
		}
	}

```

ElliotFriedman (approved):

banshee (approved):
None of my issues prevent this from merging.

musnit:
same concerns as before - why are we doing manual memory management with golang? this kinda defeats the purpose of using go.

https://golang.org/pkg/sync/atomic/ - see second paragraph

anyway, if it fixes the problem, then great for now.

i think we will need to rearchitect/refactor peggy to use proper go standards and paradigms for communication as part of the task to build multi-chain support.

PR #717, opened 2021-02-12
new pool unit tests for BE fix

AustinoBombino:
New test table for pool units that accounts for edge cases associated with these changes to sifnode:

> Set the minimal input to 1e9 (or 0.000000001 in terms of tokens). This limitation is still quite generous, because even if it is the most expensive token, it will be 1/100th of a cent.
> Before doing math, find the smallest input by checking the length of each one and divide all inputs by the smallest input.
> The max input could be limited to 1e33, which is a lot more than the total of minted tokens for all existing ones.
> For example, if we have 1e25, 1e20, 1e18, and 1e9, the min would be 1e9. Hence, after normalizing, we will have 1e14, 1e11 1e9, and 1, and the result of multiplication will be way below 1e75.


iverc (approved):
👍 🚀

Jedi2002 (approved):

iverc:
Can be closed since this table was added in PR #714

PR #747, opened 2021-02-16
[Peggy] add an interface to update threshold for cosmos bridge contract

juniuszhou:
we need update the threshold for cosmos bridge contract, then reduce the total gas cost of all ebrelayers.


smart-contracts/contracts/CosmosBridge.sol
```diff
@@ -249,4 +249,12 @@ contract CosmosBridge is CosmosBridgeStorage, Oracle {
     {
         return string(abi.encodePacked(_prefix, _suffix));
     }
+
+    /*
+     * @dev: updateProphecyThreshold
+     *       Issues a request for the Oracle to update threshold
+     */
+    function updateProphecyThreshold(uint256 _consensusThreshold) public onlyOperator {
```
ElliotFriedman, line 257: Could you please add docs that explain why we can't call this function unless the contract is paused?

If someone calls this function while the contract is not paused, this could lead to bad results as the threshold drops, but the user might not receive a payout..

smart-contracts/test/test_cosmosBridge.js
```diff
@@ -517,4 +517,68 @@ contract("CosmosBridge", function (accounts) {
...
+      await this.cosmosBridge.updateProphecyThreshold( 
+        newConsensusThreshold, {
+        from: operator
+      }).should.be.fulfilled;
+
+      const value = await this.cosmosBridge.consensusThreshold();
+      assert.equal(value, newConsensusThreshold);
+
+      const zeroConsensusThreshold = 0;
+      await expectRevert(this.cosmosBridge.updateProphecyThreshold(
```
ElliotFriedman, line 561: This should be a separate test in a different it block.

smart-contracts/test/test_cosmosBridge.js
```diff
@@ -517,4 +517,68 @@ contract("CosmosBridge", function (accounts) {
...
+
+      const zeroConsensusThreshold = 0;
+      await expectRevert(this.cosmosBridge.updateProphecyThreshold( 
+        zeroConsensusThreshold, {
+        from: operator
+      }),
+      "Consensus threshold must be positive.")
+
+      const largeConsensusThreshold = 101;
+      await expectRevert(this.cosmosBridge.updateProphecyThreshold(
```
ElliotFriedman, line 568: This should be a separate test in a different it block.

smart-contracts/test/test_cosmosBridge.js
```diff
@@ -517,4 +517,68 @@ contract("CosmosBridge", function (accounts) {
...
+      "Consensus threshold must be positive.")
+
+      const largeConsensusThreshold = 101;
+      await expectRevert(this.cosmosBridge.updateProphecyThreshold( 
+        largeConsensusThreshold, {
+        from: operator
+      }),
+      "Invalid consensus threshold.")
+
+      await expectRevert(this.cosmosBridge.updateProphecyThreshold(
```
ElliotFriedman, line 574: This should be a separate test in a different it block.

ElliotFriedman (changes requested):

smart-contracts/contracts/CosmosBridge.sol
juniuszhou: Agree, it is better update the threshold during pausing the contract. Will add it in document.

smart-contracts/test/test_cosmosBridge.js
juniuszhou: will use a new it block.

docs/peggy/Peggy-Smart-Contract-Threshold.md
```diff
@@ -0,0 +1,5 @@
+# Peggy Admin Smart Contract Threshold
+
+In CosmosBridge contract, the validators can send newProphecyClaim transaction to the contract. And each validator has its voting power. Threshold is important argument, represents how many validators need send newProphecyClaim transactions before prophecy claim be finalized.
+
+Sometimes, we need update the threshold. But the operation may lead to side effect, like someone might not receive a payout. So we should puase the contract first, then update the threshold.
```
ElliotFriedman, line 5: We also need to stop the relayers before we update the threshold.

ElliotFriedman: Please add that to the doc.

ElliotFriedman (changes requested):

docs/peggy/Peggy-Smart-Contract-Threshold.md
juniuszhou: added.

utx0:
Should this PR be put on onhold right now guys? @ElliotFriedman @juniuszhou

ElliotFriedman (approved):

ElliotFriedman:
Team decided against this as it decreases decentralization.

PR #748, opened 2021-02-16
[ui] don't display peg or unpeg buttons when the user doesn't have an…

thomasdavis:
…y meaninful balances


59023g (approved):

ui/app/src/components/shared/AssetList.vue
```diff
@@ -4,7 +4,7 @@
       <AssetItem class="token" :symbol="item.asset.symbol" />
       <div class="amount">{{ formatNumber(item.amount) }}</div>
       <div class="action">
-        <slot v-if="Number(item.amount.toFixed()) > 0" :asset="item"></slot>
+        <slot v-if="Number(item.amount.toFixed()) >= 0.00001" :asset="item"></slot>
```
59023g, line 7: @iverc What should this number be?

thomasdavis: The reason I chose this value, is because that is the limits of decimals that we show to users. So if their balances are less than that we display 0.00000 anyway, so my assumption is that we should just hide the peg/unpeg if it looks like there balance is `0.00000` anyway.

59023g (approved):

intl-man:
@thomasdavis Closing as the UI has since been decoupled.

PR #749, opened 2021-02-16
extreme case tables for clp decimal fix

AustinoBombino:


iverc (approved):

iverc:
Changes added in #750

PR #750, opened 2021-02-16
Token decimals normalization

iverc:


x/clp/keeper/calculations.go
```diff
@@ -117,14 +143,31 @@ func CalculateWithdrawal(poolUnits sdk.Uint, nativeAssetBalance string,
...
+func CalculatePoolUnits(symbol string, oldPoolUnits, nativeAssetBalance, externalAssetBalance,
 	nativeAssetAmount, externalAssetAmount sdk.Uint) (sdk.Uint, sdk.Uint, error) {
-	// refactor this to use ValidInputs
+	normalizationFactor := sdk.NewDec(1)
+	nf, ok := types.GetNormalizationMap()[symbol[1:]]
+	adjustExternalToken := true
+	if ok {
+		diffFactor := 18 - nf
+		if diffFactor < 0 {
+		    diffFactor = nf - 18
```
ElliotFriedman, line 154: If diff is negative, why would we subtract 18 from it?

x/clp/types/keys.go
```diff
@@ -46,3 +46,20 @@ func GetLiquidityProviderKey(externalTicker string, lp string) []byte {
 	key := []byte(fmt.Sprintf("%s_%s", externalTicker, lp))
 	return append(LiquidityProviderPrefix, key...)
 }
+
+func GetNormalizationMap() map[string]int64 {
```
ElliotFriedman, line 50: This should be a keeper that only an admin key can change. Otherwise, we have to upgrade the network and manually change this list before we can add new tokens to the dex whose token decimal length is not 10e18.

x/clp/keeper/calculations.go
```diff
@@ -192,19 +236,6 @@ func CalculatePoolUnits(oldPoolUnits, nativeAssetBalance, externalAssetBalance,
 // Add validations for X,x,Y
 //( x^2 * Y ) / ( x + X )^2
 func calcLiquidityFee(X, x, Y sdk.Uint) (sdk.Uint, error) {
-	// if inputs are outside range return error
```
ElliotFriedman, line 195: Are we removing the liquidity fee?

ElliotFriedman (changes requested):

x/clp/keeper/calculations.go
iverc: @ElliotFriedman  that's what we do in this line, `nf - 18`

x/clp/types/keys.go
iverc: @Jedi2002 Could you please take a look into moving it into keeper?

x/clp/keeper/calculations.go
iverc: No, just commented code

x/clp/types/keys.go
Jedi2002: This is not a finalised PR yet , this would be part of the keeper eventually , and we would need a separate TX to add / remove from this list. 
We also need to discuss who gets to execute the add / remove tx .

Jedi2002: I do propose to limit it to clp - admins as of now , which seems to be the most viable approach

Jedi2002: Also tbh I feel I, we need to gdo that as a separate pr , this pr can just be for the mathmatical changes and test . Would be easier to review .

x/clp/keeper/calculations.go
```diff
@@ -117,14 +141,31 @@ func CalculateWithdrawal(poolUnits sdk.Uint, nativeAssetBalance string,
...
-	// refactor this to use ValidInputs
+	normalizationFactor := sdk.NewDec(1)
+	nf, ok := types.GetNormalizationMap()[symbol[1:]]
+	adjustExternalToken := true
+	if ok {
+		diffFactor := 18 - nf
+		if diffFactor < 0 {
+			diffFactor = nf - 18
+			adjustExternalToken = false
+		}
```
intl-man, line 154: How is this possible? Could we realistically have an asset that uses a larger precision than 1e18? (if I'm understanding this logic correctly).

Also, not a fan of just having numbers like these floating around... might be better to set as a constant (or better yet use the `Precision` constant in `github.com/cosmos/cosmos-sdk/client/types`) ... but guess that can be done another day.

x/clp/keeper/calculations.go
```diff
@@ -189,55 +226,110 @@ func CalculatePoolUnits(oldPoolUnits, nativeAssetBalance, externalAssetBalance,
...
-	//if !ValidateInputs([]sdk.Uint{X, x, Y}) {
-	//	return sdk.ZeroUint(), types.ErrInvalid //error
-	//}
-	////if any input is 0 return 0
-	//if !ValidateZero([]sdk.Uint{X, x, Y}) {
-	//	return sdk.ZeroUint(), nil
-	//}
-	//d := x.Add(X)
-	//denom := d.Mul(d)
-	//return (x.Mul(x).Mul(Y)).Quo(denom), nil
```
intl-man, line 206: Sorry, just a pet peeve of mine -- commented code being merged in.... do we need this?

x/clp/keeper/calculations.go
```diff
@@ -277,16 +359,32 @@ func ReducePrecision(dec sdk.Dec, po int64) sdk.Dec {
 	return dec.Quo(p)
 }
 
+func ReducePrecisionInt(num sdk.Int, po int64) sdk.Int {
+	multiplier := big.NewInt(10).Exp(big.NewInt(10), big.NewInt(po), nil)
+	multiplierInt := sdk.NewIntFromBigInt(multiplier)
+	fmt.Println(num.String(), multiplierInt.String())
+	val := num.Quo(multiplierInt)
+	fmt.Println("val", val)
```
intl-man, line 367: Is there a reason why we're sending this to STDOUT? Debug?

x/clp/keeper/calculations.go
iverc: @intl-man this is not yet possible, but could be in the future if token with 18+ decimals is added, so I thought better to handle it right away. Agree about constants, there are many things including them that could be refactored

iverc: @utx0 @intl-man btw, the token list in the repo assets has latest? Want to double check that we have all tokens with < 18 decimals in keys.go file, otherwise results can be incorrect

x/clp/keeper/calculations.go
utx0: yeah that needs to be cleaned up.

iverc: On it

x/clp/keeper/calculations.go
utx0: +1

iverc: Fixing

x/clp/keeper/table_test.go
```diff
@@ -32,20 +33,132 @@ func TestCalculatePoolUnits(t *testing.T) {
...
+			Ay,
+			sdk.NewUintFromString(test.BY))
+
+		if test.Expected != "0" && !By.Equal(sdk.NewUintFromString(test.Expected)) {
+			errCount++
+			fmt.Printf("TestCalculateDoubleSwapResult: Was %s, Got %s , Expected %s \n %v \n", Ay, By, strings.Split(test.Expected, ".")[0], test)
+		}
 	}
-	fmt.Println("Error Count :", errCount, "Total :", len(testcases))
+	fmt.Printf("\nTotal/Failed: %d/%d \n", len(testcases), errCount)
```
utx0, line 163: Be good to clean these up in tests too.

utx0 (changes requested):
A little clean up work and also codecov is highlighing a few code paths without tests. Can we looks at adding extra unit tests here too please? Thanks

x/clp/keeper/calculations.go
intl-man: Sorry, looks like all that was removed anyway @iverc !

iverc:
@utx0 @intl-man done, please take another look. one print can't be removed cause we get lint error without it ⚠️

utx0 (approved):
Nice work guys! Thanks for the effort

PR #751, opened 2021-02-17
[ui] Smoke Test Bug Fixes

59023g:
- Fix issue where values on pool confirmation dialog were flipped
- Fix issue where Keplr config received ALL currencies. Now only receives Sifchain ones


ryardley (approved):
Looks good!

PR #762, opened 2021-02-17
Fix token list on sandpit

iverc:


ui/app/src/components/shared/DetailsPanel.vue
```diff
@@ -61,10 +61,9 @@ export default defineComponent({
           </Tooltip>
         </span>
 
-        <span
-          >{{ formatNumber(minimumReceived) }}
-          <span>{{ toToken.toUpperCase().replace("C", "c") }}</span></span
-        >
+        <span>{{ formatNumber(minimumReceived) }}
+          <span>{{ toToken.slice(1).toUpperCase() }}</span>
```
59023g, line 65: What's this

59023g (approved):

ui/app/src/components/shared/DetailsPanel.vue
iverc: @59023g if we replace C with c it will mess up tokens like USDc

59023g: but it's `cUSDC`

59023g: Also, `slice` will remove string character at index 0. Please revert this.

59023g: Also, slice will remove string character at index 0. Please revert this.

59023g (changes requested):

59023g (approved):
🆒

PR #763, opened 2021-02-17
More fixes

iverc:


ui/app/src/components/shared/utils.ts
```diff
@@ -16,10 +16,10 @@ export function formatPercentage(amount: string) {
 }
 // TODO: make this work for AssetAmounts and Fractions / Amounts
 export function formatNumber(displayNumber: string) {
-  if (!displayNumber) return "0"
+  if (!displayNumber) return "0";
   const amount = parseFloat(displayNumber);
   if (amount < 100000) {
-    return amount.toFixed(5);
+    return amount.toFixed(6);
```
59023g, line 22: Please test this against main net tokens: ie `USDT`. It's breaking.

59023g (changes requested):
`toFixed(decimals)` implementation with set decimals param breaks app with tokens like `USDT`

59023g (approved):

ui/app/src/components/shared/utils.ts
59023g: I'm still concerned about this issue. But more concerned about getting PR in front of testers.

59023g: I know it's just display stuff. If it throws error, they won't be able to see number. Or, over the weekend, whole sections of the interface were broken. Which token has 2 decimals?

59023g: I advocated to leave it for now since everything's been thoroughly tested. And this function affects much of the app. Hopefully I'm wrong about all this.

PR #764, opened 2021-02-17
[ui] app/CreatePoolPage: fix back link logic

59023g:
Fixes issue where back button from `/pool/add-liquidity/<token>` went back to `/pool` instead of `/pool/<token>`


ryardley (approved):
🚀

iverc (approved):

PR #765, opened 2021-02-17
[ui] bug/max button peg/unpeg

ryardley:
For this card https://app.asana.com/0/1199924158691563/1199952313470063/f

Looks like we were using Number to calculate max. This uses our Fraction representation instead.


59023g (approved):
Cool

iverc (approved):

PR #774, opened 2021-02-18
[ChainOps] Monitoring.

intl-man:
Integration of Halborn's logstash work.

### DIFF
--- a/deploy/helm/logstash/values.yaml
+++ b/deploy/helm/logstash/values.yaml
@@ -4,7 +4,7 @@ logstashConfig:
     xpack.monitoring.enabled: true
     xpack.monitoring.elasticsearch.username: '${ELASTICSEARCH_USERNAME}'
     xpack.monitoring.elasticsearch.password: '${ELASTICSEARCH_PASSWORD}'
-    xpack.monitoring.elasticsearch.hosts: ["https://54.90.115.189:9200","https://54.235.50.44:9200","https://54.226.53.195:9200"]
+    xpack.monitoring.elasticsearch.hosts: ["http://elk-prod-1125755348.us-east-1.elb.amazonaws.com"]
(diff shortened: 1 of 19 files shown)


deploy/helm/logstash/values.yaml
```diff
@@ -0,0 +1,221 @@
...
+      }
+     }
+      filter {
+        mutate {
+          add_field => { "cluster" => '{{ .Values.logstash.args.cluster }}' }
+        }
+    }
+    output {
+      elasticsearch {
+        hosts => ["http://elk-prod-1125755348.us-east-1.elb.amazonaws.com:80"]
```
utx0, line 65: do we want to setup a dns for this?

deploy/helm/logstash/values.yaml
```diff
@@ -0,0 +1,221 @@
...
+    filter {
+      json {
+        source => "message"
+        remove_field => ["message"]
+        add_field => { "cluster" => '{{ .Values.logstash.args.cluster }}' }
+      }
+    }
+    output {
+      elasticsearch {
+        hosts => ["http://elk-prod-1125755348.us-east-1.elb.amazonaws.com:80"]
```
utx0, line 42: should we setup a dns here too?

utx0 (approved):
Nice one bro.

deploy/helm/logstash/values.yaml
intl-man: Yep, once we've figured out the SSL issues we're seeing with logstash and elasticsearch. The kibana endpoint is actually available via a nice DNS entry already.

deploy/helm/logstash/values.yaml
intl-man: As above :)

utx0 (approved):

PR #777, opened 2021-02-18
[Peggy] Integration tests enhanced currency tests for sandpit

banshee:


test/integration/src/py/test_new_currency_transfers.py
```diff
@@ -79,19 +94,6 @@ def test_can_create_a_new_token_with_a_long_name_and_peg_it(
     burn_lock_functions.transfer_ethereum_to_sifchain(request1, 10)
```
khdegraaf, line 94: If you are trying to do max sifchain symbol name, this should be 9, given you are appending an "a" in front.  Or subtract one...

khdegraaf (approved):

PR #779, opened 2021-02-18
Fix issue with tabs showing the wrong peg/unpeg buttons. Remove bad config

ryardley:
There was an issue with the tab component where it was not setup to manage defaults properly.

This adds a `defaultIndex` prop to manage the default selected Index.

This meant that the wrong token peg and unpeg links were being shown to the user which meant that peg and unpeg were not working when clicked through

There was also some bad config in the localconfig where sifchain assets were being listed in the ethereum asset config.

### DIFF
--- a/ui/app/src/components/shared/Tabs.vue
+++ b/ui/app/src/components/shared/Tabs.vue
@@ -54,21 +54,11 @@
 }
 </style>
 <script lang="ts">
-import {
-  defineComponent,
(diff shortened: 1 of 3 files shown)


thomasdavis (approved):

59023g (approved):
Great. Tested working. Will need to enhance the 'back' button eventually

PR #780, opened 2021-02-19
remove mainnet tokens

iverc:


ryardley:
So I am wondering we need to be careful here as these tokens are also matched to the external tokens. I wonder if that means that people can peg external tokens (via CLI lets say) to ctokens and if we have not got those configured here things might break in the frontend.

PR #782, opened 2021-02-19
[ui] fix for making back button work on pooling in different states

thomasdavis:


59023g (approved):

ryardley (dismissed):
I believe this also fixes a bug where the whole page was disappearing when you hit back and dont have funds in a pool.

ui/app/src/views/CreatePoolPage.vue
```diff
@@ -200,6 +200,15 @@ export default defineComponent({
         selectedField.value = null;
       },
 
+      backlink: computed(() => {
+        let link = '/pool';
+        if (fromSymbol.value && connected && Number(aPerBRatioMessage.value) > 0 &&
```
ryardley, line 205: Just had a thought here it might be better to try and just use `router.back()`

59023g (changes requested):
I just tested this again. Doesn't work: `/pool/cctoken -> /pool/add-liquidity/cctoken "click back" -> /pool`

However, I tried @ryardley 's suggestion and all the cases *I know* worked. I know there's a case I'm missing.

59023g (approved):
Ship it!!

PR #799, opened 2021-02-20
[ChainOps] Set CHAINNET correctly on release builds.

intl-man:
Includes:

* Correctly sets the CHAINNET to either `testnet` or `mainnet`.

Overview:

Currently if you go to a production node endpoint, e.g.:

https://api.sifchain.finance/node_info

you'll see that `build_tags` is set to `testnet`. We need this to be set to `mainnet` (when running in `mainnet`) so that the faucet is not made available (it's unusable, but shouldn't be available in `mainnet` regardless).


59023g:
Related? https://github.com/Sifchain/issues/issues/22

intl-man:
> Related? [Sifchain/issues#22](https://github.com/Sifchain/issues/issues/22)

No, not related.

AustinoBombino (approved):

utx0 (approved):

PR #802, opened 2021-02-20
[ui] show all A/from assets when pooling

thomasdavis:


59023g (changes requested):
Hey, What's the ticket for this? 
- I see an issue where user can create a Rowan/Rowan
- This breaks the select dialog search
Screen Shot 2021-02-22 at 11 34 48 AM

ui/app/src/views/CreatePoolPage.vue
```diff
@@ -264,6 +264,7 @@ export default defineComponent({
       /></template>
       <template v-slot:default="{ requestClose }">
         <SelectTokenDialogSif
+          forceShowAllATokens="true"
```
ryardley, line 267: Should this be `:forceShowAllATokens="true"`?

ui/app/src/views/CreatePoolPage.vue
```diff
@@ -307,7 +306,7 @@ export default defineComponent({
         </FatInfoTableCell>
         <FatInfoTableCell>
           <span class="number">{{
-            formatNumber(bPerARatioProjectedMessage)
+            formatNumber(bPerARatioProjectedMessage === 'N/A' ? '0' : bPerARatioProjectedMessage)
```
iverc, line 309: can we use same logic like in DetailsPanePool `realBPerA`?

ui/core/src/hooks/addLiquidityCalculator.ts
```diff
@@ -33,10 +33,37 @@ export function usePoolCalculator(input: {
   const tokenBField = useField(input.tokenBAmount, input.tokenBSymbol);
   const balanceMap = useBalances(input.balances);
 
+
+  const preExistingPool = computed(() => {
+    if (!tokenAField.asset.value || !tokenBField.asset.value) {
```
iverc, line 38: It does not 100% mean that there is no existing pool, it could be the case that pool was created, but then one or both assets withdrawn completely. Most likely will never happen, but possible technically.

iverc: let's do some testing after moving things around

59023g (approved):
- Search works
- Rowan to Rowan fixed
- "insufficient funds" is not the best message to display if pool doesn't exist, but is edge case

Looks good to me. But I do see @iverc had some questions.

PR #809, opened 2021-02-20
[ui] Sync pools after address update and optimize liquidity provider pool data structure

ryardley:
This should solve some of the issues with pools taking a long time to load

* Using hash to store accountPools which means we don't need to wait for all the requests to return to see account pools
* Remove redundant data that now cannot get out of sync accountPool now only stores a pool ticker
* Sync pools after an address update

This does not:
* Display a loading spinner - because we will be better off working out a good pattern around queries and metastate in order to do that we need to refactor and this PR covers the main issue here. At a later point we could look at some kind of swr pattern for handling this kind of remote state but we would want to look at things cohesively as part of architecture review.
* Display error messaging when we have network errors - this caused a few errors but they were surfaced in the interface as notifications so that was not the problem here


ui/app/src/views/CreatePoolPage.vue
```diff
@@ -79,9 +79,9 @@ export default defineComponent({
     const liquidityProvider = computed(() => {
       if (!fromSymbol) return null;
       return (
-        store.accountpools.find((pool) => {
-          return pool.lp.asset.symbol === fromSymbol.value;
-        })?.lp ?? null
+        store.accountpools[store.wallet.sif.address][
+          `${fromSymbol.value}_rowan`
```
iverc, line 83: 👍

ui/app/src/views/SinglePool.vue
```diff
@@ -4,28 +4,36 @@ import { computed, toRefs } from "@vue/reactivity";
...
-import { Fraction, LiquidityProvider, Pool, usePoolCalculator, AssetAmount, Asset } from "ui-core";
+import { Fraction } from "ui-core";
 import { useCore } from "@/hooks/useCore";
-import { useRoute } from 'vue-router';
-import { getBlockExplorerUrl } from "../components/shared/utils"
+import { useRoute } from "vue-router";
+import { getBlockExplorerUrl } from "../components/shared/utils";
 
-const DECIMALS = 5
+const DECIMALS = 5;
```
iverc, line 12: @59023g @ryardley I believe we should do 6 as a minimum. We show 6 in many places, and only in pools/liquidity part we show 5 for some reason

iverc (approved):

ui/app/src/views/SinglePool.vue
ryardley: Good idea but not sure that is part of this mandate though?

59023g (approved):
This needs a basic spinner. Some sort of visual indication to the user their pools are loading.

PR #813, opened 2021-02-22
[peggy] lower gas limit.

juniuszhou:
From the transaction in Ethereum, the real gas used is about 0.01 eth. The gas limit in ebrelayer is 3M, it is too much. lower the value to 200K. And increse the gas price to 160 gagiwai for charged ceth in sifnode side.


cmd/ebrelayer/txs/relayToEthereum.go
```diff
@@ -111,12 +106,25 @@ func initRelayConfig(provider string, registry common.Address, event types.Event
...
-	log.Printf("ethereum tx current nonce from local storage is %d\n", nextNonce)
+
+	fmt.Println("suggested gas price: ", gasPrice)
+
+	gasPrice = gasPrice.Mul(gasPrice, big.NewInt(2))
+	fmt.Println("suggested gas price after multiplying by 2: ", gasPrice)
+
+	quarterGasPrice := big.NewInt(0)
+	quarterGasPrice = quarterGasPrice.Div(gasPrice, big.NewInt(4))
+	fmt.Println("quarterGasPrice: ", quarterGasPrice)
```
banshee, line 117: Remove the debug logging

ElliotFriedman: Let's merge to sandpit with this, then when we merge to betanet remove it. That way, we can get live debugging while testing it.

ElliotFriedman (approved):

PR #820, opened 2021-02-22
Liquidity bugfix, fix symbols, cleanup

iverc:


ui/app/src/views/PegListingPage.vue
```diff
@@ -111,14 +111,14 @@ export default defineComponent({
       searchText,
       peggedSymbol(unpeggedSymbol: string) {
         if (unpeggedSymbol.toLowerCase() === "erowan") {
-          return "rowan";
+          return "ROWAN";
```
ryardley, line 114: Will this cause problems as these set the route symbol? Maybe it is not as common these days but I would have thought it best to keep everything lowercase and parse on input incase certain browsers and OSs treat url casing differently.

ui/app/src/components/shared/utils.ts
```diff
@@ -27,11 +27,11 @@ export function formatNumber(displayNumber: string) {
...
-  if (symbol === "erowan") return "rowan";
-  return "c" + symbol;
+  if (symbol.toLowerCase() === "erowan") return "ROWAN";
+  return "c" + symbol.toUpperCase();
 }
 export function getUnpeggedSymbol(symbol: string) {
-  if (symbol === "rowan") return "erowan";
+  if (symbol.toLowerCase() === "rowan") return "eROWAN";
   return symbol.indexOf("c") === 0 ? symbol.slice(1) : symbol;
 }
```
ryardley, line 36: hmm... looks like there was an attempt to have these centralized but is was not implemented. Are these actually used anywhere?

ryardley (commented):
I think we need to build a centralized lib for handling the capitalization display of assets but that is out of the scope for quick fixing this stuff around the errors. Another option at a later point could be to have a display symbol on the Asset as well as a symbol. Asset symbols are stored as lowercase throughout the app and it might be more of a hassle to change right now and honestly I don't really see a reason to if we have a good formatter. I am wondering if it makes sense to fix the guard error first in a separate PR and come back to the capitalization?

59023g:
> I am wondering if it makes sense to fix the guard error first in a separate PR and come back to the capitalization?

Yes. I pinged @iverc. If don't hear back in 10 minutes. Let's proceed with your PR. This is guard error is a p0, imo

ui/app/src/views/PegListingPage.vue
59023g: This does effect routing. Not sure if breaking. But does make me nervous
Screen Shot 2021-02-22 at 2 59 35 PM

ui/app/src/components/shared/utils.ts
iverc: I believe no

thomasdavis (approved):

PR #824, opened 2021-02-23
[ui-hotfix] Remove liquidity pool from pool list

ryardley:
Because we are not just blowing away an array anymore and have replaced with a hash remove liquidity was not updating the list of pools. We don't have the option of blowing away the object or that will cause flicker as new pools load so. We need to go through each individually and remove them.


iverc (approved):

59023g (approved):
Tested working on Sandpit

PR #835, opened 2021-02-23
[ui] core/ethBridgeService: use allowance for all erc20, solve "double approve" issue

59023g:
- This solves "double approve" issue, where user would approve, then reject confirm spend, then come back to confirm spend, yet be asked to approve again
- I did not use `max`/`unlimited` spend approval as it's a security risk.
- This uses `allowance` contract method to check if user approved spend for *all* ERC20 tokens. Not just `USDT`. 
- This sets approve spend amount of current `tx`. `allowance` checks against this amount. If it's greater than value `allowance` returns, it'll prompt user to `approve` spend again


ui/core/src/api/EthbridgeService/EthbridgeService.ts
```diff
@@ -52,20 +52,17 @@ export default function createEthbridgeService({
...
-            hasAlreadyApprovedSpend
-          );
-          return;
-        } else {
-        } // else would need to approve for the difference ?
+      // TODO - give interface option to approve unlimited spend via web3.utils.toTwosComplement(-1);
+      const hasAlreadyApprovedSpend = await tokenContract.methods
+        .allowance(account, bridgebankContractAddress)
+        .call();
+      if (hasAlreadyApprovedSpend >= amount.toBaseUnits().toString()) {
```
ryardley, line 59: I think this logic was wrong from the outset. `amount.toBaseUnits().toString()` is a number that is a string. Assuming `hasAlreadyApprovedSpend` is a string number we would probably want to do something like

```ts
amount.toBaseUnitsFr().lessThan(hasAlreadyApprovedSpend);
```

Would it be a good idea to add a unit test here?

ryardley: Also as there is a business logic component to this (dont approve if already approved) It might also make more sense to promote some of this stuff to the usecase layer. What if we just want to use the service to approve some funds outside of this usecase?

59023g: It works in my tests; the number (string) comparison. I used `amount.toBaseUnits().toString()` value because same as in `approve` function.

I don't want to refactor this right now. It could be a separate function. I couldn't think of an instance we'd `approve` without checking if we already had first. I will add a note.

iverc: @59023g nice, we shouldn't check for the token name here :)
For the max approval - could you please elaborate why it is a security risk?
The only option to send money is by getting access to the private key. And if you have access to private key you can approve any amount before sending, so there is really no extra protection by preventing `max` approval, just more inconvenience and more $$ spent for needless transactions.

Many exchanges give you a choice option to either approve selected amount or max. It is very convenient. As a user, I change my mind about the amount multiple times (e.g. price changed while I was approving so I decided to send more). Current implementation will request another approval that would take $30 again and make me wait for 1-2 miniutes more. While waiting, price can change again, so I it is no longer interesting to send that amount. This is a real world case that happened to me multiple times.

59023g: I made a TODO that we should offer the *option* for user to use max spend, but not make the choice for them.

The security risk is that if there's an exploit in our smart contract, for whatever reason, they'll be able to take max amount from all the users who've approved the max amount. Versus, if they only approve what's spent, exploiter will only be able to take approved spend amount. Yes, it's more conservative and is more costly. But I don't want to be responsible for setting max spend (without explicit user consent) and for something to go wrong.

Screenshot of proposed check box:
image

ryardley: > It works in my tests; 

Ahh ok I see we are then relying on javascript implicit type coercion in comparing strings. I guess that works in this case.

59023g: This is another item to be included in the core number overhaul. Alongside the view display number overhaul

iverc (approved):

PR #836, opened 2021-02-23
[ui] core/actions/peg: change peg (rowan -> erowan), unpeg fees

59023g:
Per @ElliotFriedman :
```
The fee for unpegging ethereum native assets will be 0.05856 eth
The fee for pegging rowan to erowan will be 0.05408
```

This PR should reflect those changes.


ryardley (approved):
🚀

PR #857, opened 2021-02-26
[ui] workaround, poll cosmos instead of ws

thomasdavis:


ui/core/src/api/utils/SifClient/TendermintSocketPoll.ts
```diff
@@ -0,0 +1,33 @@
...
+    const blockResponse = await fetch(`${apiUrl}blocks/latest`);
+    const data = await blockResponse.json();
+    const eventType = "NewBlock";
+    emitter.emit(eventType, data);
+  }
+  return {
+    on(event: "Tx" | "NewBlock" | "error", handler: (event: any) => void) {
+      if (!emitter.hasListeners(event)) {
+        console.log('fetching url');
+        this.fetchNewBlocksInterval = setInterval(fetchNewBlocks, 5000)
```
ryardley, line 20: Look like this s breaking the build. Maybe try using a let var in a closure?

ui/core/src/api/utils/SifClient/TendermintSocketPoll.ts
```diff
@@ -0,0 +1,33 @@
...
+>;
+
+export function TendermintSocketPoll({ apiUrl }: { apiUrl: string }) {
+  const emitter = new EventEmitter2();
+
+  const fetchNewBlocks = async () => {
+    const blockResponse = await fetch(`${apiUrl}blocks/latest`);
+    const data = await blockResponse.json();
+    const eventType = "NewBlock";
+    emitter.emit(eventType, data);
```
ryardley, line 14: This needs to emit exactly once for every new block is there something we need to do to guarantee that?

thomasdavis: I can just do some let polling = true etc 

I noticed that the `createTendermintSocketPoll` runs a few times, any idea why?

ryardley: Because a few different client instances are instantiated. Not all listen for events. We could consolidate all polling to a singleton and then respond according to the listeners on the instance?

ryardley: I think we might want to backfetch the height to decouple the polling from the blockgeneration as we have no guarantee they will stay in sync which means we will lose events. 

```
1. first fetch - store current height and process data to emit events
2. subsequent fetches - is new height greater than current height + 1? keep fetching data decrementing requested height and storing in an array in chron order until we have every fetch then process array in chron order to dispatch events
```

ryardley: As far as txs are concerned there is a data key there:

```
data: {
  tsx: null
}
```

If that exists you could emit a Tx event perhaps?

ui/core/src/api/utils/SifClient/TendermintSocketPoll.ts
```diff
@@ -0,0 +1,33 @@
...
+
+export type TendermintSocketPoll = ReturnType<
+  typeof TendermintSocketPoll
+>;
+
+export function TendermintSocketPoll({ apiUrl }: { apiUrl: string }) {
+  const emitter = new EventEmitter2();
+
+  const fetchNewBlocks = async () => {
+    const blockResponse = await fetch(`${apiUrl}blocks/latest`);
```
ryardley, line 11: Other thing I just realised is that we should probably use axios unless we want to install a node-fetch alternative for tests. (Axios is already included as part of the cosmjs lib)

ryardley (approved):

PR #860, opened 2021-02-26
[UI] Fix pool stats

iverc:


59023g (approved):

59023g (changes requested):
Screen Shot 2021-02-26 at 8 39 44 PM
The tooltip margins look off here

iverc:
@59023g no changes to margins were made and this is actually just a rollback of the test label values. Fixing margins, but if we need to deploy  asap - I suggest merging it in as it does not make it worse compared to the existing version.

59023g (approved):

PR #865, opened 2021-02-28
Improve wallets behavior

iverc:
- Get MetaMask data on page load to keep connection state across page reloads.
- Get Keplr data on page load to keep connection state across page reloads.
- TODO: change the UI to remove disconnect buttons and replace them with "change wallet" because disconnect buttons do not actually disconnect anything.


iverc:
After this change, wallets stay connected to the app regardless of page reloads. Also, we do not have disconnect buttons anymore, instead, we have an indication of what is connected.  @ryardley @59023g

Raw version, could be designed better:
Screenshot 2021-02-28 at 16 04 24

caseyturnerarrington:
@iverc please show the entire address number so a user can easily copy+paste it if needed.But other than that, this looks great

59023g (changes requested):
Will you revert the design of the connect wallet modal? 
Screen Shot 2021-03-01 at 4 35 56 PM

And I noticed an issue where Keplr would not disconnect from the UI when I logged out.

ryardley:
I am however struggling to work out how to disconnect keplr. Is that actually possible? Seems like signing in and out doesn't work.

ui/app/src/components/wallet/KeplrWalletPanel.vue
```diff
@@ -1,50 +1,35 @@
...
 import { computed, defineComponent } from "vue";
 import { useCore } from "@/hooks/useCore";
-import BalanceTable from "./BalanceTable.vue";
 import SifButton from "@/components/shared/SifButton.vue";
 import Icon from "@/components/shared/Icon.vue";
 
-function useKeplrWallet() {
-  const { store, actions } = useCore();
-  async function handleDisconnectClicked() {
-    await actions.wallet.disconnectWallet();
```
ryardley, line 11: Should we remove the usecase for this if we are not using it?

iverc: yeah, less code to maintain is better

iverc:
@ryardley yes, there is no callback from Kepler when you disconnect, so can't really reset UI deterministically

59023g:
Screen Shot 2021-03-02 at 1 12 06 PM
Consensus design w @caseyturnerarrington and @iverc

59023g (approved):

ui/core/src/api/SifService/SifService.ts
```diff
@@ -290,11 +295,7 @@ export default function createSifService({
           gas: "300000", // TODO - see if "auto" setting
         };
 
-        const txHash = await client.signAndBroadcast([msg], fee, params.memo);
-
-        triggerUpdate();
-
-        return txHash;
+        return await client.signAndBroadcast([msg], fee, params.memo);
```
59023g, line 298: Why the changes here? Why remove `triggerUpdate()`?

59023g (changes requested):
Questions around changes to `SifService`

ui/core/src/api/SifService/SifService.ts
iverc: @59023g trigger update does not have effect on this operation, and after updates it breaks tests because client is initialised differently for tests and `triggerUpdate` resets it and breaks the test. Removing `triggerUpdate` does not break test and they keep returning same results. I think it was put into these functions to try to update data after the transfer, but since we had block updates subscription (and now there is a stable polling every 2 seconds regardless of block updates) there is no need to call it there.

ui/core/src/api/SifService/SifService.ts
```diff
@@ -129,50 +142,51 @@ export default function createSifService({
...
+      if (!offlineSigner) {
+        client = null;
+        return;
+      }
+      const accounts = await offlineSigner.getAccounts();
+      const address = accounts.length > 0 ? accounts[0].address : "";
+      if (!address) {
+        throw "No address on sif account";
+      }
+      client = new SifClient(
```
ryardley, line 155: This doesn't seem to me as if it really needs to be exposed as public API - would there be benefit from a refactorability perspective to making this simply a pure function? 

offlineSigner -> client

ui/core/src/api/SifService/SifService.ts
```diff
@@ -99,19 +94,37 @@ export default function createSifService({
 
   const triggerUpdate = debounce(
     async () => {
-      if (!client) {
+      try {
+        if (!polling) {
+          polling = setInterval(() => {
```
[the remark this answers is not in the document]
ryardley: Also could this be an opportunity to refactor and remove reactive from SifNode and replace with an event handler?

ryardley (commented):
image

I guess this is in flux but I thought I should highlight I was getting shortening for one address and not another for some reason

ui/core/src/api/SifService/SifService.ts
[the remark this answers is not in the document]
ryardley: In terms of stuff to think about as we refactor all this we probably want to think about our basic template for dealing with blockchains eg. `IBlockchainProvider` --> `IBlockchainService` seems cosmjs provides us with the offlineSigner abstraction we can wrap. I would like to eventually see some kind of provider accepting method on the `BlockchainService`. 

```ts
const provider:IBlockchainProvider = KeplrProvider.create(config)
services.sif.setProvider(provider); 

const provider:IBlockchainProvider = LunieProvider.create(config)
services.sif.setProvider(provider); 
```

ryardley:
I think we still need a mechanism for when I change the user in keplr that the address and account balances change currently this still appears to be the case. Before it was not right but I could disconnect and reconnect. Now I need to refresh the page. Neither of these are very good.

ui/app/src/views/CreatePoolPage.vue
```diff
@@ -282,19 +278,41 @@ export default defineComponent({
...
+            >{{
+              fromSymbol.toLowerCase().includes("rowan")
+                ? fromSymbol.toUpperCase()
+                : "c" + fromSymbol.slice(1).toUpperCase()
+            }}
+            per
+            {{
+              toSymbol.toLowerCase().includes("rowan")
+                ? toSymbol.toUpperCase()
+                : "c" + toSymbol.slice(1).toUpperCase()
```
thomasdavis, line 295: Will get this cleaned up asap next week. 

reminder to change all the `from` `to` semantics.

thomasdavis (approved):

ui/core/src/api/SifService/SifService.ts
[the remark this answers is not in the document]
iverc: @ryardley
1) shortening - was fixed in one of the commits, probably need to clear cache
2) if there was a direct indication from keplr when you disconnect or change address in the extension - I wouldn't even need neither polling nor block subscription. actually this is exactly the same approach as you had implemented for metamask before I change it to events, cause metamask can notify about such things.

ui/core/src/api/SifService/SifService.ts
iverc: I made a refactoring to remove offline signer from exposed vars, but the client should be exposed, at least with current setup. tests are written in such way that they initialise client from mnemonic and then set this exposed client var to it. Can be refactored for sure, but will take more time and I really want to push it asap so it gets into deployment

ui/app/src/App.vue
```diff
@@ -10,7 +10,7 @@
           </template>
           <template v-slot:connected="{ connectedText, requestDialog }">
             <SifButton small ghost success round @click="requestDialog">
```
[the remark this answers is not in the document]
iverc: nice catch thanks

ui/app/src/views/CreatePoolPage.vue
iverc: cool!

caseyturnerarrington:
@iverc seeing some weirdness in trying to connect my Keplr Wallet:

https://user-images.githubusercontent.com/45978818/110025967-0e657680-7cfe-11eb-854c-aaf60b733e92.mov

iverc:
@59023g could I get an approval here please? need to merge it in so it gets into the new deployment
@caseyturnerarrington

caseyturnerarrington (approved):

59023g (approved):

PR #867, opened 2021-03-01
[ui] prettier wip 

thomasdavis:
Just experimenting.

- [ ] add docs for people to up prettier in their editors
- [ ] finalize any deviations from defaults (i do like single quotes over double ¯\_(ツ)_/¯
- [ ] make sure all ts, js, vue files are converted 

Will post the commands to do the full transform here later. 

We likely won't do this until there are few PR's in the pipeline.


thomasdavis:
```
prettier --write "**/*.js"
prettier --write "**/*.ts"
prettier --write "**/*.vue"

```

PR #868, opened 2021-03-01
[peggy] structured logging

juniuszhou:
use structured logging with json format to make it easier for Kibana analysis.


cmd/ebrelayer/main.go
```diff
@@ -155,18 +155,27 @@ func RunInitRelayerCmd(cmd *cobra.Command, args []string) error {
 	validatorMoniker := args[3]
 	mnemonic := args[4]
 
-	// Universal logger
-	logger := tmLog.NewTMLogger(tmLog.NewSyncWriter(os.Stdout))
+	logger, err := zap.NewProduction()
```
banshee, line 158: Is this the configuration we want?  It enables stack traces and sampling, and I suspect we don't want either of those in production.

cmd/ebrelayer/relayer/cosmos.go
```diff
@@ -57,15 +57,16 @@ func (sub CosmosSub) Start(completionEvent *sync.WaitGroup) {
 	time.Sleep(time.Second)
 	client, err := tmClient.New(sub.TmProvider, "/websocket")
 	if err != nil {
-		sub.Logger.Error("failed to initialize a client", "err", err)
+		sub.SugaredLogger.Errorw("failed to initialize a sifchain client.",
+			"error message", err.Error())
```
banshee, line 61: We should extract out the "error message" and make it a constant so every time we call the logger we have consistent field names.  (And I think we shouldn't put spaces in these keys)

cmd/ebrelayer/relayer/cosmos.go
```diff
@@ -239,17 +245,16 @@ func MessageProcessed(message types.CosmosMsg, prophecyClaims []types.ProphecyCl
 func (sub CosmosSub) Replay(fromBlock int64, toBlock int64, ethFromBlock int64, ethToBlock int64) {
 	ProphecyClaims := sub.getAllProphecyClaim(ethFromBlock, ethToBlock)
 
-	fmt.Printf("found out %d prophecy claims I sent from %d to %d block", len(ProphecyClaims), ethFromBlock, ethToBlock)
+	log.Printf("found out %d prophecy claims I sent from %d to %d block\n", len(ProphecyClaims), ethFromBlock, ethToBlock)
```
banshee, line 248: Why not use the new logger here?

cmd/ebrelayer/relayer/cosmos.go
```diff
@@ -309,11 +314,17 @@ func getOracleClaimType(eventType string) types.Event {
 
 // Parses event data from the msg, event, builds a new ProphecyClaim, and relays it to Ethereum
 func (sub CosmosSub) handleBurnLockMsg(cosmosMsg types.CosmosMsg, claimType types.Event) {
-	sub.Logger.Info(cosmosMsg.String())
+	sub.SugaredLogger.Infow("handle burn lock message.",
+		"cosmos message is", cosmosMsg.String())
```
banshee, line 318: json is better if the field names ("cosmos message is") don't have spaces

cmd/ebrelayer/relayer/cosmos_test.go
```diff
@@ -1,16 +1,18 @@
...
 	"testing"
 
 	"github.com/Sifchain/sifnode/cmd/ebrelayer/txs"
 	"github.com/Sifchain/sifnode/cmd/ebrelayer/types"
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/crypto"
 	"github.com/stretchr/testify/require"
-	"github.com/tendermint/tendermint/libs/log"
+
+	// "github.com/tendermint/tendermint/libs/log"
```
banshee, line 14: Delete the commented out code?

cmd/ebrelayer/main.go
banshee: FYI, log messages with stack traces look like this, so maybe it's not so bad:

`{"level":"error","ts":1614707042.5673304,"caller":"relayer/ethereum.go:212","msg":"failed to get the last ethereum block from level db.","error message":"leveldb: not found","stacktrace":"github.com/Sifchain/sifnode/cmd/ebrelayer/relayer.EthereumSub.Start\n\t/home/james/workspace/prsifnode/cmd/ebrelayer/relayer/ethereum.go:212"}
`

banshee, line 158: We should replace the standard logger with the uber logger -

`zap.RedirectStdLog(sub.SugaredLogger.Desugar());
`

Having a mix of logging styles in the output isn't great.

banshee (changes requested):

ElliotFriedman:
Why do we need to make these changes to our logging setup? Will it help us when we search logs in kibana or data dog?

cmd/ebrelayer/main.go
juniuszhou: yes, if there is an error in process. the stacktrace can help us for debug.

cmd/ebrelayer/relayer/cosmos.go
juniuszhou: accepted.

cmd/ebrelayer/relayer/cosmos.go
juniuszhou: for the replay scripts, the log not collected and analyzed by tools. so json format not needed. it is better for people to read in console.

banshee:
> Why do we need to make these changes to our logging setup? Will it help us when we search logs in kibana or data dog?

Yes, that's precisely it.  Kibana gives you tools to look at logging and it knows how to destructure json logs.

cmd/ebrelayer/relayer/cosmos.go
juniuszhou: accepted.

cmd/ebrelayer/main.go
banshee: We need to make sure sampling is turned off though - we want all logs.

banshee (changes requested):

cmd/ebrelayer/txs/relayToCosmos.go
```diff
@@ -13,27 +13,32 @@ import (
...
+		"next sequence number", nextSequenceNumber)
 
 	for _, claim := range claims {
 		// Packages the claim as a Tendermint message
 		msg := ethbridge.NewMsgCreateEthBridgeClaim(claim)
 
 		err := msg.ValidateBasic()
 		if err != nil {
-			log.Println("failed to get message from claim with:", err.Error())
+			sugaredLogger.Errorw("failed to get message from claim.",
```
ElliotFriedman, line 40: How does this print out when you print the errorMessageKey and then you print the error afterwards?

Do errorMessageKey and error get concatenated into a single value and then `failed to get message from claim.` is the key to the object that is printed out?

ElliotFriedman: Seems like my answer is here https://pkg.go.dev/go.uber.org/zap#SugaredLogger.Errorw

There is a string, then keys and values are passed as variadic args so this should print correctly.

cmd/ebrelayer/txs/relayToCosmos.go
```diff
@@ -43,49 +48,55 @@ func RelayToCosmos(cdc *codec.Codec, moniker, password string, claims []types.Et
...
 
 	// start to control sequence number after first successful tx
 	if nextSequenceNumber == 0 {
 		setNextSequenceNumber(txBldr.Sequence() + 1)
 	} else {
 		incrementNextSequenceNumber()
 	}
-	log.Printf("ebrelayer RelayToCosmos nextSequenceNumber is %d after tx\n", nextSequenceNumber)
+	sugaredLogger.Infow("relay next sequenceNumber from memory.",
+		"next sequence number", nextSequenceNumber)
```
ElliotFriedman, line 98: It would be great if instead of having a key in the JSON logs that is `next sequence number`, it is a single word key like this `nextSequenceNumber`

cmd/ebrelayer/txs/relayToEthereum.go
```diff
@@ -25,107 +25,117 @@ const (
...
+			errorMessageKey, err.Error())
 		return nil, nil, common.Address{}, err
 	}
 
 	// rate limit the bridge so that nonce is handled correctly
 	time.Sleep(transactionInterval)
 
 	nonce, err := client.PendingNonceAt(context.Background(), sender)
-	log.Println("Current eth operator at pending nonce: ", nonce)
+	sugaredLogger.Infow("Current eth operator at pending nonce.", "pending nonce", nonce)
```
ElliotFriedman, line 101: same comment here

cmd/ebrelayer/txs/relayToEthereum.go
```diff
@@ -25,107 +25,117 @@ const (
...
 
 	quarterGasPrice := big.NewInt(0)
 	quarterGasPrice = quarterGasPrice.Div(gasPrice, big.NewInt(4))
-	log.Println("quarterGasPrice: ", quarterGasPrice)
-	log.Println("gasPrice after: ", gasPrice)
 
 	gasPrice.Sub(gasPrice, quarterGasPrice)
-	log.Println("suggested gas price after subtracting 1/4: ", gasPrice)
+	sugaredLogger.Infow("final gas price after adjustment.",
+		"final gas price", gasPrice)
```
ElliotFriedman, line 130: Same comment as above

x/ethbridge/handler.go
```diff
@@ -87,14 +114,17 @@ func handleMsgCreateEthBridgeClaim(
 
 func handleMsgBurn(
 	ctx sdk.Context, cdc *codec.Codec, accountKeeper types.AccountKeeper,
-	bridgeKeeper Keeper, msg MsgBurn,
+	bridgeKeeper Keeper, msg MsgBurn, sugaredLogger *zap.SugaredLogger,
 ) (*sdk.Result, error) {
 	if !bridgeKeeper.ExistsPeggyToken(ctx, msg.Symbol) {
+		sugaredLogger.Errorw("Native token can't be burn.",
+			"token symbol", msg.Symbol)
```
ElliotFriedman, line 121: Let's have `invalidTokenSymbolForBurn` be the key for this error message instead of "token symbol".

x/oracle/keeper/keeper.go
```diff
@@ -86,24 +86,23 @@ func (k Keeper) ProcessClaim(ctx sdk.Context, claim types.Claim) (types.Status,
...
 	if !inWhiteList {
-		fmt.Println("sifnode oracle keeper ProcessClaim validator no in whitelist")
+		sugaredLogger.Errorw("sifnode oracle keeper ProcessClaim validator no in whitelist.")
 		return types.Status{}, types.ErrValidatorNotInWhiteList
 	}
 
 	activeValidator := k.checkActiveValidator(ctx, claim.ValidatorAddress)
 	if !activeValidator {
-		fmt.Println("sifnode oracle keeper ProcessClaim validator not active")
+		sugaredLogger.Errorw("sifnode oracle keeper ProcessClaim validator no active.")
```
ElliotFriedman, line 95: error message should be:
"sifnode oracle keeper ProcessClaim validator not active"

ElliotFriedman (changes requested):

ElliotFriedman (approved):

x/ethbridge/handler.go
```diff
@@ -39,26 +56,36 @@ func NewHandler(
...
+			sugaredLogger.Errorw("bridge keeper failed to process successful claim.",
+				errorMessageKey, err.Error())
 			return nil, err
 		}
 	}
 	// set mutex lock to false
 
-	fmt.Printf("Sifnode handleMsgCreateEthBridgeClaim 56 all done, emit events statue is %s\n", status.Text.String())
+	sugaredLogger.Infow("sifnode emit create event.",
+		"CosmosSender", msg.ValidatorAddress.String(),
```
banshee, line 81: Just log the whole msg?  ("msg", msg) should work

banshee (approved):

marbar3778:
This pr should be reverted. Why are you creating multiple loggers in each module and not passing a single logger from the app.go level?

PR #876, opened 2021-03-02
[ChainOps] Testnet updates.

intl-man:
Includes:

* Documentation updates for testnet.
* Unified rake task for booting a standalone testnet and mainnet node (via docker compose).


intl-man:
> Do we need to specify - that rowan is calculated like Wei i.e. 1e18?

@alijnmerchant21 Good idea. Have added.

alijnmerchant21 (approved):

PR #877, opened 2021-03-02
🚧 [ui - no merge] Cypress wip

ryardley:
Issue https://github.com/sifchain/issues/issues/4

One example of how to setup cypress for our frontend testing.

Ideas

* This is setup for frontend testing against a local environment not integration testing we should be able to migrate tests from this sort of setup to an integration test

TODO
- [ ] Work out how to mock metamask/keplr and at what stage to mock metamask/keplr 
- [ ] Write an actual test


ryardley:
So it looks like Cypress wont yet support browser specific protocols https://github.com/cypress-io/cypress/issues/1965 and using it to test extensions like metamask within the context of an application appears to be fraught with complexity. 

Also it looks like extensions cannot be loaded in headless chrome from https://docs.cypress.io/api/plugins/browser-launch-api.html#Syntax 

```
// NOTE: extensions cannot be loaded in headless Chrome
```

PR #885, opened 2021-03-03
[ui] Amount API Proposal

ryardley:
### [Display API Proposal Here](https://github.com/Sifchain/sifnode/blob/amount-adr/ui/docs/decisions/adr-display.md)


thomasdavis:
Where do you think the lib will live? 
I am trying to look at mine in two ways, one as a function and the other as a component (which just uses the former function)

ryardley:
> I am trying to look at mine in two ways, one as a function and the other as a component (which just uses the former function)

Yeah that sounds great. I would think the formatting API function part that converts `Amount | AssetAmount -> string` should actually be a helper in core that is exported as it is a tight dependency and will change together. If we share core with some other implementation we will still need the display functions but not necessarily the vue components.

59023g:
I think it should be designed to be tightly coupled with @ryardley's core (1st priority) but ultimately I see it as its own npm package others could use.

59023g:
Looks great. Concise. And easy to understand.

My minor concerns revolve around overlap with future display library spec. 

- Why include `toBigInt()` and `toString()` ? - I think those would be display lib methods.
- Asset item #2 - It's nice but to be pure, I'd think display lib would do the capitalization
- Ultimately @thomasdavis and you should make decision on these items ^^

Are the todo items implementation? Looks like it

ryardley:
> Why include toBigInt() and toString() ? - I think those would be display lib methods.

These are to be consumed by the display lib. We need a way to get that information to the display lib and we also need a way to get values for testing.

ryardley:
>  I'd think display lib would do the capitalization

1. We don't know what the rules are necessarily for the mapping between symbol and symbol label. 
2. We might have a situation where we need to deal with assets on different blockchains with the same symbol in which case we need to add something arbitrary to the symbol to differentiate it.
3. Considering `symbol` like one would a unique identifier and then storing `label` as how we expect that symbol to be displayed is the most flexible way to handle any future requirements.

ryardley:
> Are the todo items implementation? Looks like it

Yes

59023g (dismissed):

ui/decisions/adr-display.md
```diff
@@ -0,0 +1,46 @@
+// import DisplayAmount from "@core/DisplayAmount";
+
+interface Asset {} // Re-used
+interface Amount {} // Re-used - Rudi implmentation
+
+enum Type {
```
ryardley, line 6: I am wondering is `Type` too generic for what we need here?

ryardley:
So another question: outside of testing would we ever need to actually display the asset symbol label? 

Seems like we might get away with using `asset.label` or `amount.label` and then just format the number part of the asset amount.

That means we only need a single function that takes either an Amount or an AssetAmount and displays a numeric value

ryardley:
So updated the display API here. Ready for review @59023g @thomasdavis

59023g (dismissed):
🚀 🧠

59023g (approved):

59023g (dismissed):

59023g (approved):

PR #886, opened 2021-03-03
[ChainOps] EKS Autoscaling/volume snapshots

alexhui01:
* add auto-scaler 
* add volumes back for all k8s presisent volumes


alexhui01:
@intl-man  Do we need to adjust some changes for `node_group` name here?  Or we can add a `default` node_group_name`.

intl-man (approved):
Nice work @alexhui01.

PR #892, opened 2021-03-04
[ui] EventBusService

ryardley:
Untitled Diagram (5)

This is an example of a decoupled notifications service based on an event emitter pattern.

* `EventBusService` is shared with the view
* Core dispatches events on the EventBusService
* View listens to those events and creates the appropriate view notification. 

Why is this a good idea?
* Core knows nothing about the messaging that the view uses to display the message. This will make things like localization easier.
* Google Analytics which is an implementation concern can get realtime information on what is happening within core by simply adding a listener to the NotificationsService. 
* Other implementation concerns such as logging and error reporting can listen to events driven from the EventBusService and marshall the appropriate response to their respective targets
* Events are typed so the event payload shape is matched to the event type
* Cohesion occurs through the shared definition of events which in this example happens [here](https://github.com/Sifchain/sifnode/pull/892/files#diff-1181e517a11ffd75848b4d3e55ccdaf88bf27ec325aa9c9ec5218d472e9d92e4R7) we might want to move them elsewhere.

Possible other things we could do
- [x] We could rename `NotificationsService` to be something more general such as `EventBusService` to be clearer 
- [x] We could rename `notify` to `dispatch` to be clearer.
- [ ] After we have renamed the structures within core this might look more like: `services.bus.dispatch(myevent)`

Possible other things we might be careful of doing
* We must not define events within core that are view specific events. Core should not know or care about events that are view specific. Instead we could trigger those events directly in code (eg `ga('send', 'pageview');`) but also avoid those events unless they are 100% necessary.


59023g:
👀

ui/app/src/components/Notifications.vue
```diff
@@ -1,18 +1,102 @@
...
+  message: string;
+  detail?: {
+    type: "etherscan" | "websocket" | "info";
+    message: string;
+  };
+  loader?: boolean;
+};
+
+// Visual Notifications are a view level system here we work out which ones are displayed to the user
+function parseEventToNotifications(event: AppEvent): Notification | null {
```
59023g, line 20: This is cool. Should also give us more control how we want to display dispatch. For example, use the default popup in the corner or some other UI notification (like red text under an input).

ui/core/src/actions/clp/index.ts
```diff
@@ -70,13 +69,9 @@ export default ({
...
     effect(() => {
       if (Object.keys(store.pools).length === 0) {
-        notify({
-          type: "error",
-          message: "No Liquidity Pools Found",
-          detail: {
-            type: "info",
-            message: "Create liquidity pool to swap.",
-          },
+        api.NotificationService.notify({
```
59023g, line 72: 😎

ui/core/src/api/NotificationService/NotificationService.ts
```diff
@@ -0,0 +1,109 @@
...
+  | PegTransactionErrorEvent;
+
+export type EventHandler = (event: AppEvent) => void;
+export type AppEventType = AppEvent["type"];
+export type AppEventTypes = AppEventType[];
+
+export type NotificationServiceContext = {};
+
+export default function createNotificationsService({}: NotificationServiceContext) {
+  const emitter = new EventEmitter2();
```
59023g, line 78: Same dep, nice

59023g (commented):
Looking good

ui/app/src/components/Notifications.vue
ryardley: Exactly!!!

ui/core/src/actions/ethWallet.test.ts
```diff
@@ -47,6 +57,7 @@ beforeEach(() => {
 test("Calls disconnect", async () => {
   await ethWalletActions.disconnectWallet();
   expect(mockEthereumService.disconnect).toHaveBeenCalled();
+  expect(notify).toHaveBeenCalled();
```
ryardley, line 60: I just want to point out that by having notify in a service it is easier to test that the notification happened (something we will be more and more concerned with moving forward)

ryardley:
So I fixed up this to surface the Metamask not found error. I think this could be merged if folks are happy with the direction.

59023g (approved):
🥇

PR #896, opened 2021-03-05
[ui] showing a warning during pooling

thomasdavis:


59023g:
From a user standpoint, seems like something is missing.. Like a color coded string message that matches the outline. And/or color coded tooltip... as a user I don't know the context. Why is this box red or yellow? cc @caseyturnerarrington

thomasdavis:
I still gotta make an ! icon too

59023g (approved):

ui/core/src/entities/formulae.ts
```diff
@@ -1,6 +1,31 @@
 import Big from "big.js";
 import { Fraction, IFraction } from "./fraction/Fraction";
 
+export function slipAdjustment(
```
59023g, line 4: What's this?

59023g (approved):
Looks good.  What're the formulae changes? Prettier?

PR #906, opened 2021-03-06
Feature/upgrade pool units export

Jedi2002:


iverc (approved):

app/app.go
```diff
@@ -245,6 +247,30 @@ func NewInitApp(
 
 	app.UpgradeKeeper.SetUpgradeHandler("changePoolFormula", func(ctx sdk.Context, plan upgrade.Plan) {
 		ctx.Logger().Info("Starting to execute upgrade plan for pool re-balance")
+		appState, vallist, err := app.ExportAppStateAndValidators(true, []string{})
```
utx0, line 250: Can we turn this code into its own function and then out put the files as something like `changePoolFormula=-*.json`

utx0 (changes requested):

PR #908, opened 2021-03-06
Fix keplr update issue

iverc:


iverc:
Same change added in #909

PR #919, opened 2021-03-08
proto: setup proto generation and ci

marbar3778:


iverc (approved):
Some minor cleanup could be done, but looks great as is!

iverc:
@marbar3778 please fix the `.proto` file issue

marbar3778:
> @marbar3778 please fix the `.proto` file issue

which issue are you talking about? `Protobuf / breakage (pull_request) ` fails because there are no proto files on master. 

Oh also what is the development flow here? Should it be checking against master or develop?

PR #930, opened 2021-03-10
[Peggy] Smart Contracts 2.0 Rough Cut

ElliotFriedman:
This is a first draft of what the future of peggy's smart contracts may look like. No whitelists, no limits on token transfers, no locking eth in the bridge, simplified cosmosbridge contracts and everything overall looking and feeling much better.

Lots to be done just on the smart contract side. Many edge cases are not worked out here and this has not been hooked into the relayer or sifnode.

Still lots of design decisions to be made around accepting a chain's native currency.

No tests have been fixed as this will likely break most integration and e2e flows.


ElliotFriedman:
There should be gas savings from these changes on both lock and unlock transactions as we no longer have to store the amount of a token that we own in the smart contract.

These savings will likely be marginal, however, if we can cut down on more smart contract calls or condense logic even more, it should be able to cut back on gas costs for end users as a byproduct.

pandaring2you:
covered by new peggy 2.0 spec tasks

ElliotFriedman:
I'm going to continue playing around with the smart contracts on this branch and so I will leave this PR open for the time being.

ElliotFriedman:
Note:

We really should upgrade to completely new smart contracts after this upgrade as we are moving around a lot of different pieces and getting the auditing manpower to actually confirm that no storage offsets or things at a really low level didn't change on us might be too time intensive.

ElliotFriedman:
> Note:
> 
> We really should upgrade to completely new smart contracts after this upgrade as we are moving around a lot of different pieces and getting the auditing manpower to actually confirm that no storage offsets or things at a really low level didn't change on us might be too time intensive.

Update: Talked with openzepplin team and the storage variable shadowing we did is not ideal, but it shouldn't prevent us from doing an upgrade. Also, since we are migrating to solidity version 0.8.0, I thought this could cause issues in terms of storage layout, however, apparently the solidity team will not change storage layouts on compiler upgrades as that would be a breaking change.

ElliotFriedman:
Merged to feature/peggy42, closing.

PR #931, opened 2021-03-10
[ChainOps] State export

intl-man:
Includes:

* Minor tweak to store the state export in `$HOME/.sifnoded` instead of `/`.


Jedi2002 (commented):
Not related to the work done in this pr but I noticed .ExportAppState does not return an error . I think its always better design to handle an error at the outermost level rather than inside a low level function .

AustinoBombino (approved):

PR #934, opened 2021-03-10
added new AnteHandler to make tx free for request faucet

Jedi2002:
Added a antehandler to use for TestNet 

Potential risks .
- Any tx which is in the same block as the request_faucet tx  , and comes after it in order , Would not be charged any transaction fee. This is for testnet only .  
Reason : I have only added a custom handler for faucet , to handle the above case we would need to add individual handlers for every module and handle fee seperatly .

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -489,13 +489,3 @@ func ExportAppState(name string, app *SifchainApp, ctx sdk.Context) {
 		ctx.Logger().Error(fmt.Sprintf("failed to write Validator List to file: %s", err.Error()))
 	}
 }
-
-//rake keys:generate:mnemonic
-//account already picnic earn curious gift dance wood assist arch sugar nurse clarify right tuition winter pilot escape final van educate gain insane they
-//{
-//"name": "my-node",
-//"type": "local",
-//"address": "sif12dzcwupljmzhhn88q000a9dx59e83kclm5n6rt",
-//"pubkey": "sifpub1addwnpepqftut3cfz5hr9lysfuv79jrwaftr6ljsnp6gv3z6zuan64klz8f3khkj9f7"
-//}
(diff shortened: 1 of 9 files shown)


app/app.go
```diff
@@ -368,10 +369,25 @@ func NewInitApp(
...
-			app.AccountKeeper,
-			app.SupplyKeeper,
-			auth.DefaultSigVerificationGasConsumer,
+		//auth.NewAnteHandler(
+		//	app.AccountKeeper,
+		//	app.SupplyKeeper,
+		//	auth.DefaultSigVerificationGasConsumer,
+		//),
+		sdk.ChainAnteDecorators(
+			faucet.NewRemoveFacuetFeeDecorator(),
```
khdegraaf, line 378: NewRemoveFacuetFeeDecorator -> NewRemoveFaucetFeeDecorator

spelling...

x/faucet/types/ante.go
```diff
@@ -0,0 +1,31 @@
...
+import (
+	"github.com/Sifchain/sifnode/x/clp"
+	"github.com/cosmos/cosmos-sdk/types"
+	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
+	"github.com/cosmos/cosmos-sdk/x/auth/ante"
+)
+
+type RemoveFaucetFeeDecorator struct{}
+
+func NewRemoveFacuetFeeDecorator() RemoveFaucetFeeDecorator {
```
khdegraaf, line 12: NewRemoveFacuetFeeDecorator -> NewRemoveFaucetFeeDecorator

spelling...

khdegraaf (changes requested):

Jedi2002:
Closing this pr , THe other PR for TX-FEE 2.0 has a better implementation for this

PR #939, opened 2021-03-11
proto: migration from amino to proto

marbar3778:
## Description 

- Write proto definitions and migrate oracle module. 
- This also updates many imports because the sdk moved many things around


marbar3778:
im splitting this PR into multiple, thus closing it.

PR #942, opened 2021-03-12
[ui] Implementation of Amount API

ryardley:
First step of https://github.com/Sifchain/issues/issues/133

Implements:
* Amount API
* format API

Just gets it in and working. Doesn't really tidy up too much in terms of vue code. In further PRs we should move amounts closer to the view. 

- [x] Create new `Amount` wrapper - implement all Fraction methods - write tests
- [x] Create new `AssetAmount` wrapper - write tests
- [x] Add `label` on `Asset`
- [x] Create display lib to handle `toFixed` `toSignificant` etc. -
- [x] Remove `Coin` and `Token` distinction they are not helpful and make `address` non optional on `Asset` initializing with `0x00000...` where appropriate
- [x] Get test tables working with `Amount`
- [x] Get all unit tests for entities working
- [x] Get the rest of the unit tests working
- [x] Replace `Fraction` with `Amount` throughout codebase
- [x] Replace `AssetAmount` with new `AssetAmount` throughout codebase
- [x] Move all application internal representations to baseunits - just need to check this
- [x] Fix up Vue code to use `Amount` and `AssetAmount` (Unfortunately because we are not using JSX this is difficult to have confidence here as we loose types and there are a hell of a lot of formatting hacks)


ryardley:
So marking this tentatively as ready for review. I am uncomfortable submitting this without e2e tests testing all the numbers for our major flows. I am hoping we can setup a test on develop and then port it to this branch?

ui/app/src/components/shared/AssetList.vue
```diff
@@ -3,11 +3,11 @@
     <div class="line" v-for="item in items" :key="item.asset.symbol">
       <AssetItem class="token" :symbol="item.asset.symbol" />
       <div class="amount">
-        {{ formatNumber(item.amount) }}
+        {{ formatAssetAmount(item.amount) }}
```
59023g, line 6: Is this intentional? We agreed not to use this function, right?

ryardley: This is so this branch is low touch. We are using the format function under the hood there.

ui/app/src/views/PegAssetPage.vue
```diff
@@ -235,9 +253,9 @@ export default defineComponent({
         }"
         :rows="[
           {
-            show: !!feeAmount,
+            show: !!feeDisplayAmount,
             label: 'Transaction Fee',
-            data: `${feeAmount.toFixed(8)} cETH`,
+            data: `${feeDisplayAmount} cETH`,
```
59023g, line 258: 👍

ui/core/src/entities/AssetAmount.ts
```diff
@@ -1,189 +1,136 @@
...
-
-  return new _AssetAmount(asset, jsbiAmount);
+  type _IAssetAmount = _ExposeInternal<IAssetAmount>;
+
+  const _asset = (asset as IAssetAmount)?.asset || Asset(asset);
+  const _amount = (amount as IAssetAmount)?.amount || Amount(amount);
+
+  // Proxy all methods because it is clearer and
+  // more explicit than prototypal inheritence
+  const instance: _IAssetAmount = {
```
59023g, line 37: nice

ui/core/src/entities/formulae.calculateExternalExternalSwapResult.test.ts
```diff
@@ -1,27 +1,18 @@
...
-  const bigaY = B(aY);
-  const bigbX = B(bX);
-  const bigbY = B(bY);
   test(`Swapping ${ax}, expecting ${expected}`, () => {
     const output = calculateExternalExternalSwapResult(
       // External -> Native pool
-      new Fraction(bigax, JSBI.exponentiate(TEN, JSBI.BigInt(18))), // Swap Amount
-      new Fraction(bigaX, JSBI.exponentiate(TEN, JSBI.BigInt(18))), // External Balance
-      new Fraction(bigaY, JSBI.exponentiate(TEN, JSBI.BigInt(18))), // Native Balance
+      Amount(ax), // Swap Amount
```
59023g, line 9: so clean

ui/core/src/utils/decimalShift.ts
```diff
@@ -0,0 +1,49 @@
+/**
+ * Function to shift the magnitude of a string without using any Math libs
```
59023g, line 2: This would help, right now

59023g (dismissed):
Will `format` throw gracefully? If so, Let's go!

ryardley:
> Will format throw gracefully? 

What would that look like?

59023g:
> > Will format throw gracefully?
> 
> What would that look like?

It doesn't crash the app. For example, returns `ERR` string. Awhile back `toFixed` would crash the app with the `invariant?` lib if something was amiss. Which was overkill for DOM related error

ryardley:
> > > Will format throw gracefully?
> > 
> > 
> > What would that look like?
> 
> It doesn't crash the app. For example, returns `ERR` string. Awhile back `toFixed` would crash the app with the `invariant?` lib if something was amiss. Which was overkill for DOM related error

Ok I added this:

https://github.com/Sifchain/sifnode/pull/942/files#diff-fbd2d097138272fc451481cbc5336c4f9cad3d72d11e97513e3e1c0c63f65b6fR147

ryardley:
image

hmm flakey tests! :/

59023g:
Strange. I'm not able to pass the tests locally.

ryardley:
> Strange. I'm not able to pass the tests locally. EDIT: I get the same error as CI.

👿

59023g (approved):
🎸

PR #946, opened 2021-03-13
[ui] Run prettier over the codebase

thomasdavis:
- Seems harmless, app builds, and tested a few things, everything seems fine. 
- Some dist files got run over but there temp and should also be fine. 
- I think everyone knows how to switch on prettier


ryardley:
Will this fail the build if you commit a file that isn't run over with prettier?

59023g:
> Nice work! Will this fail the build if you commit a file that isn't run over with prettier?

I second this. Pre-commit hook?

.github/workflows/prettier.yml
```diff
@@ -0,0 +1,20 @@
...
+      - name: Set up Node.js
+        uses: actions/setup-node@v1
+        with:
+        node-version: 12
+      - name: Change to UI folder
+        run: cd ui
+      # ESLint and Prettier must be in `package.json`
+      - name: Install Node.js dependencies
+        run: yarn
+      - name: Run linters
```
ryardley, line 19: Is it a waste to make a whole flow for this? Running prettier is pretty quick isn't it? Can't we just add it to `yarn build` in the ui?

ryardley: Also this is only running it on .js files

59023g: Yea. We should just run prettier before `yarn build`

ryardley:
Is the config this is working off in here? I can't seem to find it.

59023g (approved):

PR #947, opened 2021-03-13
Feature/tx fees 2.0

Jedi2002:
## **Transaction Fees 2.0**
 (**not** "first swap on us")

**HOW IT WORKS:**
- First we create a signed doc and sign a transaction on the user’s behalf (e.g. Alice). Sign transaction forces the regular Keplr popup to appear, and Alice is prompted to select the transaction fee and gas amount as usual. The resulting transaction fee is exactly what Alice is going to pay when the transaction takes place (e.g. is actually broadcasted).
- Right after that (behind the scenes), the broadcast transaction takes place. It uses a previously signed transaction as an input. At this point Alice still does not have ROWAN to pay the transaction fee and this is when the BE takes it over and makes sure that Alice has enough ROWAN before the transaction is delivered. 
- BE handles various edge cases like if Alice has 0 ROWAN or 50% of the required amount, or has more than enough ROWAN to pay transaction fees. 

- Alice will pay exactly what was selected in the Keplr popup in the first step when signing the transaction and BE adds some ROWAN to the existing amount (if any) so the sum is exactly what is required for tx. We do not cover tx fees for free. Instead, BE subtracts the corresponding amount from the input token, hence Alice receives a bit less of the output token.


iverc (approved):

khdegraaf (changes requested):
Tests are still failing, but integration tests aren't always running in each push, in particular if you are only editing UI code.

khdegraaf (approved):
Integration tests now pass, so no issues there anymore...

iverc (dismissed):

app/upgrade-poolChange.go
```diff
@@ -0,0 +1,96 @@
...
+import (
+	"encoding/json"
+	"fmt"
+	"github.com/Sifchain/sifnode/x/clp"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/cosmos/cosmos-sdk/x/upgrade"
+	"io/ioutil"
+)
+
+func GetPoolChangeFunc(app *SifchainApp) func(ctx sdk.Context, plan upgrade.Plan) {
```
Jedi2002, line 12: Just changed the location ,the code in this file is old ,and already reviewd

ui/core/src/entities/formulae.calculateDoubleSwapResult.test.ts
```diff
@@ -0,0 +1,28 @@
+import { calculateExternalExternalSwapResult } from "./formulae";
```
Jedi2002, line 1: UI code ,not relevent to the backend implementation

ui/core/src/entities/formulae.calculateReverseSwapResult.test.ts
```diff
@@ -0,0 +1,16 @@
+import { Fraction, TEN } from "./fraction/Fraction";
```
Jedi2002, line 1: UI code ,not relevent to the backend implementation

app/ante.go
```diff
@@ -0,0 +1,77 @@
...
+
+import (
+	"github.com/Sifchain/sifnode/x/clp"
+	"github.com/Sifchain/sifnode/x/faucet"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/cosmos/cosmos-sdk/x/auth"
+	"github.com/cosmos/cosmos-sdk/x/auth/ante"
+	"github.com/cosmos/cosmos-sdk/x/auth/types"
+)
+
```
alexanderbez, line 11: nit: I would add godocs to these ante-handler functions just so the reader knows the difference between each. They don't need to be too verbose.

app/upgrade-poolChange.go
```diff
@@ -0,0 +1,96 @@
...
+			if !app.clpKeeper.ValidatePool(pool) {
+				hasError = true
+				ctx.Logger().Error(fmt.Sprintf("Invalid | Pool %s ", pool.String()))
+				break
+			}
+			poolList = append(poolList, pool)
+		}
+		// If we have error dont set state
+		if hasError {
+			ctx.Logger().Error("Failed to execute upgrade plan for pool re-balance")
```
alexanderbez, line 58: Just return here and then you can avoid the `!hasError` check below.

app/upgrade-poolChange.go
```diff
@@ -0,0 +1,96 @@
...
+		poolList := clp.Pools{}
+		hasError := false
+		for _, pool := range allPools {
+			lpList := app.clpKeeper.GetLiquidityProvidersForAsset(ctx, pool.ExternalAsset)
+			temp := sdk.ZeroUint()
+			tempExternal := sdk.ZeroUint()
+			tempNative := sdk.ZeroUint()
+			for _, lp := range lpList {
+				withdrawNativeAssetAmount, withdrawExternalAssetAmount, _, _ := clp.CalculateWithdrawal(pool.PoolUnits, pool.NativeAssetBalance.String(),
+					pool.ExternalAssetBalance.String(), lp.LiquidityProviderUnits.String(), sdk.NewUint(clp.MaxWbasis).String(), sdk.NewInt(0))
```
alexanderbez, line 29: nit: a bit difficult to read
```suggestion
				withdrawNativeAssetAmount, withdrawExternalAssetAmount, _, _ := clp.CalculateWithdrawal(
				  pool.PoolUnits, 
				  pool.NativeAssetBalance.String(),
				  pool.ExternalAssetBalance.String(), 
				  lp.LiquidityProviderUnits.String(), 
				  sdk.NewUint(clp.MaxWbasis).String(), 
				  sdk.NewInt(0),
				)
```

app/upgrade-poolChange.go
```diff
@@ -0,0 +1,96 @@
...
+				newLpUnits, lpUnits, err := clp.CalculatePoolUnits(pool.ExternalAsset.Symbol, temp, tempNative, tempExternal,
+					withdrawNativeAssetAmount, withdrawExternalAssetAmount)
```
alexanderbez, line 31: ditto
```suggestion
				newLpUnits, lpUnits, err := clp.CalculatePoolUnits(
				  pool.ExternalAsset.Symbol,
				   temp,
				    tempNative,
				     tempExternal,
				  withdrawNativeAssetAmount,
				  withdrawExternalAssetAmount,
				  )
```

app/upgrade-poolChange.go
```diff
@@ -0,0 +1,96 @@
...
+	}
+
+	err = ioutil.WriteFile(fmt.Sprintf("%v-state.json", name), appStateJSON, 0600)
+	if err != nil {
+		ctx.Logger().Error(fmt.Sprintf("failed to write state to file: %s", err.Error()))
+	}
+	err = ioutil.WriteFile(fmt.Sprintf("%v-validator.json", name), valList, 0600)
+	if err != nil {
+		ctx.Logger().Error(fmt.Sprintf("failed to write Validator List to file: %s", err.Error()))
+	}
```
alexanderbez, line 95: Why do you not want to preemptively return an any non-nil error case?

x/clp/alias.go
```diff
@@ -15,6 +15,8 @@ const (
 	MaxWbasis         = types.MaxWbasis
 	PoolThrehold      = types.PoolThrehold
 	PoolUnitsMinValue = types.PoolUnitsMinValue
+	SwapType          = types.SwapType
+	TxFeeMultiplier   = types.TxFeeMultiplier
 )
```
alexanderbez, line 21: Just a friendly reminder, that the SDK has moved away from the `alias.go` file convention -- it really didn't add any benefits. That being said, feel free to keep it if it makes sense for your team.

x/faucet/ante.go
```diff
@@ -0,0 +1,23 @@
+package faucet
+
+import (
+	"github.com/Sifchain/sifnode/x/clp"
+	types2 "github.com/Sifchain/sifnode/x/faucet/types"
```
alexanderbez, line 5:
```suggestion
	ftypes "github.com/Sifchain/sifnode/x/faucet/types"
```

x/faucet/ante.go
```diff
@@ -0,0 +1,23 @@
...
+
+type RemoveFaucetFeeDecorator struct{}
+
+func NewRemoveFacuetFeeDecorator() RemoveFaucetFeeDecorator {
+	return RemoveFaucetFeeDecorator{}
+}
+
+var _ types.AnteDecorator = RemoveFaucetFeeDecorator{}
+
+func (r RemoveFaucetFeeDecorator) AnteHandle(ctx types.Context, tx types.Tx, simulate bool, next types.AnteHandler) (newCtx types.Context, err error) {
```
alexanderbez, line 17: I would add a small godoc here illustrating what is happening. e.g. `// On a testnet only where the chain-id is not equal to "sifchain" ...`

x/faucet/handler.go
```diff
@@ -67,8 +67,8 @@ func handleMsgRequestCoins(ctx sdk.Context, keeper Keeper, msg types.MsgRequestC
 // Handle the add coins message and send coins from the signers account to the module account.
 func handleMsgAddCoins(ctx sdk.Context, keeper Keeper, msg types.MsgAddCoins) (*sdk.Result, error) {
 	if ctx.ChainID() != "sifchain" {
-		bank := keeper.GetBankKeeper()
-		err := bank.SendCoins(ctx, msg.Signer, types.GetFaucetModuleAddress(), msg.Coins)
+		supply := keeper.GetSupplyKeeper()
```
alexanderbez, line 70: Why the change in name? The bank keeper is still used, no?

x/clp/ante.go
```diff
@@ -0,0 +1,89 @@
...
+	msg := feeTx.GetMsgs()[0]
+	switch msg := msg.(type) {
+	case clpTypes.MsgSwap:
+		payer := feeTx.FeePayer()
+		// Fee Payer is always the msg signer for swap transactions
+		if !payer.Equals(msg.Signer) {
+			return types.Context{}, errors.New("Fee Payer and MSG Signer are not the same ")
+		}
+		// User is sending rowan , that means they already have rowan balance
+		if msg.SentAsset.Equals(GetSettlementAsset()) {
```
alexanderbez, line 39: What's the difference between `clpTypes.GetSettlementAsset()` and `GetSettlementAsset()`?

x/clp/ante.go
```diff
@@ -0,0 +1,89 @@
...
+			return next(ctx, tx, simulate)
+		}
+		// The amount of Fee user agreed to pay
+		feeInRowan := feeTx.GetFee()
+		requiredRowan := feeInRowan.AmountOf(clpTypes.GetSettlementAsset().Symbol)
```
alexanderbez, line 44: You call `clpTypes.GetSettlementAsset().Symbol` twice. May I suggest storing it in a variable?

app/upgrade-poolChange.go
Jedi2002: Yeah , this can be added in , but this is an older upgrade which already went through .

app/upgrade-poolChange.go
Jedi2002: I think the idea was, to write to a whatever we can , becuase I see ExportAppState does not return an error . 
So its like write to a file if you can , if you cant just continue . 
@iverc , is that correct ?

x/faucet/handler.go
Jedi2002: The refactor is just to mainitain consistence ,the request and add coins now use similar functions .

iverc (approved):

intl-man:
@Jedi2002 The UI has since been decoupled. If this is still relevant, then a new PR should be raised. Let me know, thanks!

PR #951, opened 2021-03-15
[ui] app/RowanPrice: pollData(), set interval

59023g:
This wraps existing `fetch` for token stats in a `setInterval` so the value of Rowan is updated every 10 seconds


ui/app/src/components/shared/RowanPrice.vue
```diff
@@ -1,26 +1,30 @@
...
+      const json = await data.json();
+      const rowanPriceInUSDT = json.body ? json.body.rowanUSD : "";
+
+      console.log('rowanPriceInUSDT')
+      if (isNumeric(rowanPriceInUSDT)) {
+        this.rowanUSD = "ROWAN: $" + parseFloat(rowanPriceInUSDT).toPrecision(6);
+      }   
     }
+  },
+  created() {
```
ryardley, line 25: Should this clear the interval when the component is unmounted or does it not matter in this instance?

59023g: It wouldn't matter in this instance as the component should never be destroyed. But it's probably good practice either way

thomasdavis (approved):
From my comments on Slack, probably good to add a `clearInterval` on component unmount. 

Also a wild console log in there.

thomasdavis (approved):

thomasdavis (approved):

PR #971, opened 2021-03-18
[ChainOps] Updated explorer config.

intl-man:
Includes:

* Latest tokens as per [here](https://app.zenhub.com/workspaces/sifchain-603ed608675da3001744d736/issues/sifchain/issues/15).

### DIFF
--- a/deploy/helm/block-explorer/templates/configmap.yaml
+++ b/deploy/helm/block-explorer/templates/configmap.yaml
@@ -7,75 +7,360 @@ metadata:
 data:
   settings: |
     {
-        "public":{
-            "chainName": "Sifchain",
-            "chainId": "{{ .Values.blockExplorer.env.chainnet }}",
-            "gtm": "",
-            "slashingWindow": 10000,
-            "uptimeWindow": 250,
-            "initialPageSize": 30,
-            "secp256k1": false,
-            "bech32PrefixAccAddr": "sif",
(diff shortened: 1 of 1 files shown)


deploy/helm/block-explorer/templates/configmap.yaml
```diff
@@ -7,75 +7,360 @@ metadata:
...
+          "bech32PrefixConsPub":"sifvalconspub",
+          "bondDenom":"rowan",
+          "powerReduction":1000000,
+          "coins":[
+             {
+                "symbol":"rowan",
+                "decimals":18,
+                "name":"Rowan",
+                "imageUrl":"./images/siflogo.png",
+                "network":"sifchain"
```
59023g, line 32: Unless I'm missing something, this isn't right. See: https://github.com/Sifchain/sifnode/pull/969/files#diff-7033ea6634ac68835b560460888894b0f7ea755c6952be25104b98c44050c082

59023g (changes requested):

deploy/helm/block-explorer/templates/configmap.yaml
intl-man: You're right. I just pulled in the frontend config, but this won't parse at all. This will completely break things. I'll close this PR and will raise a new one.

PR #972, opened 2021-03-18
[ChainOps] Volume snapshot role/policy tweaks

intl-man:
Includes:

* Use the cluster name in the role and policy name.


alexhui01 (approved):
LGTM!

PR #978, opened 2021-03-18
Inflationary Block Rewards

AustinoBombino:
Mint module import to turn on stock inflationary block rewards


59023g:
Will this unlock `https://api.sifchain.finance/minting/annual-provisions`?

AustinoBombino:
> Will this unlock `https://api.sifchain.finance/minting/annual-provisions`?

@59023g Yes it will

AustinoBombino:
Can anyone with more knowledge about simapp comment on whether we should be adding mint / distribution / etc to its app.go? @Jedi2002 
I didn't add it because distribution and others are also not present.

iverc (commented):
code looks good, doing some testing @Jedi2002 appreciate it if you could give it a shot as well

Jedi2002:
> Can anyone with more knowledge about simapp comment on whether we should be adding mint / distribution / etc to its app.go? @Jedi2002
> I didn't add it because distribution and others are also not present.

The simapp is used by individual modules for unit tests , it really does not need to have all the keepers. In the future if while testing a module ,we do need mint ,distribution etc . The developer can make the changes if required to the simapp structure .

With that said however , if you want to do it now you could , it wont affect present test cases

app/app.go
```diff
@@ -336,7 +352,8 @@ func NewInitApp(
 	// During begin block slashing happens after distr.BeginBlocker so that
 	// there is nothing left over in the validator fee pool, so as to keep the
 	// CanWithdrawInvariant invariant.
-	app.mm.SetOrderBeginBlockers(distr.ModuleName,
+	app.mm.SetOrderBeginBlockers(mint.ModuleName,
```
Jedi2002, line 355: Do the minted tokens get distributed in the same block ?

tools/sifgen/common/types/genesis.go
```diff
@@ -175,6 +175,20 @@ type StakingParams struct {
 	BondDenom         string `json:"bond_denom"`
 }
 
+type Mint struct {
+	Minter interface{} `json:"minter" yaml:"minter"`
+	Params MintParams  `json:"params" yaml:"params"`
+}
+
+type MintParams struct {
```
Jedi2002, line 183: Are the default values okay ? 
MintDenom:           sdk.DefaultBondDenom,
InflationRateChange: sdk.NewDecWithPrec(13, 2),
InflationMax:        sdk.NewDecWithPrec(20, 2),
InflationMin:        sdk.NewDecWithPrec(7, 2),
GoalBonded:          sdk.NewDecWithPrec(67, 2),
BlocksPerYear:       uint64(60 * 60 * 8766 / 5), // assuming 5 second block times

If not , we might need to add custom values to tools -> sifgen - > genesis.json

tools/sifgen/common/types/genesis.go
```diff
@@ -175,6 +175,20 @@ type StakingParams struct {
 	BondDenom         string `json:"bond_denom"`
 }
 
+type Mint struct {
+	Minter interface{} `json:"minter" yaml:"minter"`
```
Jedi2002, line 179: Same for the minter ,in case the default value 
sdk.NewDecWithPrec(13, 2) is not okay we need to add it to sifgen tool

app/app.go
AustinoBombino: Yes, they get sent to the fee collector every block

tools/sifgen/common/types/genesis.go
AustinoBombino: Yeah, the goal right now is to just bring in stock cosmos. We'll be modifying down the road when we implement rebalancing.

tools/sifgen/common/types/genesis.go
AustinoBombino: Yup, all default for now

caseyturnerarrington:
@utx0 @intl-man FYI

intl-man:
@AustinoBombino What's the status of this? Is this still going ahead? This PR is stale.

AustinoBombino:
This feature is pushed back to post 0.42. Closing for now

PR #983, opened 2021-03-19
[ui] don't allow for asymetrical when creating a pool for the first time

thomasdavis:
When adding liquidity, if it is a new pool, then we hide the toggle guided liquidity button which makes sure users always pool symmetrically when creating a new pool.


thomasdavis:
This PR implements the fix incorrectly. 

When creating a pool for the first time 
- We should hide the toggle guided liquidity button 
- The behavior of the two inputs should be that both must have a value larger than 0.

thomasdavis:
@59023g @caseyturnerarrington 

This PR is ready to review. When creating a new pool, it disables/hides the guided button and relies on the user to add a value to both A and B amount, and otherwises shows a new error message on the submit button.

59023g:
Looks good. A couple items: 
1. The copy should be shorter: From "Both inputs needed to create a pool" to something like"Symmetric Pool Required" (<- @caseyturnerarrington )
2. The guided liquidity toggle is hidden, but doesn't appear to be active. When I add a value, the corresponding value is not shown (Could it be it's not possible to do this since the pool has no values, and therefore ratio?)
https://user-images.githubusercontent.com/5235538/112207993-d1e2b780-8be5-11eb-9dd1-ac05ea0c493e.mov

caseyturnerarrington:
@59023g a few responses to your comments above:
- The copy should be shorter: From "Both inputs needed to create a pool" to something like"Symmetric Pool Required" (<- @caseyturnerarrington ): Let's change this to 'Both inputs required'.
- The guided liquidity toggle is hidden, but doesn't appear to be active. When I add a value, the corresponding value is not shown (Could it be it's not possible to do this since the pool has no values, and therefore ratio?) -> This is how it should be. There is no way to do a 'pool equally calculation' due to the pool being initialized. We hide it and turn it off here since it isn't possible to 'pool equally'.

thomasdavis:
@caseyturnerarrington @59023g 

1) I updated the copy on button
2) Left the behavior as is 
3) Passed the tests

caseyturnerarrington:
Looks good. Approving

caseyturnerarrington (approved):

59023g (approved):

PR #985, opened 2021-03-22
[ChainOps] Autoscaler fixes.

intl-man:
Includes:

* Cluster autoscaler service account name to be unique, to prevent clashes with clusters on the same AWS account.


gzukel (approved):
Looks good changing naming structure.

PR #995, opened 2021-03-23
[ui] sort balanaces alphabetically and then numerically

thomasdavis:
image

Balances should move to the top desc-> asc, then everything should be alphabetically asc-> desc


thomasdavis:
image
image

@caseyturnerarrington @59023g  

Hey guys, this is ready to test. It mostly works, but there is one slight edge case. Because we only display `2` decimals, sometimes it is in mechanically in order, but because we render only render 2 decimals for display the alphabetic ordering looks slightly off.

caseyturnerarrington (approved):

59023g (approved):

PR #1007, opened 2021-03-26
[ChainOps] TF Config.

intl-man:
Includes:

* Option to define the capacity per node-group.


gzukel (approved):
Looks good I don't see any issues with it.

PR #1008, opened 2021-03-26
Integration tests - better handling of rowan costs

banshee:


khdegraaf (approved):

test/integration/src/py/conftest.py
```diff
@@ -178,8 +178,14 @@ def is_ganache(ethereum_network):
...
-def sifchain_fees():
-    return "200000rowan"
+def sifchain_fees(sifchain_fees_int):
+    """returns a string suitable for passing to sifnodecli"""
+    return f"{sifchain_fees_int}rowan"
+
+
+@pytest.fixture
+def sifchain_fees_int():
+    return "200000"
```
khdegraaf, line 188: causing failure currently

khdegraaf: probably should return int not string?

PR #1009, opened 2021-03-26
[testnet-hotfix][ui] ui/core: testnet, devnet chainId (semi) fix

59023g:
UI fix to rename `testnet` and `devnet` chains to `sifchain-testnet` and `sifchain-devnet` respectively. This will allow for Keplr to properly differentiate between environments so transactions are broadcast to the right endpoints.


intl-man (approved):

gzukel (approved):
Looks good. Don't see any issues.

PR #1012, opened 2021-03-26
app: merge binaries & clean app.go

marbar3778:
## Description

Merge binaries and cleanup app.go

This will not build

- [x] comment out custom modules for now
- [x] remove sifnodecli
- [x] merge sifnode* binaries
- [x] cleanup `app.go` for 0.42
- [x] fix exporting commands


tools/sifgen/node/node.go
```diff
@@ -3,12 +3,13 @@ package node
...
 	"encoding/json"
 	"fmt"
-	"github.com/Sifchain/sifnode/app"
 	"io/ioutil"
 	"net/http"
 	"os"
 	"strings"
 
+	"github.com/Sifchain/sifnode/app"
+
```
marbar3778, line 12:
```suggestion
```

app/app.go
```diff
@@ -56,333 +77,314 @@ var (
...
-		clp.StoreKey,
-		gov.StoreKey,
-		faucet.StoreKey,
-		distr.StoreKey,
-		slashing.StoreKey,
+		authtypes.StoreKey,
+		stakingtypes.StoreKey,
+		paramstypes.StoreKey,
+		upgradetypes.StoreKey,
+		// oracle.StoreKey,
```
iverc, line 160: can we remove unused keys instead of commenting them out?

app/app.go
```diff
@@ -56,333 +77,314 @@ var (
...
+	// 	app.BankKeeper,
+	// 	app.OracleKeeper,
+	// 	keys[ethbridge.StoreKey],
+	// )
+
+	// app.clpKeeper = clp.NewKeeper(
+	// 	appCodec,
+	// 	keys[clp.StoreKey],
+	// 	app.BankKeeper,
+	// 	app.subspaces[clp.ModuleName])
```
iverc, line 233: Same here, can we simply remove commented code?

app/app.go
```diff
@@ -56,333 +77,314 @@ var (
...
+	// 			ctx.Logger().Error(fmt.Sprintf("Invalid | Pool %s ", pool.String()))
+	// 			break
+	// 		}
+	// 		poolList = append(poolList, pool)
+	// 	}
+	// 	// If we have error dont set state
+	// 	if hasError {
+	// 		ctx.Logger().Error("Failed to execute upgrade plan for pool re-balance")
+	// 	}
+	// 	// If we have no errors , Set state .
```
iverc, line 295: Same here

app/app.go
```diff
@@ -451,28 +468,58 @@ func GetMaccPerms() map[string][]string {
...
+func (app *SifchainApp) RegisterTendermintService(clientCtx client.Context) {
+	tmservice.RegisterTendermintService(app.BaseApp.GRPCQueryRouter(), clientCtx, app.interfaceRegistry)
+}
+
+// RegisterSwaggerAPI registers swagger route with API Server
+func RegisterSwaggerAPI(ctx client.Context, rtr *mux.Router) {
+	statikFS, err := fs.New()
 	if err != nil {
-		ctx.Logger().Error(fmt.Sprintf("failed to write Validator List to file: %s", err.Error()))
+		panic(err)
```
iverc, line 507: Do we really want to panic here instead of just logging an error?

app/app_test.go
```diff
@@ -1,40 +1,35 @@
...
+// 	_, _, err = app.ExportAppStateAndValidators(false, []string{})
+// 	require.NoError(t, err, "ExportAppStateAndValidators should not have an error")
+// }
+
+// // ensure that black listed addresses are properly set in bank keeper
+// func TestBlackListedAddrs(t *testing.T) {
+// 	db := db.NewMemDB()
+// 	app := NewSifApp(log.NewTMLogger(log.NewSyncWriter(os.Stdout)), db, nil, true, 0)
+
+// 	for acc := range maccPerms {
```
iverc, line 23: Same here, let's just remove unused code instead of commenting it out

app/export.go
```diff
@@ -2,38 +2,49 @@ package app
 
 import (
 	"encoding/json"
+	"fmt"
```
iverc, line 5: Can we use new logger right away for new files / logs ? No need to replace all the logs since it's been done in a separate PR, but for new logs it could be cleaner to use new approach initially

timlind:
What's the plan to get this building @marbar3778?

timlind (approved):

app/app.go
marbar3778: yea i can remove. I commented because it will be broght back later in subsequent PRs.

app/app.go
marbar3778: I would say panics are fine here because this code gets called at the start of a node/chain. It would not panic in the middle of running the process.

app/export.go
marbar3778: im removed some fmt.Sprintf but for others its fine unless you are asking to concat strings??

marbar3778:
> What's the plan to get this building @marbar3778?

once other modules are migrated over this will compile

PR #1016, opened 2021-03-28
proto: orcale module split out

marbar3778:
## Description

Split out oracle module from larger PR

- [x] migrate storage to protobuf
- [x] update module.go
- [x] update expected_keeper interfaces


app/app.go
```diff
@@ -78,9 +82,8 @@ var (
 		params.AppModuleBasic{},
 		// clp.AppModuleBasic{},
 		upgrade.AppModuleBasic{},
-		// oracle.AppModuleBasic{},
-		// ethbridge.AppModuleBasic{},
-		// faucet.AppModuleBasic{},
+		oracle.AppModuleBasic{},
```
iverc, line 85: Please remove commented code

app/test_helpers.go
```diff
@@ -0,0 +1,107 @@
...
+
+	return app
+}
+
+// CreateTestPubKeys returns a total of numPubKeys public keys in ascending order.
+func CreateTestPubKeys(numPubKeys int) []cryptotypes.PubKey {
+	var publicKeys []cryptotypes.PubKey
+	var buffer bytes.Buffer
+
+	// start at 10 to avoid changing 1 to 01, 2 to 02, etc
```
iverc, line 77: Should we start at 10 or fix the comment?

app/test_helpers.go
```diff
@@ -0,0 +1,107 @@
...
+// DefaultConsensusParams defines the default Tendermint consensus params used in
+// SimApp testing.
+var DefaultConsensusParams = &abci.ConsensusParams{
+	Block: &abci.BlockParams{
+		MaxBytes: 200000,
+		MaxGas:   2000000,
+	},
+	Evidence: &tmproto.EvidenceParams{
+		MaxAgeNumBlocks: 302400,
+		MaxAgeDuration:  504 * time.Hour, // 3 weeks is the max duration
```
iverc, line 29: We could refactor these constants, so it is easier to keep track / reuse them (optional)

iverc (approved):
Have some comments / suggestions, other than that the code looks good. Please remove commented code and address todo(s). The only thing that could actually affect the logic is https://github.com/Sifchain/sifnode/pull/1016/files#diff-bf7bfe9abfe4f3a4bad724b9e155c1a360666a6ed0f6b535ac29ebe8e6de5849R77 (depending on if we should fix the loop or just update the comment)

app/test_helpers.go
marbar3778: Easier to keep track of?

app/app.go
timlind: Don't these need to be re-added once they are upgraded?

x/oracle/module.go
```diff
@@ -2,95 +2,115 @@ package oracle
...
+	_ module.AppModuleBasic      = AppModuleBasic{}
+	_ module.AppModuleSimulation = AppModule{}
 )
 
 // AppModuleBasic defines the basic application module used by the oracle module.
-type AppModuleBasic struct{}
-
-var _ module.AppModuleBasic = AppModuleBasic{}
+type AppModuleBasic struct {
+	cdc codec.Marshaler
```
timlind, line 32: Is this new codec being setup in app and passed through correctly yet? Will help fix code breaks if it is.

marbar3778: it's passed in below, just added it in. It is used for simulations, but don't think this repo uses those

PR #1017, opened 2021-03-28
modules: use provided logger

marbar3778:
## Description

Use provided logger instead of creating new ones


iverc (approved):

x/ethbridge/handler.go
```diff
@@ -56,28 +39,30 @@ func NewHandler(
...
-			errorMessageKey, err.Error())
+		logger.Error("bridge keeper failed to process claim.",
+			"error:", err.Error())
 		return nil, err
 	}
 	if status.Text == oracle.SuccessStatusText {
-		if err = bridgeKeeper.ProcessSuccessfulClaim(ctx, status.FinalClaim, sugaredLogger); err != nil {
-			sugaredLogger.Errorw("bridge keeper failed to process successful claim.",
-				errorMessageKey, err.Error())
+		if err = bridgeKeeper.ProcessSuccessfulClaim(ctx, status.FinalClaim); err != nil {
```
ElliotFriedman, line 57: Does this logger format things in a JSON format? That was one of the motivations behind using the sugared logger.

ElliotFriedman (changes requested):

x/ethbridge/handler.go
marbar3778: you can define any logger you want. It is defined here https://github.com/Sifchain/sifnode/blob/develop/app/app.go#L129. You can have plain text or json by default or define your own logger by implementing the interface.

marbar3778: Here are some docs I recommend reading https://docs.tendermint.com/master/nodes/logging.html and here you can see the change you can make to define json.

ElliotFriedman: I read that doc and it did not mention JSON anywhere in it.

To use JSON, do I need to build a JSON object, stringify it, then log it out?

marbar3778: The one that would be used on the current chain is https://github.com/go-kit/kit/blob/master/log/json_logger.go. You can change a  node that is currently running to json format. You may need to define string methods if they are not defined, but if zap was working so will this.  The new version of the sdk uses zerolog.

x/ethbridge/handler.go
```diff
@@ -56,28 +39,30 @@ func NewHandler(
...
+			logger.Error("bridge keeper failed to process successful claim.",
+				"error:", err.Error())
```
banshee, line 59: Why did you switch this to using a string instead of errorMessageKey?

marbar3778: I can revert. I found it confusing this was a key, but makes sense for your logging system.

marbar3778: reverted

x/clp/handler.go
```diff
@@ -2,12 +2,13 @@ package clp
 
 import (
 	"fmt"
```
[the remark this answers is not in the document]
marbar3778: its still used here 

```
			errMsg := fmt.Sprintf("unrecognized %s message type: %T", ModuleName, msg)
			return nil, sdkerrors.Wrap(sdkerrors.ErrUnknownRequest, errMsg)
```

iverc (approved):

iverc (approved):

PR #1027, opened 2021-03-29
Feature/multi sig

Jedi2002:
The PR contains the basic implementation for creating an airdrop and related queries . 

An additional feature for this module has been implemented in 
https://github.com/Sifchain/sifnode/pull/1082


intl-man (approved):
@Jedi2002 this LGTM. I've gone through it a few times and can't see anything that raises red flags. Would like to get this merged and deployed into DevNet sooner rather than later.

x/dispensation/types/records.go
```diff
@@ -0,0 +1,100 @@
...
+		return "Pending"
+	case Completed:
+		return "Completed"
+	default:
+		return "All"
+	}
+}
+
+// DistributionRecord is created for every recipient for a distribution
+// TODO : Remove ClaimStatus from this struct and use it as prefix.
```
Jedi2002, line 28: Or keep it in both places

intl-man (approved):

app/poolChange.go
```diff
@@ -0,0 +1,96 @@
...
+				break
+			}
+			poolList = append(poolList, pool)
+		}
+		// If we have error dont set state
+		if hasError {
+			ctx.Logger().Error("Failed to execute upgrade plan for pool re-balance")
+		}
+		// If we have no errors , Set state .
+		if !hasError {
```
timlind, line 61: Using sub functions that returns errors might make this a bit cleaner / easier to follow.

timlind (approved):

iverc (approved):
🚀

app/poolChange.go
Jedi2002: Yeah that would have been the better thing to do here . 
But this is an old piece of code ,and the upgrade for this has already been executed . Just keeping it here for historical records .

PR #1029, opened 2021-03-29
[ui] ui/chains: sifchain -> sifchain-local

59023g:
This allows us to connect local config in local development environment. It also gets us to parity with the other config pattern.


ryardley (commented):
So I did a quick grep of the code does this matter at all with the new scheme (it might not)? 

https://github.com/Sifchain/sifnode/blob/develop/ui/app/src/components/shared/utils.ts#L78

ryardley:
Also need to change  `chains/peggy/run.sh` I think

ryardley (approved):
🚀 🌙

PR #1033, opened 2021-03-30
UI fixes

iverc:
Some UI fixes to make it look even better.


ryardley (changes requested):
`trimZeros` is deprecated and should not be used. We should not merge a PR that affects display amounts until we get the AmountAPI in.  We can get away with a PR that changes the arrow as it does not affect the amounts being displayed. I suggest removing the number formatting stuff from this PR but also removing `canSwap` as it is no longer used in the `CurrencyPairPanel`

intl-man:
@iverc Closing as the UI has since been decoupled.

PR #1043, opened 2021-03-31
[ChainOps] Updated testnet contracts

intl-man:
Includes:

* Latest contracts for DevNet.
* Latest contracts for TestNet.


gzukel (approved):
Loos Good.

PR #1045, opened 2021-03-31
[ChainOps] Workflow tweaks

intl-man:
Includes:

* Look for `devnet` in the release name and set `CHAINNET=testnet` accordingly.


gzukel (approved):
Approved

PR #1063, opened 2021-04-01
faucet: remove module

marbar3778:


iverc (dismissed):

banshee (changes requested):
We need to change the tests as well.  I added https://github.com/Sifchain/issues/issues/264

PR #1076, opened 2021-04-02
Truffle update requires no sudo for install

banshee:
Possible fix for our truffle issue.  We used to need sudo to install, and it's possible that we can't use sudo any more.


khdegraaf (approved):
Lovely

PR #1079, opened 2021-04-03
Ethbridge querier and query / msg service stubs

timlind:
@marbar3778 Let me know if things need to be done differently or anything was missed. Would like to get this merged in just focussing on the querier so we're working off a current branch.


proto/sifnode/ethbridge/v1/query.proto
```diff
@@ -0,0 +1,35 @@
...
+
+option go_package = "github.com/Sifchain/sifnode/x/ethbridge/types";
+
+// Query service for queries
+service Query {
+  // EthProphecy queries an EthProphecy
+  rpc EthProphecy(QueryEthProphecyRequest) returns (QueryEthProphecyResponse) {}
+}
+
+// TODO check rest property names
```
iverc, line 16: Let's do some cleanup if we merge it in as is so we don't have to do it when it gets bigger

iverc (approved):

x/ethbridge/client/cli/query.go
```diff
@@ -42,25 +47,21 @@ func GetCmdGetEthBridgeProphecy(queryRoute string, cdc *codec.Codec) *cobra.Comm
...
 
-			var out types.QueryEthProphecyResponse
-			err = cdc.UnmarshalJSON(res, &out)
+			res, err := queryClient.EthProphecy(context.Background(), req)
 			if err != nil {
 				return err
 			}
 
-			return cliCtx.PrintOutput(out)
+			return clientCtx.PrintProto(res)
```
timlind, line 64: @marbar3778 what should this line look like?

proto/sifnode/ethbridge/v1/query.proto
marbar3778: Is this todo carried over from the previous query?

timlind: I've added it as there was one field name distinction in the json tag, will update that in the Proto now.

PR #1080, opened 2021-04-03
upgrade to 0.42

marbar3778:
## Description

Upgrade to protobuf


cmd/ebrelayer/relayer/ethereum.go
```diff
@@ -364,21 +359,21 @@ func (sub EthereumSub) Replay(fromBlock int64, toBlock int64, cosmosFromBlock in
 	bridgeClaims := sub.getAllClaims(cosmosFromBlock, cosmosToBlock)
 	log.Printf("found out %d bridgeClaims\n", len(bridgeClaims))
 
-	client, err := SetupRPCEthClient(sub.EthProvider)
+	c, err := SetupRPCEthClient(sub.EthProvider)
```
ElliotFriedman, line 362: This variable has to be renamed. It is so non descriptive. ethRPCClient would work

cmd/ebrelayer/relayer/ethereum.go
```diff
@@ -461,15 +456,15 @@ func (sub EthereumSub) logToEvent(clientChainID *big.Int, contractAddress common
 
 // handleEthereumEvent unpacks an Ethereum event, converts it to a ProphecyClaim, and relays a tx to Cosmos
 func (sub EthereumSub) handleEthereumEvent(events []types.EthereumEvent) error {
-	var prophecyClaims []ethbridge.EthBridgeClaim
+	var prophecyClaims []*ethbridge.EthBridgeClaim
```
ElliotFriedman, line 459: Why does this array of EthBridgeClaims have to be an array of pointers? cc @iverc @timlind

ElliotFriedman: Why can't it just be an array of EthBridgeClaims?

cmd/ebrelayer/txs/parser.go
```diff
@@ -68,14 +68,14 @@ func EthereumEventToEthBridgeClaim(valAddr sdk.ValAddress, event types.EthereumE
 	nonce := int(event.Nonce.Int64())
 
 	// Package the information in a unique EthBridgeClaim
-	witnessClaim.EthereumChainID = chainID
-	witnessClaim.BridgeContractAddress = bridgeContractAddress
-	witnessClaim.Nonce = nonce
-	witnessClaim.TokenContractAddress = tokenContractAddress
+	witnessClaim.EthereumChainId = int64(chainID)
```
ElliotFriedman, line 71: The old version of this EthBridgeClaim object had the chainid be an int. We should do the same thing to our protos so that we don't have to do this weird casting here.

cmd/ebrelayer/txs/parser.go
```diff
@@ -68,14 +68,14 @@ func EthereumEventToEthBridgeClaim(valAddr sdk.ValAddress, event types.EthereumE
...
+	witnessClaim.BridgeContractAddress = bridgeContractAddress.String()
+	witnessClaim.Nonce = int64(nonce)
```
ElliotFriedman, line 73: Same thing with nonce. It should just be an int.

ElliotFriedman: Did we do this because protobuffs do not support regular ints? cc @timlind 
https://developers.google.com/protocol-buffers/docs/proto3#scalar

cmd/ebrelayer/txs/relayToCosmos.go
```diff
@@ -23,8 +20,8 @@ var (
 
 // RelayToCosmos applies validator's signature to an EthBridgeClaim message containing
 // information about an event on the Ethereum blockchain before relaying to the Bridge
-func RelayToCosmos(cdc *codec.Codec, moniker, password string, claims []types.EthBridgeClaim, cliCtx context.CLIContext,
-	txBldr authtypes.TxBuilder, sugaredLogger *zap.SugaredLogger) error {
+func RelayToCosmos(moniker, password string, claims []*types.EthBridgeClaim, cliCtx client.Context,
```
ElliotFriedman, line 23: Same question here: Why do we need to make it an array of pointers to EthBridgeClaims instead of just an array of EthBridgeClaims?

cmd/ebrelayer/types/types.go
```diff
@@ -40,16 +40,16 @@ func (d Event) String() string {
...
-	ID                    [32]byte
-	From                  common.Address
 	To                    []byte
-	Token                 common.Address
 	Symbol                string
+	EthereumChainID       *big.Int
 	Value                 *big.Int
 	Nonce                 *big.Int
 	ClaimType             ethbridge.ClaimType
+	ID                    [32]byte
```
ElliotFriedman, line 49: We should not be changing the order in which items in our structs are declared if we can help it. It seems like this change doesn't really even change the keys or values or their types, instead it just moves things around. We should revert these changes.

proto/sifnode/ethbridge/v1/tx.proto
```diff
@@ -0,0 +1,80 @@
...
+
+// MsgLock defines a message for locking coins and triggering a related event
+message MsgLock {
+  string cosmos_sender     = 1;
+  string amount            = 2 [(gogoproto.customtype) = "github.com/cosmos/cosmos-sdk/types.Int",
+    (gogoproto.nullable)   = false,
+    (gogoproto.moretags)   = "yaml:\"amount\""];
+  string symbol            = 3 [(gogoproto.moretags) = "yaml:\"symbol\""];
+  int64  ethereum_chain_id = 4 [(gogoproto.moretags) = "yaml:\"ethereum_chain_id\""];
+  string ethereum_receiver = 5 [(gogoproto.moretags) = "yaml:\"ethereum_receiver\""];
```
ElliotFriedman, line 27: Can we use a custom type here like we do with amount, but instead of sdk.int, use ethereum.address? cc @timlind

proto/sifnode/oracle/v1/types.proto
```diff
@@ -0,0 +1,47 @@
...
+//TODO: Add parameters to Genesis state ,such as minimum liquidity required to create a pool
+message GenesisState {
+  repeated string address_whitelist = 1;
+  string          admin_address     = 2;
+}
+
+// Claim contrains an arbitrary claim with arbitrary content made by a given validator
+message Claim {
+  string id                = 1;
+  string validator_address = 2;
```
ElliotFriedman, line 18: I don't see why we can't use a custom gogo proto type here as well for this instead of just using a string.

x/ethbridge/client/cli/tx.go
```diff
@@ -1,56 +1,57 @@
...
+
 			tokenContract := types.NewEthereumAddress(tokenContractString)
 
 			if !common.IsHexAddress(args[0]) {
 				return errors.Errorf("invalid [bridge-registry-contract]: %s", args[0])
 			}
 			bridgeContract := types.NewEthereumAddress(args[0])
 
-			nonce, err := strconv.Atoi(args[1])
+			nonce, err := strconv.ParseInt(args[1], 10, 64)
```
ElliotFriedman, line 54: Why are we using ParseInt instead of ATOI?

x/ethbridge/client/rest/rest.go
```diff
@@ -93,33 +96,35 @@ func createClaimHandler(cliCtx context.CLIContext) http.HandlerFunc {
...
 			return
 		}
 
+		ct := types.ClaimType(claimType)
+
 		// create the message
 		ethBridgeClaim := types.NewEthBridgeClaim(
-			req.EthereumChainID, bridgeContractAddress, req.Nonce, req.Symbol,
-			tokenContractAddress, ethereumSender, cosmosReceiver, validator, req.Amount, claimType)
+			int64(req.EthereumChainID), bridgeContractAddress, int64(req.Nonce), req.Symbol,
```
ElliotFriedman, line 109: We really shouldn't be having to type cast right here

x/ethbridge/client/rest/rest.go
```diff
@@ -93,33 +96,35 @@ func createClaimHandler(cliCtx context.CLIContext) http.HandlerFunc {
...
 
-		claimType, err := types.StringToClaimType(req.ClaimType)
-		if err != nil {
+		claimType, exist := types.ClaimType_value[req.ClaimType]
+		if !exist {
 			rest.WriteErrorResponse(w, http.StatusBadRequest, types.ErrInvalidClaimType.Error())
 			return
 		}
 
+		ct := types.ClaimType(claimType)
```
ElliotFriedman, line 105: Let's find a better name for this variable

x/ethbridge/handler.go
```diff
@@ -3,315 +3,41 @@ package ethbridge
...
-		log.Fatalln("failed to init zap logging")
-	}
-	return logger.Sugar()
-}
-
 // NewHandler returns a handler for "ethbridge" type messages.
-func NewHandler(
-	accountKeeper types.AccountKeeper, bridgeKeeper Keeper, cdc *codec.Codec) sdk.Handler {
+func NewHandler(k Keeper) sdk.Handler {
+	msgServer := keeper.NewMsgServerImpl(k)
```
ElliotFriedman, line 15: What keeper is this? I'm assuming it's an ethbridge module keeper.

x/ethbridge/keeper/grpc_query.go
```diff
@@ -0,0 +1,55 @@
...
+
+	prophecy, found := srv.Keeper.oracleKeeper.GetProphecy(sdkCtx, id)
+	if !found {
+		return nil, sdkerrors.Wrap(oracletypes.ErrProphecyNotFound, id)
+	}
+
+	bridgeClaims, err := types.MapOracleClaimsToEthBridgeClaims(
+		req.EthereumChainId,
+		types.NewEthereumAddress(req.BridgeContractAddress),
+		req.Nonce,
```
ElliotFriedman, line 39: Any type casting we need to do here before calling the `MapOracleClaimsToEthBridgeClaims` function?

x/ethbridge/keeper/grpc_query.go
```diff
@@ -0,0 +1,55 @@
...
+		req.Symbol,
+		types.NewEthereumAddress(req.TokenContractAddress),
+		types.NewEthereumAddress(req.EthereumSender),
+		prophecy.ValidatorClaims,
```
ElliotFriedman, line 43: This is incorrect as the cosmos receiver address should be included here when creating the new OracleClaim from EthBridgeClaim

This means we need to modify this function: `CreateEthClaimFromOracleString` to accept a sifchain address.

We need to ensure that the prophecy ID's get constructed exactly the same way they did before to prevent hash collisions.

x/ethbridge/keeper/keeper.go
```diff
@@ -4,36 +4,38 @@ import (
...
 // exposes getter/setter methods for the various parts of the state machine
 type Keeper struct {
-	cdc *codec.Codec // The wire codec for binary encoding/decoding.
+	cdc codec.BinaryMarshaler // The wire codec for binary encoding/decoding.
 
-	supplyKeeper types.SupplyKeeper
-	oracleKeeper types.OracleKeeper
-	storeKey     sdk.StoreKey
+	accountKeeper types.AccountKeeper
+	bankKeeper    types.BankKeeper
```
ElliotFriedman, line 26: Bank keeper is now standing in for supply keeper I take it?

ElliotFriedman (changes requested):

x/ethbridge/keeper/msg_server.go
```diff
@@ -0,0 +1,322 @@
...
+	}
+
+	cosmosSender, err := sdk.AccAddressFromBech32(msg.CosmosSender)
+	if err != nil {
+		return nil, err
+	}
+
+	account := srv.Keeper.accountKeeper.GetAccount(ctx, cosmosSender)
+	if account == nil {
+		logger.Error("account is nil.", "CosmosSender", msg.CosmosSender)
```
ElliotFriedman, line 44: should be `sender account is nil` instead of just `account is nil`.

x/ethbridge/keeper/msg_server.go
```diff
@@ -0,0 +1,322 @@
...
+	ctx.EventManager().EmitEvents(sdk.Events{
+		sdk.NewEvent(
+			sdk.EventTypeMessage,
+			sdk.NewAttribute(sdk.AttributeKeyModule, types.AttributeValueCategory),
+			sdk.NewAttribute(sdk.AttributeKeySender, msg.CosmosSender),
+		),
+		sdk.NewEvent(
+			types.EventTypeLock,
+			sdk.NewAttribute(types.AttributeKeyEthereumChainID, strconv.FormatInt(msg.EthereumChainId, 10)),
+			sdk.NewAttribute(types.AttributeKeyCosmosSender, msg.CosmosSender),
```
ElliotFriedman, line 71: We might need to call the .String() method on the msg.CosmosSender if it has been converted to another type by this point.

x/ethbridge/keeper/msg_server.go
```diff
@@ -0,0 +1,322 @@
...
+			sdk.NewAttribute(types.AttributeKeyCosmosSenderSequence, strconv.FormatUint(account.GetSequence(), 10)),
+			sdk.NewAttribute(types.AttributeKeyEthereumReceiver, msg.EthereumReceiver),
```
ElliotFriedman, line 73: Same comment here if EthereumReceiver has been converted to an address type, we will need to convert it.

x/ethbridge/keeper/msg_server.go
```diff
@@ -0,0 +1,322 @@
...
+			sdk.NewAttribute(types.AttributeKeyAmount, msg.Amount.String()),
+			sdk.NewAttribute(types.AttributeKeySymbol, msg.Symbol),
+			sdk.NewAttribute(types.AttributeKeyCethAmount, msg.CethAmount.String()),
```
ElliotFriedman, line 76: Looks like we removed the last item here which was the coins object stringified. It had a key for the event of `AttributeKeyCoins` which had a value of coins.String(). As long as the relayer didn't use this information this is fine.

x/ethbridge/keeper/msg_server.go
```diff
@@ -0,0 +1,322 @@
...
+	})
+
+	return &types.MsgBurnResponse{}, nil
+
+}
+func (srv msgServer) CreateEthBridgeClaim(goCtx context.Context, msg *types.MsgCreateEthBridgeClaim) (*types.MsgCreateEthBridgeClaimResponse, error) {
+	ctx := sdk.UnwrapSDKContext(goCtx)
+
+	logger := srv.Keeper.Logger(ctx)
+	var mutex = &sync.RWMutex{}
```
ElliotFriedman, line 143: We should remove all of these mutex locks.

x/ethbridge/keeper/msg_server.go
```diff
@@ -0,0 +1,322 @@
...
+		sdk.NewEvent(
+			sdk.EventTypeMessage,
+			sdk.NewAttribute(sdk.AttributeKeyModule, types.AttributeValueCategory),
+			sdk.NewAttribute(sdk.AttributeKeySender, msg.EthBridgeClaim.ValidatorAddress),
+		),
+		sdk.NewEvent(
+			types.EventTypeCreateClaim,
+			sdk.NewAttribute(types.AttributeKeyCosmosSender, msg.EthBridgeClaim.ValidatorAddress),
+			sdk.NewAttribute(types.AttributeKeyEthereumSender, msg.EthBridgeClaim.EthereumSender),
+			sdk.NewAttribute(types.AttributeKeyEthereumSenderNonce, strconv.FormatInt(msg.EthBridgeClaim.Nonce, 10)),
```
ElliotFriedman, line 181: Let's revert this back to Itoa instead of FormatInt.

x/ethbridge/keeper/msg_server.go
```diff
@@ -0,0 +1,322 @@
...
+		"CethReceiverAccount", msg.CethReceiverAccount)
+
+	ctx.EventManager().EmitEvents(sdk.Events{
+		sdk.NewEvent(
+			sdk.EventTypeMessage,
+			sdk.NewAttribute(sdk.AttributeKeyModule, types.AttributeValueCategory),
+			sdk.NewAttribute(sdk.AttributeKeySender, msg.CosmosSender),
+		),
+		sdk.NewEvent(
+			types.EventTypeLock,
```
ElliotFriedman, line 287: This is an incorrect event name for what this function does.

x/ethbridge/keeper/peggyTokens.go
```diff
@@ -15,14 +15,14 @@ func (k Keeper) AddPeggyToken(ctx sdk.Context, token string) {
 	store := ctx.KVStore(k.storeKey)
 	key := types.PeggyTokenKeyPrefix
 
-	tokens = append(tokens, token)
-	store.Set(key, k.cdc.MustMarshalBinaryBare(tokens))
+	tokens.Tokens = append(tokens.Tokens, token)
```
ElliotFriedman, line 18: Why do we need to make this change to the way we store data in the keeper? We should be able to do it the old way and just append to an array of type tokens without having to use a property of the object to append to.

x/ethbridge/keeper/querier_test.go
```diff
@@ -7,27 +7,25 @@ import (
...
-	cdc := keeperLib.MakeTestCodec()
+	ctx, keeper, _, _, _, encCfg, _ := test.CreateTestKeepers(t, 0.7, []int64{3, 3}, "")
 
 	query := abci.RequestQuery{
 		Path: "",
 		Data: []byte{},
 	}
 
-	querier := NewQuerier(oracleKeeper, cdc)
+	querier := ethbridgekeeper.NewLegacyQuerier(keeper, encCfg.Amino)
```
ElliotFriedman, line 28: Why are we using amino here? I thought we moved to protobufs

x/ethbridge/module.go
```diff
@@ -34,34 +41,39 @@ func (AppModuleBasic) Name() string {
...
 // DefaultGenesis returns default genesis state as raw bytes for the ethbridge
 // module.
-func (AppModuleBasic) DefaultGenesis() json.RawMessage {
+func (b AppModuleBasic) DefaultGenesis(marshaler codec.JSONMarshaler) json.RawMessage {
 	return nil
 }
 
 // ValidateGenesis performs genesis state validation for the ethbridge module.
-func (AppModuleBasic) ValidateGenesis(_ json.RawMessage) error {
+func (b AppModuleBasic) ValidateGenesis(marshaler codec.JSONMarshaler, config sdkclient.TxEncodingConfig, message json.RawMessage) error {
```
ElliotFriedman, line 55: Why pass all of these parameters if they aren't used?

x/ethbridge/module.go
```diff
@@ -34,34 +41,39 @@ func (AppModuleBasic) Name() string {
...
 	return nil
 }
 
+func (b AppModuleBasic) RegisterGRPCGatewayRoutes(c sdkclient.Context, serveMux *runtime.ServeMux) {
+	// TODO: Register grpc gateway
```
ElliotFriedman, line 60: Is there something else we should be doing here @timlind @Jedi2002?

x/ethbridge/module.go
```diff
@@ -109,36 +126,36 @@ func (am AppModule) RegisterInvariants(ir sdk.InvariantRegistry) {
...
 // QuerierRoute returns the ethbridge module's querier route name.
 func (AppModule) QuerierRoute() string {
 	return QuerierRoute
 }
 
-// NewQuerierHandler returns the ethbridge module sdk.Querier.
-func (am AppModule) NewQuerierHandler() sdk.Querier {
-	return NewQuerier(am.OracleKeeper, am.Codec)
+// Deprecated: LegacyQuerierHandler use RegisterServices
+func (am AppModule) LegacyQuerierHandler(amino *codec.LegacyAmino) sdk.Querier {
```
ElliotFriedman, line 144: Why are we supporting this legacy behavior? I think we should completely just deprecate it if we don't have a use for it.

x/ethbridge/types/claim.go
```diff
@@ -3,43 +3,35 @@ package types
...
+	bridgeContract EthereumAddress,
+	nonce int64,
+	symbol string,
+	tokenContact EthereumAddress,
+	ethereumSender EthereumAddress,
+	cosmosReceiver sdk.AccAddress,
+	validator sdk.ValAddress,
+	amount sdk.Int,
+	claimType ClaimType,
+) *EthBridgeClaim {
```
ElliotFriedman, line 25: Why did we make this function return a pointer to a newly allocated EthBridgeClaim instead of just a new object?

We should roll this back to not using pointers.

x/ethbridge/types/claim.go
```diff
@@ -72,28 +64,39 @@ func NewOracleClaimContent(
...
-	ethereumChainID int, bridgeContract EthereumAddress, nonce int,
-	ethereumAddress EthereumAddress, validator sdk.ValAddress, oracleClaimString string,
-) (EthBridgeClaim, error) {
+	ethereumChainID int64,
+	bridgeContract EthereumAddress,
+	nonce int64,
+	ethereumAddress EthereumAddress,
+	validator sdk.ValAddress,
+	oracleClaimString string,
+) (*EthBridgeClaim, error) {
```
ElliotFriedman, line 96: Same question here: why do we use a pointer to an EthBridgeClaim instead of just a regular EthBridgeClaim.

x/ethbridge/types/msgs.go
```diff
@@ -17,24 +17,14 @@ const (
...
-	Symbol           string          `json:"symbol" yaml:"symbol"`
-	EthereumChainID  int             `json:"ethereum_chain_id" yaml:"ethereum_chain_id"`
-	EthereumReceiver EthereumAddress `json:"ethereum_receiver" yaml:"ethereum_receiver"`
-	CethAmount       sdk.Int         `json:"ceth_amount" yaml:"ceth_amount"`
-}
-
 // NewMsgLock is a constructor function for MsgLock
 func NewMsgLock(
-	ethereumChainID int, cosmosSender sdk.AccAddress,
+	ethereumChainID int64, cosmosSender sdk.AccAddress,
```
ElliotFriedman, line 22: Should this be a uint64 instead of an int64? I'm having a hard time imagining a chainid or chain descriptor being negative.

x/ethbridge/types/msgs.go
```diff
@@ -49,19 +39,19 @@ func (msg MsgLock) Type() string { return "lock" }
 
 // ValidateBasic runs stateless checks on the message
 func (msg MsgLock) ValidateBasic() error {
-	if strconv.Itoa(msg.EthereumChainID) == "" {
-		return sdkerrors.Wrapf(ErrInvalidEthereumChainID, "%d", msg.EthereumChainID)
+	if strconv.FormatInt(msg.EthereumChainId, 10) == "" {
```
ElliotFriedman, line 42: Let's switch back to Itoa.

cmd/ebrelayer/types/types.go
ElliotFriedman: Non issue. Closing.

x/ethbridge/client/cli/tx.go
ElliotFriedman: Reason for using Parseint instead of Atoi is parseint stores result in larger data container.

x/ethbridge/keeper/msg_server.go
ElliotFriedman: Peggy team should resolve this as this is legacy code that needs to be fixed.

marshsif:
This PR should probably be closed.

timlind:
This is an old draft PR for 42 work from Marko.

PR #1084, opened 2021-04-06
[ChainOps] Explorer updates

intl-man:
Includes:

* Decoupled mongodb from the explorer
* and the ability to provide the image and image tag when deploying the explorer, much like sifnode/sifnodecli/ebrelayer.


gzukel (approved):
I don't see anything that stands out immediately as a redflag looks good.

PR #1093, opened 2021-04-07
Legacy rest fixes

timlind:


x/ethbridge/client/rest/rest.go
```diff
@@ -102,7 +104,7 @@ func createClaimHandler(cliCtx context.CLIContext) http.HandlerFunc {
 
 		// create the message
 		ethBridgeClaim := types.NewEthBridgeClaim(
-			req.EthereumChainID, bridgeContractAddress, req.Nonce, req.Symbol,
+			int64(req.EthereumChainID), bridgeContractAddress, int64(req.Nonce), req.Symbol,
```
marbar3778, line 107: out of scope of this pr but we should size the nonce and id. Leaving it as int could potentially cause issues.

marbar3778 (approved):

x/ethbridge/client/rest/rest.go
timlind: Yeah, will add a todo in before I merge.

PR #1096, opened 2021-04-08
[ui] max button minus fees for rowan

thomasdavis:
When;

- Clicking max button when unpegging, you should have enough rowan still to pay for the gas
- When swapping rowan, you should have enough rowan to still pay for the gas


59023g:
This works well. Could you put the calculations in a util function somewhere so we can DRY this up? May need to change the Fraction (Rowan fee) at some time in the future. Would be good to only have to do it in one place

ui/app/src/views/utils/getMaxAmount.ts
```diff
@@ -0,0 +1,22 @@
...
+  accountBalance: IAssetAmount,
+): IAmount {
+  if (!symbol) {
+    return Amount("0");
+  }
+  if (symbol.value !== "rowan") {
+    return accountBalance;
+  } else {
+    if (accountBalance.greaterThan(Amount("0.5"))) {
+      const fee = 5 * Math.pow(10, accountBalance.decimals - 1); // 0.5 ROWAN
```
59023g, line 16: Will you set this as a config value? Or make clearer note this is a constant ?

59023g (approved):

PR #1106, opened 2021-04-09
[DO NOT MERGE] not mint rowan when unpeg erowan to rowan

juniuszhou:


x/ethbridge/keeper/keeper.go
```diff
@@ -75,7 +75,10 @@ func (k Keeper) ProcessSuccessfulClaim(ctx sdk.Context, claim string, sugaredLog
 		err = k.supplyKeeper.MintCoins(ctx, types.ModuleName, coins)
 	case types.BurnText:
 		coins = sdk.Coins{sdk.NewCoin(oracleClaim.Symbol, oracleClaim.Amount)}
-		err = k.supplyKeeper.MintCoins(ctx, types.ModuleName, coins)
+		// not mint if burn token is rowan
+		if oracleClaim.Symbol != types.NativeSymbol {
+			err = k.supplyKeeper.MintCoins(ctx, types.ModuleName, coins)
+		}
```
ElliotFriedman, line 81: Shouldn't we also put logic here to handle the unlocking for rowan.

ElliotFriedman (changes requested):

x/ethbridge/keeper/keeper.go
juniuszhou: If put the unlocking code in the burn case, then we need copy/paste the err handling and SendCoinsFromModuleToAccount code in different cases. it is not good for code reuse. it is better just handle the token symbol and coin mint according to different case. then use the unique code to handle error and SendCoinsFromModuleToAccount logic

ElliotFriedman: I'm unsure why this code is both minting coins, and calling SendCoinsFromModuleToAccount later on line 92.

juniuszhou: will put the mint out of case branch.

x/ethbridge/keeper/keeper.go
```diff
@@ -70,14 +70,20 @@ func (k Keeper) ProcessSuccessfulClaim(ctx sdk.Context, claim string, sugaredLog
...
+			errorMessageKey, err.Error())
+		return err
+	}
+
+	// not mint if burn token is rowan
+	if oracleClaim.Symbol != types.NativeSymbol {
+		err = k.supplyKeeper.MintCoins(ctx, types.ModuleName, coins)
 	}
 
 	if err != nil {
```
ElliotFriedman, line 89: Line 95 should not attempt to send coins from module to account as this could double unlock coins.

Instead, it needs an if statement around it, checking that the oracleClaim.Symbol is equal to the types.NativeSymbol before attempting to send coins from the module to the account.

ElliotFriedman (changes requested):

x/ethbridge/keeper/keeper.go
juniuszhou: agree, and during working on the 42 upgrade, I found out the interface to send coin from account to account in bank module. SendCoinsFromModuleToAccount will not be used for lock/burn, code will be changed when merge with 42 upgrade.

ElliotFriedman: Was that change made on this branch: https://github.com/Sifchain/sifnode/pull/1365

ElliotFriedman: Would you mind merging that change to this branch? It still seems like we will duplicate non rowan tokens unlocking if we merge this branch.

PR #1111, opened 2021-04-09
Ethbridge test compilation fixes

timlind:


timlind:
thanks for the fixes @marbar3778 lgtm

PR #1124, opened 2021-04-13
Feature/filter addresses

Jedi2002:


x/dispensation/utils/parser_test.go
```diff
@@ -59,20 +61,84 @@ func init() {
...
+		panic("Err Reading file :" + err.Error())
+	}
+	err = json.Unmarshal(o, &addresStrings)
+	if err != nil {
+		panic("Err Unmarshall :" + err.Error())
+	}
+	for _, add := range addresStrings {
+		_, err := sdk.AccAddressFromBech32(add)
+		if err != nil {
+			fmt.Println("Invalid :", add)
```
iverc, line 101: Should we panic here as well?

iverc (approved):
1 comment, otherwise lgtm

x/dispensation/utils/parser_test.go
Jedi2002: this test is just to print out the invalid address . So that they can be removed from the input list .

Jedi2002: Its actaully a utility and not a test , jsut kept it under tests temporarily

Jedi2002:
Closing and adding to .42

Issue #1117, opened 2021-04-12
Stopping and restarting `sifnoded` causes error - regression


ryardley:
When a node operator wants to restore from a backup sifnoded will not be able to be started. This also has implications for testing as due to this bug we cannot snapshot state effectively.

**Current behavior**
When restarting a single node sifnoded instance after it has been killed it appears that it cannot continue from the last block and returns an error `failed to load Store: initial version set to X, but found earlier version 1` when it should simply restart. 

**Expected behavior**
It should simply continue from the last block in the database

**Steps to reproduce**

Run the following test script within the ui package:

Create the following script in `ui/chains/cosmos/test.sh`
```
parallelizr() {
  for cmd in "$@"; do {
    echo "Process \"$cmd\" started";
    $cmd & pid=$!
    PID_LIST+=" $pid";
  } done

  trap "kill $PID_LIST" SIGINT

  echo "Parallel processes have started";

  wait $PID_LIST

  echo "All processes have completed";
}

# setup genesis etc. and run sifnoded
./run.sh&

# wait a bit
sleep 30

# kill the processes
killall sifnoded sifnodecli

# wait a bit
sleep 5

# try running the processes again
parallelizr "sifnoded start --trace" "sifnodecli rest-server  --unsafe-cors --trace"
```

output is

```
...
I[2021-04-12|20:10:24.351] Committed state                              module=state height=4 txs=0 appHash=F4D3E85770277C2F9CCE2FA271A5114FAB94E2D5CBB23E2A5537F2AC9201A382
I[2021-04-12|20:10:29.399] Executed block                               module=state height=5 validTxs=0 invalidTxs=0
I[2021-04-12|20:10:29.404] Committed state                              module=state height=5 txs=0 appHash=C8149AC4BAD0972994B9E3C70E87772B0CD076B522F19F34AEF7DDF7D105DA61
I[2021-04-12|20:10:31.365] RPC HTTP server stopped                      module=rest-server err="accept tcp 127.0.0.1:1317: use of closed network connection"
ERROR: accept tcp 127.0.0.1:1317: use of closed network connection
E[2021-04-12|20:10:31.370] Not stopping BlockPool service -- has not been started yet module=blockchain impl=BlockPool
I[2021-04-12|20:10:31.370] exiting...                                   module=main 
All processes have completed
Process "sifnoded start --trace" started
Process "sifnodecli rest-server  --unsafe-cors --trace" started
Parallel processes have started
[...]

PR #1130, opened 2021-04-14
fixed state loader for dispensation

Jedi2002:


alexanderbez (commented):
So this addresses the fact that a module's store, `dispensation`, wasn't being able to be loaded correctly during an upgrade? Was the module in question a _new_ module part of the upgrade?

Jedi2002:
> So this addresses the fact that a module's store, `dispensation`, wasn't being able to be loaded correctly during an upgrade? Was the module in question a _new_ module part of the upgrade?

The ugrade happened already , at some height (X) .  ( So the store for dispensation is already loaded in )

But now the app cannot be stopped stopped and started again. I was able to replicate it simply by 
```
git checkout master
make init-run
pkill sifnoded  < - stop app
./scripts/run.sh <- start app
```

My theroy is the SetState(app) , would be called everytime we restart the app . It sets the state again at that height (X+h) . Which results in the error we see.

This upgrade is simply to remove that piece of code ,so that it does not execute any more .

ryardley:
This fixed the issue I logged here: https://github.com/Sifchain/sifnode/issues/1117 will let others approve.

PR #1142, opened 2021-04-16
[ui] Feature/prevent unpeg bsc

ryardley:
This PR disables pegging when the user is not connected to a supported EVM network. 

After consultation with @j-kappa this adds a few interim (not fancily designed) usability guards for when connecting to an unsupported EVM network:

image

image

This probably can't happen because of the guards but if you succeed at executing the peg action on an unsupported network you get this:

image

### DIFF
--- a/ui/app/src/components/shared/SifButton.stories.ts
+++ b/ui/app/src/components/shared/SifButton.stories.ts
@@ -15,4 +15,4 @@ const Template = (args: any) => ({
 });
 
 export const Primary = Template.bind({});
-Primary.args = {};
+(Primary as any).args = {};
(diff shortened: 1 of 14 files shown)


ryardley:
So I am going to try and revisit this with reference to which sifchain testnet I am connected to and ensure we are dealing with compatible networks: 

image

59023g (approved):
Awesome! Tested, working (with Rinkeby)

caseyturnerarrington:
@ryardley did some testing on this. Looks great! 1 note:
1) I first connected to Binance smart chain.
2) I successfully got the 'network not supported' message.
--- I agree that this message should be more of an error and then guidance message. Like 'Network not supported. Please use your wallet and connect to a supported chain'. 
3) The ETH symbol displayed in the external token list was not actually ETH (it is this dark blue circle).
4) I then switched my MetaMask to be connected to Ethereum. 
5) It auto-updated the UI.
6) But the ETH symbol is still this dark blue circle?

PR #1147, opened 2021-04-17
🚧 [ui] Prevent chain mismatch

ryardley:
Related to https://github.com/sifchain/issues/issues/316
Closes https://github.com/Sifchain/issues/issues/333

So https://github.com/Sifchain/sifnode/pull/1142 is a good first pass but it doesn't address what happens when we are using the wrong EVM/Sifchain network combination.

image

I chose a pink panel blocker and typically comforting Australian vernacular which is going to be in draft until we get some style love from @j-kappa 😆

image

Network mapping happens here: https://github.com/Sifchain/sifnode/pull/1147/files#diff-78224f1ed24df0b48ad95e72d4773254a84a56d8d313b8c5fc5e0aba84236f59R93-R98


ryardley:
@j-kappa the reason I went ahead and put this together over trying to manage within the micro UX was that this was the simplest way to accomplish things. It might be possible to still execute a swap when on the wrong network currently this is global but we can limit it to only showing up on `/peg` routes if required.

ryardley:
Have moved the blocker/overlay to within the dialog. This makes sense as we can still swap when connected to a valid sif network but should not be able to peg.

ryardley:
I attempted to specify the mainnet chain using the provider API but it wouldn't work:

image

intl-man:
@ryardley closing as the UI has since been decoupled.

PR #1155, opened 2021-04-20
export ethbridge keeper data to genesis.

juniuszhou:
To upgrade the cosmos sdk 0.42, ethbridge and oracle module need migrate the data in the keeper to new sifchain network. The pr is to export data to genesis file, then new version network can get the data from genesis.


x/ethbridge/genesis.go
```diff
@@ -0,0 +1,37 @@
...
+import (
+	"github.com/Sifchain/sifnode/x/ethbridge/types"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	abci "github.com/tendermint/tendermint/abci/types"
+)
+
+// InitGenesis import genesis data from GenesisState
+func InitGenesis(ctx sdk.Context, keeper Keeper, data types.GenesisState) (res []abci.ValidatorUpdate) {
+
+	for _, token := range data.PeggyTokens {
```
ElliotFriedman, line 12: My understanding was that we weren't migrating data from genesis, but rather from keepers that were already live and storing data in our system. Does this code address this?

ElliotFriedman (changes requested):

x/ethbridge/genesis.go
juniuszhou: the code just for 42 upgrade. export current data in keeper to genesis file and then new version sifchain can read the data from it. export command and sample genesis file put in the docs folder.

ElliotFriedman (approved):
LGTM!

ElliotFriedman (approved):

PR #1160, opened 2021-04-21
Update betanet.md

dolphintwo:
This operation made me jailed once. I hope others will not make this mistake.


intl-man (changes requested):
@dolphintwo Thanks for the PR, however, we need to ensure that the instructions are not coupled to a specific path. The previous step clearly states to run the command from within your running container.

Maybe include a couple of additional steps to determine the container name and the path to the running binary. Thoughts?

intl-man:
@dolphintwo am closing this as it's been updated (based on what you provided) in #1248 .

PR #1161, opened 2021-04-22
Python Dependencies

gboie:
Added standard requirements.txt  with any external dependencies needed to be install with pip3. These can be easily installed with: pip3 install -r requirements.txt


banshee:
Did you mean to pull in the 112_more_validators branch into this?  Seems like the requirements.txt thing is useful on its own

banshee (changes requested):
Can we just pull in test/integration/src/py/requirements.txt instead of the new env_* stuff?

PR #1162, opened 2021-04-22
[ChainOps] MongoDB updates.

intl-man:
Includes:

* Updated the resource limits for a mongodb container.


gzukel (approved):
approve looks good.

PR #1169, opened 2021-04-23
[ChainOps] Set the address prefix.

intl-man:
Includes:

* Minor update to correctly set the prefix as `sif`, not `cosmos` (default).
* Fixed the spelling mistake `.sifenoded` and replaced with `.sifnoded`.

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -142,7 +142,7 @@ func init() {
 		panic(err)
 	}
 
-	DefaultNodeHome = filepath.Join(userHomeDir, ".sifenoded")
+	DefaultNodeHome = filepath.Join(userHomeDir, ".sifnoded")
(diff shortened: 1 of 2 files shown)


app/app.go
```diff
@@ -142,7 +142,7 @@ func init() {
 		panic(err)
 	}
 
-	DefaultNodeHome = filepath.Join(userHomeDir, ".sifenoded")
+	DefaultNodeHome = filepath.Join(userHomeDir, ".sifnoded")
```
intl-man, line 145: @iverc @marbar3778 - not sure why we're doing this anyway, as `DefaultNodeHome` is already set on line 99...

iverc (approved):

PR #1170, opened 2021-04-23
Pipeline Development Work

gzukel:
Removed as much non ruby as possible
Consolidate all of the recent automation changes and made revisions to the way automation was being executed.
Added new rake tasks to facilitate vault pipelines.
Templated the helm yaml for ebrelayer.


intl-man (approved):
Nice work @gzukel.

PR #1172, opened 2021-04-23
Fix deployment script gas issue

ElliotFriedman:


banshee:
These scripts will never run again for mainnet, right?  This is only a change for test and dev nets?

ElliotFriedman:
> These scripts will never run again for mainnet, right? This is only a change for test and dev nets?

These scripts shouldn't be run against mainnet again. We will make new scripts if we want to do an upgrade.

This fix is just to ensure that we can deploy to testnets if their block gas limit is low.

intl-man (approved):

PR #1197, opened 2021-04-28
Refactor Retry Logic

ElliotFriedman:


cmd/ebrelayer/txs/relayToEthereum.go
```diff
@@ -67,52 +49,31 @@ func RelayProphecyClaimToEthereum(provider string, contractAddress common.Addres
...
-		}
-
-		if i == 30 {
-			sugaredLogger.Errorw("failed to send ProphecyClaim to CosmosBridge.",
-				errorMessageKey, err.Error())
-			return err
-		}
-	}
+	// sleep 30 seconds to wait for tx to go through.
+	time.Sleep(transactionInterval * 30)
```
banshee, line 55: Can we pass this in as a command line argument?  Having a fixed delay means the tests have to wait that long too.

cmd/ebrelayer/txs/relayToEthereum.go
```diff
@@ -67,52 +49,31 @@ func RelayProphecyClaimToEthereum(provider string, contractAddress common.Addres
...
 				"retry", i,
 				"getTxReceiptErr", err.Error(),
 			)
+			_, _ = cosmosBridgeInstance.NewProphecyClaim(auth, uint8(claim.ClaimType),
+				claim.CosmosSender, claim.CosmosSenderSequence, claim.EthereumReceiver, claim.Symbol, claim.Amount.BigInt())
 
-			receipt, err := client.TransactionReceipt(context.Background(), tx.Hash())
-			if err == nil {
+			receipt, err := client.TransactionReceipt(context.Background(), txHash)
+			if err == nil || receipt.Status == 1 {
```
banshee, line 76: I don't think we can do this.  If err is set, there's no promise that receipt is set to anything reasonable, up to and including nil

cmd/ebrelayer/txs/relayToEthereum.go
```diff
@@ -67,52 +49,31 @@ func RelayProphecyClaimToEthereum(provider string, contractAddress common.Addres
...
 	sugaredLogger.Infow("get NewProphecyClaim tx hash:", "ProphecyClaimHash", tx.Hash().Hex())
 
+	txHash := tx.Hash()
+
 	// Get the transaction receipt
-	receipt, err := client.TransactionReceipt(context.Background(), tx.Hash())
-	if err != nil {
+	receipt, err := client.TransactionReceipt(context.Background(), txHash)
+	// if there is an error getting the tx, or if the tx fails, retry 30 times
+	if err != nil || receipt.Status == 0 {
```
ElliotFriedman, line 64: Receipt status being 0 is not necessarily incorrect.

cmd/ebrelayer/relayer/cosmos.go
```diff
@@ -367,6 +369,50 @@ func getOracleClaimType(eventType string) types.Event {
...
+		cosmosBridgeInstance, err := cosmosbridge.NewCosmosBridge(target, client)
+		if err != nil {
+			sub.SugaredLogger.Errorw("failed to get cosmosBridge instance.",
+				errorMessageKey, err.Error())
+			continue
+		}
+		return cosmosBridgeInstance, nil
+	}
+
+	return nil, errors.New("hit max initRelayConfig retries")
```
juniuszhou, line 391: error message is wrong.

cmd/ebrelayer/relayer/cosmos.go
```diff
@@ -367,6 +369,50 @@ func getOracleClaimType(eventType string) types.Event {
...
+func tryGettingCosmosBridge(
+	target common.Address,
+	client *ethclient.Client,
+	sub CosmosSub,
+	) (
+		*cosmosbridge.CosmosBridge,
+		error,
+	) {
+
+	for i := 0; i < 5; i++ {
```
juniuszhou, line 381: not sure if retry is useful for it, get cosmos bridge just create some structure instance, it may fail if something as abi data not found. it doesn't depend on anything like network connection. if there is a case the we can't get the bridge contract instance at first, then we can successful get one later.

ElliotFriedman: I believe this call to NewCosmosBridge does require a network connection as it takes an ethereum client as a parameter.

ElliotFriedman: Removed.

cmd/ebrelayer/relayer/cosmos.go
ElliotFriedman: Fixed

cmd/ebrelayer/txs/relayToEthereum.go
ElliotFriedman: Fixed

cmd/ebrelayer/relayer/cosmos.go
```diff
@@ -377,10 +401,47 @@ func (sub CosmosSub) handleBurnLockMsg(cosmosMsg types.CosmosMsg, claimType type
...
+	}
 
-	err := txs.RelayProphecyClaimToEthereum(sub.EthProvider, sub.RegistryContractAddress,
-		claimType, prophecyClaim, sub.PrivateKey, sub.SugaredLogger)
+	// Initialize CosmosBridge instance
+	cosmosBridgeInstance, err := cosmosbridge.NewCosmosBridge(target, client)
 	if err != nil {
-		log.Println(err)
+		sub.SugaredLogger.Errorw("failed to get cosmosBridge instance.",
+			errorMessageKey, err.Error())
```
banshee, line 416: Don't we have to return here?  cosmosBridgeInstance isn't valid, so we can't pass it to RelayProphecyClaimToEthereum

cmd/ebrelayer/relayer/cosmos.go
```diff
@@ -377,10 +401,47 @@ func (sub CosmosSub) handleBurnLockMsg(cosmosMsg types.CosmosMsg, claimType type
...
+			)
+		} else {
+			break
+		}
+		i++
+	}
+
+	if i == maxRetries {
+		sub.SugaredLogger.Errorw(
+			"failed to broadcast transaction after 30 attempts",
```
banshee, line 443: Looks like 5, not 30

cmd/ebrelayer/txs/relayToEthereum.go
```diff
@@ -20,64 +23,87 @@ import (
...
+	)
+
+	tx, err := cosmosBridgeInstance.NewProphecyClaim(
+		auth, uint8(claim.ClaimType), claim.CosmosSender,
+		claim.CosmosSenderSequence, claim.EthereumReceiver,
+		claim.Symbol, claim.Amount.BigInt(),
+	)
+	
+	// sleep 2 seconds to wait for tx to go through.
+	sleepThread(2)
```
banshee, line 57: could move the sleep after the error check

banshee (changes requested):

cmd/ebrelayer/relayer/cosmos.go
ElliotFriedman: Return added.

cmd/ebrelayer/txs/relayToEthereum.go
ElliotFriedman: I moved the sleep here before the error check on purpose. This way, if infura is down, we wait two seconds before re-broadcasting that same transaction so it gives them time to go back online.

banshee (approved):

PR #1208, opened 2021-04-30
Update Relayer to Manually Manage Cosmos Sequence

ElliotFriedman:


banshee (changes requested):
If we still want this, we need to changes the branches - this is a PR to something we're not using any more

ElliotFriedman:
This has already been addressed in a separate PR. Closing.

PR #1227, opened 2021-05-03
[ui] Fix approval baseunits bug

ryardley:
Referenced by https://github.com/Sifchain/issues/issues/396

With reference to https://sifchain.slack.com/archives/G01BX71QY1F/p1619822131293500

Unfortunately [this fix](https://github.com/Sifchain/sifnode/pull/1222) solved the problem at the wrong level of abstraction missing the actual bug which was that the amount was not being converted to baseunits in the view code.

This fixes the bug in approval baseunit as well as adds an accompanying test. 

~I believe we will want to revert #1222 and merge this to the release from develop.~

This reverts #1222 and replaces it with a fix within the view.

This does not acknowledge the underlying architectural issue which is that approve should be called in the usecase not the view. 

---

Testing changes were done in order to make the test work.

To do this properly I had to add a test. I was finding that sometimes events didn't quite happen in the order we were expecting causing instabilities and using Promis.all around clicks and listening for pop ups helps. Metamask may have been caching its state between tests which can cause weirdness. I could not seem to add a beforeEach block to the current setup for some reason (likely a bug). So I restructured the tests to be outside of the describe block. This allowed beforeEach to run correctly. At the same time I ensured the tests were not using magic numbers for waiting for block confirmations and instead wait for the UI to say has succeeded.

### DIFF
--- a/ui/core/src/api/EthbridgeService/EthbridgeService.ts
+++ b/ui/core/src/api/EthbridgeService/EthbridgeService.ts
@@ -122,19 +122,15 @@ export default function createEthbridgeService({
   }
 
   return {
-    async approveBridgeBankSpend(account: string, assetAmount: IAssetAmount) {
+    async approveBridgeBankSpend(account: string, amount: IAssetAmount) {
       // This will popup an approval request in metamask
       const web3 = await ensureWeb3();
-      const tokenContract = await getTokenContract(
-        web3,
-        assetAmount.asset.address!,
-      );
+      const tokenContract = await getTokenContract(web3, amount.asset.address!);
       const sendArgs = {
         from: account,
         value: 0,
         gas: 100000,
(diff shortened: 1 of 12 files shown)


ui/core/src/api/EthbridgeService/EthbridgeService.ts
```diff
@@ -122,27 +122,28 @@ export default function createEthbridgeService({
   }
 
   return {
-    async approveBridgeBankSpend(account: string, assetAmount: IAssetAmount) {
+    async approveBridgeBankSpend(account: string, amount: IAssetAmount) {
```
59023g, line 125: I still like this param as `assetAmount` but it's nit

59023g (approved):
Excellent

PR #1228, opened 2021-05-03
[ui] random fixes to get develop running storybook correctly

thomasdavis:


ryardley:
Looks like there was a merge conflict in develop right now too with the `RewardsPage.vue` referencing actions which is causing tests to fail. I was wondering however with this this should probably have a CI workflow that at least builds storybook without error if not deploys it and will fail the build if it doesn't pass?

ryardley (approved):

59023g (approved):

PR #1234, opened 2021-05-03
Feature/add claim

Jedi2002:


x/dispensation/handler_test.go
```diff
@@ -32,3 +33,19 @@ func TestNewHandler(t *testing.T) {
...
+
+func TestNewHandler_CreateClaim(t *testing.T) {
+	app, ctx := test.CreateTestApp(false)
+	keeper := app.DispensationKeeper
+	handler := dispensation.NewHandler(keeper)
+	address := sdk.AccAddress(crypto.AddressHash([]byte("User1")))
+	msgClaim := types.NewMsgCreateClaim(address, types.ValidatorSubsidy)
+	res, err := handler(ctx, msgClaim)
+	require.NoError(t, err)
+	require.NotNil(t, res)
```
timlind, line 45: Are there other assertions that can be put here?

x/dispensation/keeper/distrubtions_test.go
```diff
@@ -0,0 +1,24 @@
...
+
+func TestKeeper_GetDistributions(t *testing.T) {
+	app, ctx := test.CreateTestApp(false)
+	keeper := app.DispensationKeeper
+	for i := 0; i < 10; i++ {
+		name := uuid.New().String()
+		distribution := types.NewDistribution(types.Airdrop, name)
+		err := keeper.SetDistribution(ctx, distribution)
+		assert.NoError(t, err)
+		_, err = keeper.GetDistribution(ctx, name, types.Airdrop)
```
timlind, line 19: Are there other assertion to put on distribution res?

timlind (dismissed):

x/dispensation/types/msgs.go
```diff
@@ -28,6 +28,7 @@ func (m MsgDistribution) Route() string {
 	return RouterKey
 }
 
+//TODO Replace with constant defined in keys.go with value CreateDispensation
```
iverc, line 31: Should we address it in this PR?

x/dispensation/types/querier.go
```diff
@@ -8,14 +8,15 @@ const (
 	QueryAllDistributions   = "distributions"
 	QueryRecordsByDistrName = "records_by_name"
 	QueryRecordsByRecipient = "records_by_recipient"
+	QueryClaimsByType       = "claims_by_type"
 )
 
 type QueryRecordsByDistributionName struct {
```
iverc, line 14: I believe we should define types in proto now.

x/dispensation/types/records.go
```diff
@@ -25,19 +26,20 @@ func (d ClaimStatus) String() string {
 }
 
 // DistributionRecord is created for every recipient for a distribution
-// TODO : Add ClaimStatus to the prefixed key for records
+// TODO : Add DistributionStatus to the prefixed key for records
```
iverc, line 29: Should it be addressed as part of this PR?

x/dispensation/types/querier.go
timlind: This may already be done in the gRPC PR

x/dispensation/keeper/distrubtions_test.go
Jedi2002: I added the test for GetDistributions and it has a check for the total list .

x/dispensation/types/msgs.go
Jedi2002: No , this is more of a cosmetic change , I put that there to keep track . Just trying to keep changes minimal right now , and add everything in bulk witht he .42 upgrade.

x/dispensation/types/records.go
Jedi2002: Same as above , ill make these changes in the .42 version

x/dispensation/handler_test.go
Jedi2002: We are checking the using GetClaim below , so the functionality works ,the claims get saved . 
Checking the hash and verifying it by querying the blockchain should come under the integration tests I guess.

marshsif (approved):

PR #1241, opened 2021-05-05
Feature/grpc dispensation records prefixed

Jedi2002:


scripts/demos/dispensation/queries.sh
```diff
@@ -1,5 +1,5 @@
 #!/usr/bin/env bash
 rm -rf all.json pending.json completed.json
-sifnodecli q dispensation records-by-name-all ar1 >> all.json
-sifnodecli q dispensation records-by-name-pending ar1 >> pending.json
-sifnodecli q dispensation records-by-name-completed ar1 >> completed.json
\ No newline at end of file
+sifnodecli q dispensation records-by-name ar1 All>> all.json
+sifnodecli q dispensation records-by-name ar1 Pending >> pending.json
+sifnodecli q dispensation records-by-name ar1 Completed>> completed.json
```
timlind, line 5: sifnoded not sifnodecli

x/dispensation/abci.go
```diff
@@ -7,5 +7,8 @@ import (
 
 func BeginBlocker(ctx sdk.Context, req abci.RequestBeginBlock, k Keeper) {
 	// Distribute drops if any are pending
-	_ = k.DistributeDrops(ctx, req.Header.Height)
+	err := k.DistributeDrops(ctx, req.Header.Height)
+	if err != nil {
+		return
```
timlind, line 12: Should we at least log the error?

x/dispensation/client/cli/query.go
```diff
@@ -21,8 +21,6 @@ func GetQueryCmd(queryRoute string) *cobra.Command {
 		GetCmdDistributions(queryRoute),
 		GetCmdDistributionRecordForRecipient(queryRoute),
 		GetCmdDistributionRecordForDistNameAll(queryRoute),
```
timlind, line 23: Should remove `All` suffix if it's filtering by status as well.

x/dispensation/client/cli/query.go
```diff
@@ -88,7 +86,7 @@ func GetCmdDistributionRecordForRecipient(queryRoute string) *cobra.Command {
 //GetCmdDistributionRecordForDistNameAll returns all records for a given distribution name
 func GetCmdDistributionRecordForDistNameAll(queryRoute string) *cobra.Command {
 	return &cobra.Command{
-		Use:   "records-by-name-all [distribution name]",
+		Use:   "records-by-name [distribution name] [status]",
 		Short: "get a list of all distribution records ",
 		Args:  cobra.ExactArgs(1),
```
timlind, line 91: Should be 2.

x/dispensation/client/cli/query.go
```diff
@@ -88,7 +86,7 @@ func GetCmdDistributionRecordForRecipient(queryRoute string) *cobra.Command {
 //GetCmdDistributionRecordForDistNameAll returns all records for a given distribution name
 func GetCmdDistributionRecordForDistNameAll(queryRoute string) *cobra.Command {
 	return &cobra.Command{
-		Use:   "records-by-name-all [distribution name]",
+		Use:   "records-by-name [distribution name] [status]",
```
timlind, line 89: Needs an example to indicate what status options are valid.

x/dispensation/keeper/distributionRecords.go
```diff
@@ -31,95 +31,111 @@ func (k Keeper) GetDistributionRecord(ctx sdk.Context, airdropName string, recip
...
+func (k Keeper) GetDistributionRecordsIterator(ctx sdk.Context, status types.DistributionStatus) sdk.Iterator {
 	store := ctx.KVStore(k.storeKey)
-	return sdk.KVStorePrefixIterator(store, types.DistributionRecordPrefix)
+	switch status {
+	case types.DistributionStatus_DISTRIBUTION_STATUS_PENDING:
+		return sdk.KVStorePrefixIterator(store, types.DistributionRecordPrefixPending)
+	case types.DistributionStatus_DISTRIBUTION_STATUS_COMPLETED:
+		return sdk.KVStorePrefixIterator(store, types.DistributionRecordPrefixCompleted)
+	default:
+		return sdk.KVStorePrefixIterator(store, types.DistributionRecordPrefixCompleted)
```
timlind, line 50: Is this a useful default or is it not better to return an iterator for unspecified status?

x/dispensation/keeper/grpc_querier.go
```diff
@@ -33,30 +31,20 @@ func (q Querier) AllDistributions(ctx context.Context,
...
 	records := new(types.DistributionRecords)
-	switch request.Status {
-	case types.ClaimStatus_CLAIM_STATUS_PENDING:
-		*records = q.keeper.GetRecordsForNamePending(sdk.UnwrapSDKContext(ctx), request.DistributionName)
-	case types.ClaimStatus_CLAIM_STATUS_COMPLETED:
-		*records = q.keeper.GetRecordsForNameCompleted(sdk.UnwrapSDKContext(ctx), request.DistributionName)
-	default:
-		*records = q.keeper.GetRecordsForNameAll(sdk.UnwrapSDKContext(ctx), request.DistributionName)
+	*records = q.keeper.GetRecordsForName(sdk.UnwrapSDKContext(ctx), request.DistributionName, request.Status)
+	if request.Status == types.DistributionStatus_DISTRIBUTION_STATUS_UNSPECIFIED {
```
timlind, line 35: Why are the pending records being added if request status is unspecified?

x/dispensation/client/cli/query.go
Jedi2002: the default returns unspecified , so technically everything is valid in this case

x/dispensation/keeper/grpc_querier.go
Jedi2002: User can either specify , Pending Completed , or anything else , I am assuming the user does 
records-by-name [distribution name].  , which I am assuming they are not interested in the status .

scripts/demos/clp/init.sh
```diff
@@ -1,31 +1,31 @@
...
-sifnodecli config output json
-sifnodecli config indent true
-sifnodecli config trust-node true
-sifnodecli config chain-id sifchain
-sifnodecli config keyring-backend test
+sifnoded config output json
+sifnoded config indent true
+sifnoded config trust-node true
+sifnoded config chain-id sifchain
+sifnoded config keyring-backend test
```
timlind, line 14: config command has been removed, so if you want this to work you'll have to add some of these args in the below commands i.e keyring

x/dispensation/client/cli/query.go
timlind: An example would indicate if an int or status string is expected.

timlind (approved):

PR #1243, opened 2021-05-05
Adding binance

gzukel:
added kms retrieve or creation to the vault install script just need … …
Updated Vault Deployment Pipeline
added initial binance helm chart
removed changed to repository image and tag in the values yaml for bi… …
removed executive cluster from logic only need betanet for the interm
added debug logging to kms key because it didn't replace the variable… …
removed kms debugging for vault kms generation
Initial helm charts for deploying the cryptoeconomics endpoint.
Merge branch 'feature/cryptoeconomics' into adding-binance
added crypto economics chart from Brents feature/cryptoeconomics bran… …


intl-man (approved):
LGTM. Nice work.

PR #1249, opened 2021-05-06
[ChainOps] Cryptoeconomics container resource tweaks.

intl-man:
Includes:

* Set a minimum of 8Gi and a max of 16Gi for the cryptoeconomics container.


gzukel (approved):
approved

PR #1253, opened 2021-05-06
Move instant stack docker container

ryardley:
This PR moves the docker image we base the instant stack on to the ghcr container registry we use for sifnode.

### DIFF
--- a/ui/scripts/build-stack-backend-docker.sh
+++ b/ui/scripts/build-stack-backend-docker.sh
@@ -3,5 +3,11 @@
 
 # this is temporarily being published on rudis docker hub account - @ryarfley
 # TODO: investigate publishing on sifchain repository (possibly GH repo)
-cd .. && docker build -f ./ui/scripts/stack.Dockerfile -t ryardley/sifdevstack01:experimental1 .
-docker push ryardley/sifdevstack01:experimental1
+
+
+NOW=$(date +%s)
+IMAGE_NAME=ghcr.io/sifchain/sifnode/ui-stack:$NOW
+cd .. && docker build -f ./ui/scripts/stack.Dockerfile -t $IMAGE_NAME .
+docker push $IMAGE_NAME
+
+echo $IMAGE_NAME > ./scripts/latest
--- a/ui/scripts/build-stack-backend-docker.sh
+++ b/ui/scripts/build-stack-backend-docker.sh
@@ -1,9 +1,18 @@
 #!/bin/bash
 
+if  [[ $(cat ~/.docker/config.json  | jq '.auths["ghcr.io"].auth') == 'null' ]]; then
+  echo "In order to run this script and push a new container to the github registry you need to create a personal access token and use it to login to ghcr with docker"
+  echo ""
+  echo "  echo \$MY_PAT | docker login ghcr.io -u USERNAME --password-stdin"
+  echo ""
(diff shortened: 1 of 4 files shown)


ui/README.md
```diff
@@ -106,6 +91,31 @@ Every part of this system is designed to facilitate easy testing.
...
+3. `yarn core:watch` - update the webserver with core code on change
+4. `yarn pw:test:debug` - Run the test suite over the stack - you probably want to isolate the test you need with `.only` methods.
+
+### To update instant stack
+
+If you need to update the instant stack image:
+
+1. Make changes to migrate scripts
+1. run `./scritps/run-stack-save-default-snapshot.sh`
+1. run `./scripts/build-stack-backend-docker.sh`
```
pgoos, line 115: Please fix numbering

ui/README.md
```diff
@@ -106,6 +91,31 @@ Every part of this system is designed to facilitate easy testing.
...
+
+Note you will need to ensure you have logged into the github container registry and have write access to the `/sifchain/sifnode` namespace.
```
pgoos, line 117: Just a question: how to verify that write access is granted (before running the scripts)?

ui/scripts/build-stack-backend-docker.sh
```diff
@@ -1,7 +1,27 @@
 #!/bin/bash
 
+if  [[ $(cat ~/.docker/config.json  | jq '.auths["ghcr.io"].auth') == 'null' ]]; then
+  echo "In order to run this script and push a new container to the github registry you need to create a personal access token and use it to login to ghcr with docker"
+  echo ""
+  echo "  echo \$MY_PAT | docker login ghcr.io -u USERNAME --password-stdin"
+  echo ""
+  echo "For more information see https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-container-registry#authenticating-to-the-container-registry"
```
pgoos, line 8: I think my answer for the above question is here at this link

ui/README.md
ryardley: Markdown will number elements for you?

pgoos (dismissed):
Approved - just some minor request to change but overall looking solid.

pgoos (approved):

59023g (approved):

59023g (dismissed):

khdegraaf:
Looks like tests are failing.  Maybe because they now depend on new docker image that hasn't been pushed yet?

ryardley:
> Maybe because they now depend on new docker image that hasn't been pushed yet?

Nah it's been pushed. Hopefully just a permissions issue

ryardley:
@khdegraaf Another pass over this might be to avoid building and pushing the image locally. That might be a security risk? Thinking about it this morning we probably only want to let GH actions build stuff. If so it is ok to spam the registry with a build for every commit hash and then just point to the one we want to base instant stack on? I thought the answer was probably no as it would probably cost money being a tonne of storage but if it is ok then that is a simple solution. If not then in my head the next step might be use commit hash as tag and have github actions pull the tag if exists or if not build it by checking out that commit hash and running docker build.

ryardley:
So this looks like the secret PAT we are using doesn't seem to have access to the scope. I am not sure I have the requisite permissions to look into this

khdegraaf:
OK, I'll debug this.

ryardley:
Closing in favour of #1282

PR #1272, opened 2021-05-08
Feature/grpc dispensation claims

Jedi2002:


proto/sifnode/dispensation/v1/query.proto
```diff
@@ -37,4 +39,11 @@ message QueryRecordsByRecipientAddrRequest { string address = 1; }
 message QueryRecordsByRecipientAddrResponse {
   DistributionRecords distribution_records = 1;
   int64 height = 2;
+}
+
+message QueryClaimsByTypeRequest { DistributionType user_claim_type = 1; }
```
timlind, line 44: Is user_claim_type a good name here? Isn't claim_type / distribution_type sufficient?

proto/sifnode/dispensation/v1/query.proto
```diff
@@ -37,4 +39,11 @@ message QueryRecordsByRecipientAddrRequest { string address = 1; }
...
+
+message QueryClaimsResponse {
```
timlind, line 46: Needs to be QueryClaimsByTypeResponse ... shouldn't be reused by other claim queries.

proto/sifnode/dispensation/v1/tx.proto
```diff
@@ -9,6 +9,8 @@ option go_package = "github.com/Sifchain/sifnode/x/dispensation/types";
 service Msg {
   rpc CreateDistribution(MsgCreateDistribution)
       returns (MsgCreateDistributionResponse);
+  rpc CreateClaim(MsgCreateUserClaim)
```
timlind, line 12: Can we keep consistency here, i.e CreateClaim(MsgCreateClaim)

proto/sifnode/dispensation/v1/types.proto
```diff
@@ -51,4 +56,15 @@ message Distributions { repeated Distribution distributions = 1; }
...
+}
+
+message UserClaim {
+  string user_address = 1;
+  DistributionType user_claim_type = 2;
+  string user_claim_time = 3;
+  bool locked = 4;
+}
+
+message UserClaims {
```
timlind, line 68: Can you justify why this is needed instead of just embedding repeated UserClaim in some proto?

timlind (approved):

x/dispensation/keeper/executors_test.go
```diff
@@ -30,8 +30,8 @@ func TestKeeper_AccumulateDrops(t *testing.T) {
 func TestKeeper_CreateAndDistributeDrops(t *testing.T) {
 	app, ctx := test.CreateTestApp(false)
 	keeper := app.DispensationKeeper
-	inputList := test.GenerateInputList("15000000000000000000")
-	outputList := test.GenerateOutputList("10000000000000000000")
+	inputList := test.CreatInputList(3, "15000000000000000000")
```
u50, line 33: `CreatInputList` should probably be `CreateInputList` - could be worth fixing the typo and then updating tests with the correct function name

proto/sifnode/dispensation/v1/query.proto
Jedi2002: Claims response ,is used to respond to QueryClaimsByTypeRequest for now . but in the future will also be used to return QueryClaimsbyAddress , and QueryClaimsByStatus  ,which both return a list of claims

proto/sifnode/dispensation/v1/types.proto
Jedi2002: THis was added to keep consitency witht he way DistributionRecords and Distributions are presently declared

proto/sifnode/dispensation/v1/tx.proto
Jedi2002: Ill change this to CreateUserClaim(MsgCreateUserClaim) . It will maintian consistency with the rest of the module

proto/sifnode/dispensation/v1/types.proto
timlind: Yeah I think we've done this the wrong way, will merge in as is and fix this all over in the PR it's going into.

PR #1281, opened 2021-05-09
add global setup and page objects for dex and extensions (keplr, metamask)

pgoos:
This PR implements global setup, global variables as well as page objects for both dex pages and metamask&keplr extensions
`e2e/global-setup.js` stores gobal setup
`e2e/.eslintrc.js` enables global variables:
```
  globals: {
    page: true,
    browser: true,
    context: true,
  },
```
This eliminates the need of setting up `browserContext` and each test suite as well as allows to use `page`, `context` and `browser` objects globally without the need of instantiating it in each test suite or single test
Page objects included under `e2e/pages`
Extensions - `keplr` and `metamask` have now their abstraction encapsulated in page objects and communication (connecting) between `dex` app and the extensions is achievied through helper methods inside `e2e/helpers.js`.
Page objects for Dex are now introduced:
`components`:
```
DexHeader.js
TokenSelection.js
```
`pages`:
```
ConfirmSupplyModal.js
ConfirmSwapModal.js
ConnectPopup.js
GenericPage.js
KeplrNotificationPopup.js
KeplrPage.js
MetaMaskPage.js
MetamaskConnectPage.js
MetamaskNotificationPage.js
PegPage.js
PoolPage.js
SwapPage.js
```
The main goal with introducing page objects is generally to cleanup the test scripts and provide simple page methods like `pegPage.peg("erowan", "100")` etc. to map to certain business actions.


ui/e2e/setup.js
```diff
@@ -0,0 +1,28 @@
...
+
+  context = await chromium.launchPersistentContext(userDataDir, {
+    // headless required with extensions. xvfb used for ci/cd
+    headless: false,
+    args: [
+      `--disable-extensions-except=${pathToKeplrExtension},${pathToMmExtension}`,
+      `--load-extension=${pathToKeplrExtension},${pathToMmExtension}`,
+    ],
+  });
+  [page] = await context.pages();
```
ryardley, line 27: This is a little thing but I hadn't read your description of this PR and was looking at the code and it was only once I came across this that I understood that you were saving page globally. It is not the most common js syntax to de structure an array to a global var but it is fine here. I think I would just love either a comment as I scan the code to ensure it is not a mistake. Something like `// exposing "page" object globally` and `// exposing "context" globally` alternatively you could do the following which also signals intent:
```
[globalThis.page] = await context.pages();
```

ui/e2e/pages/PoolPage.js
```diff
@@ -0,0 +1,144 @@
...
+        '[data-handle="pool-estimates-forwards-symbols"]',
+      poolEstimatesBackwardNumber:
+        '[data-handle="pool-estimates-backwards-number"]',
+      poolEstimatesBackwardSymbols:
+        '[data-handle="pool-estimates-backwards-symbols"]',
+      poolEstimatesShareNumber: '[data-handle="pool-estimates-share-number"]',
+      managePoolButton: (tokenA, tokenB) =>
+        `[data-handle="${tokenA}-${tokenB}-pool-list-item"]`,
+      totalPooled: (token) => `[data-handle="total-pooled-${token}"]`,
+      totalPoolShare: '[data-handle="total-pool-share"]',
```
ryardley, line 32: I really love the way you have consolidated our selectors like this. I am not crash hot on having pages abstracted away but I can live with it as it has been done in a way that is pretty clear however to me selector consolidation is the main value of this refactoring.

ui/e2e/pages/PegPage.js
```diff
@@ -0,0 +1,78 @@
...
+    const element = await page.$(this.el.assetAmount(asset));
+    await expect(element).toHaveText(expectedAmount);
+  }
+
+  async closeSubmissionWindow() {
+    await page.waitForTimeout(1000);
+    await page.click("text=×"); // sometimes clicking 'x' doesn't close the window (even if Playwright says it clicked). Waiting a bit helps
+  }
+
+  async verifyTransactionPending(asset) {
```
ryardley, line 71: I really like the convention of only putting expectations in verifyFoo methods

ryardley (commented):
Gonna hold off approving until the tests are passing. I was not convinced we needed to encapsulate our pages but I think consolidating selectors like that is awesome and tests are pretty clear now and this sets up a clear precedent for future work which is ace.

59023g (dismissed):

ui/e2e/setup.js
pgoos: Sure, thanks for this comment. For now I put the comments as you requested and in the near future, we would utilize global objects with `this.global.page` and `this.global.context` as we introduce custom environments like described in https://github.com/playwright-community/jest-playwright#usage-with-custom-testenvironment

intl-man:
@pgoos Closing as the UI has since been decoupled.

PR #1284, opened 2021-05-10
[ui] fixed warning when swapping to correctly say insufficient liquidity when necessary

thomasdavis:
This one wasn't due for this sprint but I keep losing development time when swapping because it's hard to immediately know why the swap button is disabled (a lot of the time it is insufficient liquidity but it says "Select tokens" instead. 

No need to merge it in, just wanted to get the PR up.


thomasdavis:
@59023g this was is good to review now. 

You can test it selecting two tokens to swap that have no liquidity. 

It should now say on the button label "Insufficient Liquidity"

ui/core/src/usecases/clp/calculators/swapCalculator.ts
```diff
@@ -257,12 +260,6 @@ export function useSwapCalculator(input: {
     if (!balance.value?.greaterThanOrEqual(fromField.fieldAmount.value || "0"))
       return SwapState.INSUFFICIENT_FUNDS;
 
-    if (
-      fromTokenLiquidity.lessThan(fromField.fieldAmount.value) ||
```
59023g, line 261: Why remove this? Seems like this should remain

ryardley:
Should we be writing a unit test for this change?

PR #1289, opened 2021-05-10
ebrelayer fixes to move eth => ceth for 0.42

banshee:
This branch has some changes that clearly need to come out (logging, etc) but I thought I'd go ahead and start with a PR for it.


app/app.go
```diff
@@ -130,6 +131,7 @@ var (
 
 		clptypes.ModuleName:     {authtypes.Burner, authtypes.Minter},
 		dispensation.ModuleName: {authtypes.Burner, authtypes.Minter},
+		ethbridge.ModuleName:    {authtypes.Burner, authtypes.Minter},
```
banshee, line 134: Is this the right thing to do?  It definitely works, but it seems strange it wasn't here already

app/app.go
```diff
@@ -467,6 +469,8 @@ func (app *SifchainApp) InitChainer(ctx sdk.Context, req abci.RequestInitChain)
 		panic(err)
 	}
 
+	genesisState["ethbridge"] = json.RawMessage{}
```
banshee, line 472: This is strange.  It's just a hack to get InitGenesis called - if the genesis data for "ethbridge" is empty, InitGenesis isn't called at all, and that's where we set up:

```
	bridgeAccount := authtypes.NewEmptyModuleAccount(ModuleName, authtypes.Burner, authtypes.Minter)
	am.AccountKeeper.SetModuleAccount(ctx, bridgeAccount)
```

juniuszhou: don't understand, why just clean the the genesis state of this module, if also need clean other module like oracle, clp.

app/app.go
timlind: Must've missed this one, ethbridge was one of the last modules we re-added once it was migrated.

x/ethbridge/client/cli/tx.go
```diff
@@ -30,7 +30,12 @@ func GetCmdCreateEthBridgeClaim() *cobra.Command {
 
 			flags := cmd.Flags()
 
-			ethereumChainID, err := flags.GetInt64(types.FlagEthereumChainID)
+			ethereumChainIDStr, err := flags.GetString(types.FlagEthereumChainID)
```
banshee, line 33: Why can't we just use GetInt64 here?

cmd/ebrelayer/main.go
```diff
@@ -206,8 +212,14 @@ func RunInitRelayerCmd(cmd *cobra.Command, args []string) error {
 	zap.RedirectStdLog(sugaredLogger.Desugar())
 
 	// Initialize new Ethereum event listener
+	//validatorAddress, err := relayer.AddToKeyringWithMnemonic(cliContext.Keyring, validatorMoniker, mnemonic)
+	//if err != nil {
+	//	panic("could not add relayer to keyring")
+	//}
+	//
+	//valAddrAsFrom := sdk.ValAddress(validatorAddress.GetAddress())
```
ElliotFriedman, line 220: Let's remove commented code

cmd/ebrelayer/main.go
```diff
@@ -187,7 +193,7 @@ func RunInitRelayerCmd(cmd *cobra.Command, args []string) error {
 		return errors.Errorf("invalid [validator-moniker]: %s", args[3])
 	}
 	validatorMoniker := args[3]
-	mnemonic := args[4]
+	//mnemonic := args[4]
```
ElliotFriedman, line 196: Let's remove commented code

cmd/ebrelayer/main.go
```diff
@@ -206,8 +212,14 @@ func RunInitRelayerCmd(cmd *cobra.Command, args []string) error {
...
 	// Initialize new Ethereum event listener
+	//validatorAddress, err := relayer.AddToKeyringWithMnemonic(cliContext.Keyring, validatorMoniker, mnemonic)
+	//if err != nil {
+	//	panic("could not add relayer to keyring")
+	//}
+	//
+	//valAddrAsFrom := sdk.ValAddress(validatorAddress.GetAddress())
 	ethSub, err := relayer.NewEthereumSub(cliContext, rpcURL, validatorMoniker, chainID, web3Provider,
-		contractAddress, privateKey, mnemonic, db, sugaredLogger)
+		contractAddress, privateKey, nil, db, sugaredLogger, "")
```
ElliotFriedman, line 222: If these parameters aren't needed, let's remove the params from `NewEthereumSub`.

cmd/ebrelayer/txs/relayToCosmos.go
```diff
@@ -49,20 +43,18 @@ func RelayToCosmos(factory tx.Factory, moniker, password string, claims []*types
...
 	}
 
-	sugaredLogger.Infow(
-		"relay sequenceNumber from builder.",
-		"nextSequenceNumber", nextSequenceNumber,
+	sugaredLogger.Infow("RelayToCosmos building, signing, and broadcasting", "messages", messages)
+	// TODO this WithGas isn't correct
+	// TODO we need to investigate retries
+	// TODO we need to investigate what happens when the transaction has already been completed
+	err := tx.BroadcastTx(
```
ElliotFriedman, line 50: Created a card for this in zenhub: https://app.zenhub.com/workspaces/sifchain-603ed608675da3001744d736/issues/sifchain/sifnode/1346 @banshee

cmd/ebrelayer/txs/relayToEthereum.go
```diff
@@ -45,30 +44,32 @@ func RelayProphecyClaimToEthereum(provider string, contractAddress common.Addres
...
-	if err != nil {
-		sugaredLogger.Errorw("failed convert Amount", errorMessageKey, err.Error())
-		return err
+	sugaredLogger.Infow(
+		"Sending new ProphecyClaim to CosmosBridge.",
+		"claim", claim,
+	)
+	// TODO this is leftover from when we were converting from a string; don't need it any more
+	var claimType ethbridge.ClaimType
+	switch claim.ClaimType {
```
ElliotFriedman, line 53: Like this todo says, we can probably change this away from a switch statement to a direct assignment like so:

`claimType = claim.ClaimType`

app/app.go
banshee: This was removed.

cmd/ebrelayer/main.go
banshee: Done

cmd/ebrelayer/main.go
banshee: I added more comments

cmd/ebrelayer/main.go
banshee: Added a comment; this needs to be addressed when we do replay

cmd/ebrelayer/txs/relayToEthereum.go
banshee: Cleaned up this code and added a failure if the claim type isn't lock or burn

PR #1292, opened 2021-05-11
[ui] remove negative sign on arb opps on stats page

thomasdavis:
image

@caseyturnerarrington Did you still want the red and green colors? I removed the negative sign.


caseyturnerarrington:
Yes read and green, no negative signs. So what you have above looks great

PR #1323, opened 2021-05-13
[hotfeature][ui][rewards] Validator Subsidy

59023g:
Adds validator subsidy data to the Rewards Page. Also DRYs it out into `RewardContainer`


mccallofthewild (approved):
Looks good. Some of the tooltip copy might not be english-teacher approved but the meaning is very clear. Passed the "click everywhere" test. Double checked that the numbers are consistent w/ Cryptoecon. All working as expected 👍

nikohosn (approved):

PR #1338, opened 2021-05-15
cosmos 42 - some protobuf fixes

banshee:


proto/sifnode/ethbridge/v1/types.proto
```diff
@@ -35,35 +35,33 @@ message EthBridgeClaim {
...
 
 // ProphecyClaim contains data required to make a ProphecyClaim
 message ProphecyClaim {
-  bytes cosmos_sender           = 1;
-  string symbol                 = 2;
-  string amount                 = 3;
-  bytes ethereum_receiver       = 4;
-  string claim_type             = 5;
-  bytes cosmos_sender_sequence  = 6;
+  bytes cosmos_sender = 1;
```
banshee, line 59: The tools now format our proto files differently, I didn't do this by hand

proto/sifnode/ethbridge/v1/types.proto
```diff
@@ -35,35 +35,33 @@ message EthBridgeClaim {
...
 enum ClaimType {
   // Unspecified claim type
   CLAIM_TYPE_UNSPECIFIED = 0;
-  // Lock claim type
-  CLAIM_TYPE_LOCK = 1;
   // Burn claim type
-  CLAIM_TYPE_BURN = 2;
+  CLAIM_TYPE_BURN = 1;
+  // Lock claim type
+  CLAIM_TYPE_LOCK = 2;
```
timlind, line 47: I think we need to look into what's happening here a bit more closely, probably especially around export / import and string serialisation. Would be good to see what our unit tests are or should be asserting here.

Was this required as a fix for getting transfers to work?

marshsif: is there a reason why the order is being changed? Changing the order of protobuf data elements shouldn't happen once they've been created, though I'm sure you know this.

marshsif (approved):
this looks good to me. mostly code format changes.

proto/sifnode/ethbridge/v1/types.proto
banshee: @marshsif this is values for an enum, not ordering of fields in a message

banshee: @timlind Yes, this was required for getting transfers to work.  I believe this proto file has to match our Solidity code since this the ClaimType used in EthBridgeClaim.  I might be confused about this though.

banshee: And this did need the test changes that were in feature/peggy42, I pulled those in

cmd/ebrelayer/txs/parser_test.go
```diff
@@ -128,46 +130,6 @@ func TestFailedLockEventToCosmosMsg(t *testing.T) {
 	require.Error(t, err)
 }
 
-func TestMsgBurnToProphecyClaim(t *testing.T) {
```
banshee, line 131: I pulled in the test changes from feature/peggy42, and in that branch ethbridge.ProphecyClaim has been removed, so the tests were also removed.

PR #1340, opened 2021-05-17
[ChainOps] Sync master to testnet.

intl-man:


banshee (changes requested):
We can't do this until the next release since master had develop merged in

PR #1343, opened 2021-05-17
[ui] Fix up error handling in swap

ryardley:
Closes https://github.com/Sifchain/issues/issues/240 

* Adds gas price to our local stack so it is possible to simulate out of gas responses.
* Builds a new stack image added alpine to cut image size by 1.5G
* Fixed bug in location of stack/latest
* Need docker with buildkit (ie 19+) to get slightly slimmer images
* Simplify view state in swap to be `"idle" | "confirm" | "submit" | "fail" | "success"` because we cannot go on having a view state for every type of error state. This PR makes it so the `txStatus` is sent to the confirm modal for swap. (This is kind of done in a cludgy way for add liquidity but should be refactored.)
* Refactored `AnimatedConfirmation` to JSX as switching based on combination variables required an intermediate computed prop which leads to a crap tonne of confusing boilerplate which we then get no type safety around. At the same time using css modules means we keep styles within JSX. 
* This works but is only half done. It is an intermediate way forward. A follow up PR to this will combine all modals to a single strategy and remove redundancy and probably add more tests. Ideally we would drive available states from the page as we don't know necessarily that page state shapes are universal in the future despite the fact they are similar enoughish now it is an incorrect assumption they will be like that throughout the app. We don't want to collect a massive list of random page states as it crosses domains.

Known issues:
* Not using JSX means sharing our types cannot benefit from TS deduction and we can't pass generics to downstream components which means more maintenance unless we either use JSX or define all possible confirmation states in the page.


ui/app/src/components/confirmationDialog/AnimatedConfirmation.vue
```diff
@@ -1,128 +1,160 @@
...
+      pre: JSX.Element;
+      fromAmount?: string;
+      toAmount?: string;
+      fromToken?: string;
+      toToken?: string;
+      post: JSX.Element;
+    }) => (
+      <div class={styles.text}>
+        <p>{p.header}</p>
+        <p class={styles.thin} data-handle="swap-message">
```
59023g, line 38: `styles` turns attached styles into object?

59023g:
Nice

ui/app/src/components/confirmationDialog/AnimatedConfirmation.vue
ryardley: yep just like css modules

ui/core/src/services/SifService/SifService.ts
```diff
@@ -328,7 +328,9 @@ export default function createSifService({
       }
       try {
         const fee = {
-          amount: coins(0, "rowan"),
+          // Keplr overwrites this in app but for unit/integration tests where we
+          // dont connect to keplr we need to specify an amount of rowan to pay for the fee.
+          amount: coins(250000, "rowan"),
```
ryardley, line 333: So this surprised me. Turns out keplr is filling this out for us. Once I enabled the gas on the node the integration test failed.

59023g (approved):

59023g:
Tested working

59023g:
Screen Shot 2021-05-18 at 8 00 13 PM

PR #1344, opened 2021-05-17
Try the github.token method with pull-request write

ryardley:
Gabriel said he managed to get docker pull working with the `github.token` feature but we have to particular about permissions. This attempts to see if we add pull request write to permissions for the flow if that results in the github.token working.


ryardley:
hmm doesn't seem to be working...

PR #1350, opened 2021-05-18
No need to check against breaking changes to protobuf files

banshee:
We can turn this on after we release protobuf files.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -127,5 +127,6 @@ proto-lint:
 .PHONY: proto-lint
 
 proto-check-breaking:
-	@$(DOCKER_BUF) breaking --against $(HTTPS_GIT)#branch=master
+	# we should turn this back on after our first release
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -127,5 +127,6 @@ proto-lint:
 .PHONY: proto-lint
 
 proto-check-breaking:
-	@$(DOCKER_BUF) breaking --against $(HTTPS_GIT)#branch=master
+	# we should turn this back on after our first release
+	# $(DOCKER_BUF) breaking --against $(HTTPS_GIT)#branch=master
```
banshee, line 131: Another way to do this is change to check against feature/cosmos-0.42 instead of master, but I think that's not useful and may be annoying.

timlind (approved):
lgtm

marshsif (commented):
this looks good to me.

PR #1352, opened 2021-05-19
[ui] adds netloss calculation to single pool page

thomasdavis:
image


59023g:
Issue with space between bottom and Block Explorer block, also button is off (see RewardsPage for 'more-info' button) and dotted line missing. See: Figma (attached below)
Screen Shot 2021-05-18 at 8 57 17 PM
Screen Shot 2021-05-18 at 9 11 09 PM

ui/app/src/views/SinglePool.vue
```diff
@@ -57,6 +78,27 @@ export default defineComponent({
       return t.imageUrl;
     });
 
+    watch([address, fromSymbol], async () => {
```
59023g, line 81: The data does not load when navigate from other pages. However it does when I reload the single pool page. I suggest wrapping this functionality then adding it to `onMounted` lifecycle method. See: RewardsPage.vue

59023g (changes requested):

59023g:
It now loads but there's some pixel-perfect design issues that are still bugging me (will DM you). Also, what if we add the `$` _after_ the number as in the other number values (have denom symbol follow)

59023g (dismissed):

59023g (approved):

PR #1355, opened 2021-05-19
[ui] Formatting in pool confirmation panel

ryardley:
Closes Sifchain/issues#408

In attempting to deliver this work correctly there were several architectural issues that required substantial rework. 
1. Converted a few components to use JSX as I was struggling to catch type errors while refactoring
2. Discovered an accidental SASS setup mistake which was causing a huge CSS bundle being bundled in with our styles. This had to be fixed here as it interfered with CSS modules which would mean new styles wouldn't work. Basically what was happening was that includes in the vue config were actually exporting styles which end up being included once and scoped for every component so we were getting a whole css reset being bundled in per each component. In the end this made our compiled CSS 90% smaller. 
3. Discovered and fixed an error where as a result of us not using JSX the format function was being called with no typesafety which meant that a string was being passed to it where it normally doesn't take strings. This happened to work because of a fluke in the way the format function was written which would not have been apparent if the function was refactored at some point in the future. This serves to highlight the importance of converting components slowly to JSX which would ensure these errors don't happen.
4. Consolidate info component for showing two values with tooltips 
5. Ensure amount api is used until values are displayed
6. Tidied up props where necessary as problems were uncovered one by one
7. Add new `trimMantissa:"integer"` mode to the asset amount format function to trim mantissa all the way to an integer if it happens to be an integer.


59023g:
I think it'd be great if this used a similar pattern as Swap.
Screen Shot 2021-05-20 at 5 02 31 PM

ryardley:
> I think it'd be great if this used a similar pattern as Swap.

Agreed will move back to WIP and implement

ryardley:
image

Created a shared component for use with both pool and swap.

ui/app/src/scss/typography.scss
```diff
@@ -1,6 +1,3 @@
-// Import fonts:
-@import url("https://fonts.googleapis.com/css2?family=PT+Serif:ital,wght@0,400;0,700;1,400;1,700&family=Rouge+Script&display=swap");
-@import url("https://fonts.googleapis.com/css2?family=Inter:wght@400;600;700&display=swap"rel="stylesheet");
```
ryardley, line 3: Not 100% sure if this writes an import statement but moving it out to the App component to be sure.

ui/app/vue.config.js
```diff
@@ -20,11 +20,8 @@ module.exports = {
     loaderOptions: {
       sass: {
         additionalData: `
-            @import "normalize-scss";
             @import "@/scss/typography.scss";
             @import "@/scss/variables.scss";
-            @import "@/scss/reset.scss";
-            @import "@/scss/utilities.scss";
```
ryardley, line 27: These were moved because they were writing out styles and everything here will be added to each individual component.

59023g (approved):

PR #1365, opened 2021-05-20
Peggy2.0 Parts 1 & 2

ElliotFriedman:


x/ethbridge/keeper/keeper.go
```diff
@@ -104,30 +104,31 @@ func (k Keeper) ProcessBurn(ctx sdk.Context, cosmosSender sdk.AccAddress, msg *t
...
+
+	} else {
+		if msg.Symbol == types.CethSymbol {
+			coins = sdk.NewCoins(sdk.NewCoin(types.CethSymbol, msg.CethAmount.Add(msg.Amount)))
+		} else {
+			coins = sdk.NewCoins(sdk.NewCoin(msg.Symbol, msg.Amount), sdk.NewCoin(types.CethSymbol, msg.CethAmount))
+		}
+	}
+
+	err := k.bankKeeper.SendCoinsFromAccountToModule(ctx, cosmosSender, types.ModuleName, coins)
```
timlind, line 127: What will happen to these funds after they are sent to module account?

Wasn't the previous code only sending to module account first, so that it could then send to ceth receive account?

Should this send maybe be within the else statement above?

Do we have a unit test case for this?

cmd/ebrelayer/relayer/cosmos.go
```diff
@@ -372,8 +374,33 @@ func getOracleClaimType(eventType string) types.Event {
 	return claimType
 }
 
+func tryInitRelayConfig(sub CosmosSub, claimType types.Event) (*ethclient.Client, *bind.TransactOpts, common.Address, error) {
```
marshsif, line 377: would 'retryInitRelayConfig' be a slightly better method name here?

cmd/ebrelayer/relayer/cosmos.go
```diff
@@ -382,16 +409,48 @@ func (sub CosmosSub) handleBurnLockMsg(cosmosMsg types.CosmosMsg, claimType type
...
+		sub.SugaredLogger.Errorw("failed to get cosmosBridge instance.",
+			errorMessageKey, err.Error())
+		return
+	}
+
+	maxRetries := 5
+	i := 0
+
+	for i < maxRetries {
+		err = txs.RelayProphecyClaimToEthereum(
```
marshsif, line 431: would it make sense for this loop to be its own method, mirroring the tryInitRelayConfig method?

x/ethbridge/keeper/keeper.go
```diff
@@ -104,30 +104,31 @@ func (k Keeper) ProcessBurn(ctx sdk.Context, cosmosSender sdk.AccAddress, msg *t
...
-	}
-
 	if k.IsCethReceiverAccountSet(ctx) {
 		coins = sdk.NewCoins(sdk.NewCoin(types.CethSymbol, msg.CethAmount))
-		err = k.bankKeeper.SendCoinsFromModuleToAccount(ctx, types.ModuleName, k.GetCethReceiverAccount(ctx), coins)
+
+		err := k.bankKeeper.SendCoins(ctx, cosmosSender, k.GetCethReceiverAccount(ctx), coins)
 		if err != nil {
-			logger.Error("failed to send ceth from module to account.",
+			logger.Error("failed to send ceth from account to account.",
```
marshsif, line 112: I know there could be security implications, but could this provide more details about each "account"? Otherwise it may not be very helpful as a log msg

x/ethbridge/keeper/keeper.go
```diff
@@ -143,25 +144,31 @@ func (k Keeper) ProcessBurn(ctx sdk.Context, cosmosSender sdk.AccAddress, msg *t
...
 
 	if k.IsCethReceiverAccountSet(ctx) {
 		coins = sdk.NewCoins(sdk.NewCoin(types.CethSymbol, msg.CethAmount))
-		err = k.bankKeeper.SendCoinsFromModuleToAccount(ctx, types.ModuleName, k.GetCethReceiverAccount(ctx), coins)
+
+		err := k.bankKeeper.SendCoins(ctx, cosmosSender, k.GetCethReceiverAccount(ctx), coins)
 		if err != nil {
-			logger.Error("failed to transfer ceth from module to account.",
+			logger.Error("failed to send ceth from account to account.",
 				errorMessageKey, err.Error())
```
marshsif, line 156: perhaps here as well, provide more info about each 'account'

x/ethbridge/keeper/keeper.go
banshee: FYI, there was some confusion about branches, and this change doesn't belong in this PR.  Elliot's aware and will fix.

cmd/ebrelayer/relayer/cosmos.go
ElliotFriedman: Adding Re-try implies that this method is only used for retrying, which is misleading as it is used for the initialization. If that fails, then this method is used for retrying.

cmd/ebrelayer/relayer/cosmos.go
ElliotFriedman: Having the loop in this method makes the code clearer.

x/ethbridge/keeper/keeper.go
ElliotFriedman: This is helpful context to log if there is a message. We will be able to find out what tx it came from in the sifnoded logs.

x/ethbridge/keeper/keeper.go
ElliotFriedman: This should be visible in the tx hash. The error should also give us information we need.

banshee (changes requested):
Merge #1392 first

x/ethbridge/keeper/keeper.go
juniuszhou: What will happen to these funds after they are sent to module account?
there are two cases for the fund after sent to module account.
1. if the ceth receiver not set, the fund include both ceth and other token. other token will be burned, ceth kept in the module's account.
2. if the ceth receiver set, the ceth go to the receiver's account directly. other token sent to module account then burn

Wasn't the previous code only sending to module account first, so that it could then send to ceth receive account?
yes, previoud code send to module, then from module send to ceth receiver. but in cosmos sdk .42, we can't send from module to account, the receiver need some priviledge. I didn't find out interface to configure it.

Should this send maybe be within the else statement above?
it should be outside of if/else, then we will miss send other token except ceth from account to module.
then we don't have balance for burn.

Do we have a unit test case for this?
yes, we have lock/burn test case.

PR #1366, opened 2021-05-21
Cosmos 42 restore ebrelayer's replay functions

banshee:


cmd/ebrelayer/replay.go
```diff
@@ -18,90 +21,88 @@ import (
...
+	}
+
+	// Load the validator's Ethereum private key from environment variables
+	privateKey, err := txs.LoadPrivateKey()
+	if err != nil {
+		return errors.Errorf("invalid [ETHEREUM_PRIVATE_KEY] environment variable")
+	}
+
+	// Parse flag --chain-id
+	chainID := viper.GetString(flags.FlagChainID)
```
timlind, line 37: Might want to replace viper call with cmd.Flags()

cmd/ebrelayer/replay.go
```diff
@@ -18,90 +21,88 @@ import (
...
+	if strings.TrimSpace(chainID) == "" {
+		return errors.Errorf("Must specify a 'chain-id'")
+	}
+
+	// Parse flag --rpc-url
+	rpcURL := viper.GetString(FlagRPCURL)
```
timlind, line 43: How does this interact with --node param and ternmintNode = args[0] defined below? Do we need / want more than one way to specify?

cmd/ebrelayer/replay.go
```diff
@@ -168,7 +169,7 @@ func RunReplayCosmosCmd(cmd *cobra.Command, args []string) error {
 }
 
 // RunListMissedCosmosEventCmd executes initRelayerCmd
-func RunListMissedCosmosEventCmd(cmd *cobra.Command, args []string) error {
+func RunListMissedCosmosEventCmd(_ *cobra.Command, args []string) error {
 	// Parse flag --rpc-url
 	rpcURL := viper.GetString(FlagRPCURL)
```
timlind, line 174: Replace viper usage and standardise rpc / node flag / arg

cmd/ebrelayer/replay.go
banshee: Good point, those were leftovers from the older version.

banshee:
Moved this to draft; Tim's right and the old-fashioned viper stuff needs to be removed.

cmd/ebrelayer/replay.go
banshee: The sdk is doing the viper calls for us in 42, removed this code.

cmd/ebrelayer/replay.go
banshee: Removed viper, still need to standardise rpc / node flag / arg

banshee: As far as I can tell, the right thing to use is the node argument, so I just removed rpc-url.

cmd/ebrelayer/replay.go
banshee: I removed the rpc-url flag completely.  Maybe it was useful before 42?

test/integration/src/py/test_ebrelayer_replay.py
```diff
@@ -0,0 +1,81 @@
...
+        ensure_relayer_restart
+):
+    starting_block = test_utilities.current_ethereum_block_number(smart_contracts_dir)
+    logging.info("stopping ebrelayer")
+    test_utilities.get_shell_output("pkill -9 ebrelayer || true")
+    request, credentials = build_request(smart_contracts_dir, source_ethereum_address, solidity_json_path)
+    logging.info("(no transactions should happen without a relayer)")
+    # test_utilities.whitelist_token(request.ethereum_symbol, request.smart_contracts_dir, True)
+    logging.info(f"send {request.amount / 10 ** 18} eth ({request.amount} wei) to {request.sifchain_address}")
+    test_utilities.send_from_ethereum_to_sifchain(request)
```
ElliotFriedman, line 49: Let's assert right here after 50 blocks that the tx to sifchain did not go through and then move on to the next steps of this test.

test/integration/src/py/test_ebrelayer_replay.py
```diff
@@ -0,0 +1,81 @@
...
+    mon = test_utilities.get_required_env_var("MONIKER")
+    mn = test_utilities.get_required_env_var("MNEMONIC")
+    cn = test_utilities.get_required_env_var("CHAINNET")
+    ending_block = test_utilities.current_ethereum_block_number(smart_contracts_dir) + 1
+    cmd = f"""yes | ebrelayer replayEthereum tcp://0.0.0.0:26657 {ews} {bra} {mon} '{mn}' {starting_block} {ending_block} 1 2 --chain-id {cn} --gas 5000000000000 \
+ --keyring-backend test --node tcp://0.0.0.0:26657 --from {mon}"""
+    test_utilities.get_shell_output(cmd)
+    time.sleep(15)
+    logging.info(f"check the ending balance of {request.sifchain_address} after replaying blocks")
+    ending_balance = test_utilities.get_sifchain_addr_balance(request.sifchain_address, request.sifnodecli_node,
```
ElliotFriedman, line 64: We should add this balance check before restarting the relayer to ensure the balance didn't increase.

ElliotFriedman (changes requested):

test/integration/src/py/test_ebrelayer_replay.py
banshee: Done

test/integration/src/py/test_ebrelayer_replay.py
banshee: Done

banshee:
Temporarily setting this to draft; I think we should merge #1374 first

banshee:
> Temporarily setting this to draft; I think we should merge #1374 first

That's been merged and this still passes tests, so it's out of draft.

ElliotFriedman:
Looks good to me!

ElliotFriedman (approved):

PR #1371, opened 2021-05-21
Test add-genesis-validator cmd

timlind:
This tests the cmd interface for add-genesis-validator and the init genesis in oracle module, using a temporary home directory.

A regression test for previous bug.

### DIFF
--- a/cmd/sifnoded/cmd/add_genesis_validator_test.go
+++ b/cmd/sifnoded/cmd/add_genesis_validator_test.go
@@ -1,0 +1,66 @@
+package cmd_test
+
+import (
+	"bytes"
+	"io/ioutil"
+	"os"
+	"testing"
+
+	"github.com/cosmos/cosmos-sdk/server"
+	svrcmd "github.com/cosmos/cosmos-sdk/server/cmd"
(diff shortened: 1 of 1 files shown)


cmd/sifnoded/cmd/add_genesis_validator_test.go
```diff
@@ -0,0 +1,66 @@
...
+	initCmd.SetArgs([]string{"init", "test", "--home=" + homeDir})
+
+	app.SetConfig(true)
+	expectedValidatorBech32 := "sifvaloper1rwqp4q88ue83ag3kgnmxxypq0td59df4782tjn"
+	expectedValidator, err := sdk.ValAddressFromBech32(expectedValidatorBech32)
+	require.NoError(t, err)
+
+	addValCmd, _ := sifnodedcmd.NewRootCmd()
+	addValBuf := new(bytes.Buffer)
+	addValCmd.SetOut(addValBuf)
```
banshee, line 40: minor: FYI, these only set stdout and stderr for the commands themselves - it's basically just for cobra useage output, not command stdout and stderr.  I think these are no-ops in the test.

banshee (approved):

PR #1392, opened 2021-05-26
Minor improvements for liquidity pool tests

banshee:
Need to pull out faucet tests since we don't have a faucet


test/integration/src/py/fail_test_liquidity_pools.py
```diff
@@ -59,12 +52,8 @@ def create_pool(
         f"--home {credentials.sifnodecli_homedir} ",
         "-y --output json"
     ])
-    json_str = get_shell_output_json(cmd)
```
ElliotFriedman, line 62: Why does this need to get deleted?

AustinoBombino (approved):

PR #1403, opened 2021-05-28
Fixed issues with ebrelayer tests caused by 42 timing changes and fixed issues from a merge of the ebrelayer retry work

banshee:
Fixed issues with ebrelayer tests caused by 42 timing changes and fixed issues from a merge of the ebrelayer retry work

### DIFF
--- a/cmd/ebrelayer/main.go
+++ b/cmd/ebrelayer/main.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"fmt"
+	"github.com/Sifchain/sifnode/cmd/ebrelayer/txs"
 	"log"
 	"net/url"
 	"os"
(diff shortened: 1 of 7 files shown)


cmd/ebrelayer/main.go
```diff
@@ -204,8 +205,13 @@ func RunInitRelayerCmd(cmd *cobra.Command, args []string) error {
 		sugaredLogger,
 	)
 
+	key, err := txs.LoadPrivateKey()
```
banshee, line 208: Adding the private key here since CosmosSub has always had a PrivateKey field, and NewCosmosSub should take that key

ElliotFriedman (approved):
LGTM!

PR #1434, opened 2021-06-03
dispensation upgrade / migration

timlind:
This fixes the panic issue on loading distribution records. It also introduces a Distribution.Runner address so that distributions can be run in a handler instead of begin block, which has been merged in here.


marshsif (commented):
This looks good to me. Very clean test bodies.

Jedi2002 (approved):

PR #1443, opened 2021-06-04
Fix/dispensation-run-handler

timlind:
This Pr moves the processing of stored distribution drops out of the begin block function, and introducing a new run handler to process a round of 10 drops. 

An authorised runner address has been added to creation of distribution, which allows the investor to delegate running the drops to a third party. 

The distribution record stores that authorised runner, and requires a transaction signed by this runner to process drops of a given name and type that has that runner stored in the drop.


marshsif (commented):
reviewed. looks good to me, based on our conversation yesterday.

PR #1445, opened 2021-06-04
Removed a test for ebrelayer replay that wasn't valid.

banshee:


test/integration/src/py/test_ebrelayer_replay.py
```diff
@@ -78,11 +78,3 @@ def test_transfer_eth_to_ceth_using_replay_blocks(
     ending_balance2 = test_utilities.get_sifchain_addr_balance(request.sifchain_address, request.sifnodecli_node,
                                                                request.sifchain_symbol)
     assert (ending_balance2 == request.amount)
-
```
banshee, line 81: Since ebrelayer doesn't record the block number in relayerdb, we shouldn't be doing this last part of the test.

iverc: 👍

iverc (approved):

ElliotFriedman (approved):

PR #1453, opened 2021-06-07
Merge develop into 42 upgrade work

banshee:
This is the last commit on develop that merges cleanly, might as well pull it in.  The next commit on develop (38a836f368ab48a3f480e033a5b740b5f08fa297) has conflicts.


marshsif (commented):
should we address the lint issues?

timlind (approved):

banshee:
> should we address the lint issues?

Yes, but not here.  They're already in develop.

PR #1454, opened 2021-06-07
ebrelayer meeting changes

banshee:
During a meeting to talk about the relayer, we found some unused code.


marshsif (approved):
These have been ID'd as no longer in use, therefore they are being removed.

juniuszhou (approved):

proto/sifnode/ethbridge/v1/types.proto
```diff
@@ -47,13 +47,6 @@ enum ClaimType {
...
 
-// OracleClaim contains data required to make an OracleClaim
-message OracleClaim {
-  string prophecy_id = 1;
-  bytes message = 2;
-  bytes signature = 3;
-}
-
 // ProphecyClaim contains data required to make a ProphecyClaim
 message ProphecyClaim {
```
timlind, line 51: Prophecy Claim can also be removed here.

timlind (changes requested):

timlind (approved):

PR #1457, opened 2021-06-07
Update fees to assume 40gwei gas price

ElliotFriedman:


jkbrooks (approved):

juniuszhou (dismissed):

banshee (approved):

banshee (approved):

banshee (dismissed):

ryardley:
Is there a possibility we can query sifnode for the fee amount? This state should only live in one place moving forward or we will find ourselves in a difficult situation. If sifnode could release this change with a module route that returns the amount we can build logic in place to try to fetch the route and if it fails default to what we have so that way we can stay in sync and not break production

juniuszhou (approved):

PR #1458, opened 2021-06-08
[ui] Remove UI code and refactor remaining scripts for maintainability

ryardley:
This PR will remove the UI code from here in leu of migration to the sifchain-ui repo

## TLDR

* Deletes all UI code from the sifnode repo. Gone are `./ui/core` and `./ui/app`
* Deletes all e2e tests from the sifnode repo. 
* Refactors the scripts on the sifnode repo used for creating the ui-stack
* Appends the docker ui-stack image creator to include **remote** testing of `sifchain-ui`

## Reviewing this PR

To review this PR you can avoid looking at all the delete noise by looking at this diff: 

https://github.com/Sifchain/sifnode/compare/8fc8203..ryardley/438-remove-ui-code

## New commands

| Script                       | Description                                                                                                                  |
| ---------------------------- | ---------------------------------------------------------------------------------------------------------------------------- |
| `yarn stack`                 | Launch stack from scratch building sifnode.                                                                                  |
| `yarn start`                 | Start the stack from the saved snapshot                                                                                      |
| `yarn build`                 | Build and prepare the chains                                                                                                 |
| `yarn stack --save-snapshot` | Launch the stack then save the snapshot files                                                                                |
| `yarn stack --push`          | Take snapshot files that have already been saved and wrap them in a docker image and push it to the registry for consumption |

### [👉👉 Read the new docs here](https://github.com/Sifchain/sifnode/blob/ryardley/438-remove-ui-code/ui/README.md)

Tested and works on branch so should work for develop


59023g:
> Might want to wait until we are deploying 100% from the UI before merging this.

Agreed. Waiting on a testnet deployment that's supposed to go out today.

banshee:
I don't understand this:

TLDR
Deletes all UI code from the sifnode repo. Gone are ./ui/core and ./ui/app
Refactors the scripts on the sifnode repo used for creating the ui-stack

I expected to see:
```git rm -rf ui```

There shouldn't be ui code in the sifnode repo any more.

ryardley:
> There shouldn't be ui code in the sifnode repo any more.

So on the frontend we need to test against backend releases to ensure we are compatible. As well as this we need to ensure the backend tests itself against live versions of the frontend before it can be release. The only code here that is left generates a docker image after code has been merged to develop or master using the binary put together from the change. That docker image is used by the frontend to test against. This in my opinion is an interim step as eventually we can create a third repo to do this and use repository_dispatch to communicate between these repos however the added complexity in setting this up is not really worth the benefit at this time.

ryardley:
Ok so I have added the code to test against `sifchain-ui` on merge to `develop` and `master`

59023g (dismissed):

ryardley:
@59023g I just added test tables to the docker image. We will want to keep those in there - currently they are duplicated in the ui repo but they should come from here and be synced that way. We can extract them at a later point

59023g (approved):

PR #1474, opened 2021-06-09
Fix cosmos 0.42 test bulk transfers

timlind:
This is a WIP PR that currently just runs the test on CI.


timlind:
This test shouldn't run on CI and should be tested against test net.

PR #1484, opened 2021-06-10
Feature/390-pgoos-e2e-swap-normalization-pools

pgoos:
This PR creates liquidity pool for test token `cTest` as well as adds the appropriate balance to all `Akasha`, `Shadowfiend` and `Juniper` accounts. `cTest` is a 12 digit token that could be then used in `e2e` tests, for example, to test if normalization of assets is done properly.


ryardley:
Just an FYI you don't need to include the tar files actually but it doesnt matter now.

ryardley (approved):
Looks great!

59023g:
This looks good, should it be in `sifchain-ui` repo?

pgoos:
> This looks good, should it be in `sifchain-ui` repo?

Actually it should be here. We are pushing ui-stack snapshots from `sifnode` repo. Then, updated stack is used in https://github.com/Sifchain/sifchain-ui/pull/73 (`sifchain-ui`). Stack PR needs to be merged first, then we can merge ui PR

PR #1511, opened 2021-06-16
Add prophecies to oracle genesis cosmos-0.42

timlind:


marshsif (approved):

x/oracle/keeper/keeper.go
```diff
@@ -60,15 +78,19 @@ func (k Keeper) GetProphecy(ctx sdk.Context, id string) (types.Prophecy, bool) {
 	return deSerializedProphecy, true
 }
 
-// setProphecy saves a prophecy with an initial claim
-func (k Keeper) setProphecy(ctx sdk.Context, prophecy types.Prophecy) {
-	store := ctx.KVStore(k.storeKey)
-	serializedProphecy, err := prophecy.SerializeForDB()
+// SetProphecy saves a prophecy with an initial claim
+func (k Keeper) SetProphecy(ctx sdk.Context, prophecy types.Prophecy) {
+	dbProphecy, err := prophecy.SerializeForDB()
```
Jedi2002, line 83: I know this is older code ,but it might make sense to return the error here , instead of panic. We can always panic at a higher level if needed , or just fail the transaction if possible .

Jedi2002 (commented):
Looks good ,left a minor comment. 

@timlind @marshsif  We should redefine the lint rules maybe , because some of them are not usefull for us . 

If not you might want to fix the lint errors before merge .

PR #1522, opened 2021-06-18
Add new tests | Fix existing tests

iverc:


intl-man:
@iverc Nice work. One thing I did mention to @timlind was that we should update the go version to 1.16, as `io/fs`, which is used in `cmd/sifnoded/migrate_test.go` is only available in 1.16, and the project `go.mod` is using 1.15.

output.json
```diff
@@ -0,0 +1,22 @@
+{
```
intl-man, line 1: Can we find a better place for this to live? The project root is not the place for this file.

iverc: @intl-man Agree, dispensation module would be a better place for it. Thanks!

PR #1531, opened 2021-06-22
Feature/single signer 0.42

Jedi2002:


marshsif (approved):
I only gave these a cursory review.

x/dispensation/client/cli/query.go
```diff
@@ -89,7 +89,7 @@ func GetCmdDistributionRecordForRecipient(queryRoute string) *cobra.Command {
 func GetCmdDistributionRecordForDistName(queryRoute string) *cobra.Command {
 	return &cobra.Command{
 		Use:   "records-by-name [distribution name] [status]",
-		Short: "get a list of all distribution records .Status : [Completed/Pending/All]",
+		Short: "get a list of all distribution records Status : [Completed/Pending/All]",
```
timlind, line 92: "get a list of all distribution records *by* status (completed/pending/all)" <- does the caller pass in these words for status? If not needs a numbers -> status mapping in the description.

x/dispensation/handler_test.go
```diff
@@ -47,6 +84,5 @@ func TestNewHandler_CreateClaim(t *testing.T) {
 
 	cl, err := keeper.GetClaim(ctx, address.String(), types.DistributionType_DISTRIBUTION_TYPE_VALIDATOR_SUBSIDY)
 	require.NoError(t, err)
-	assert.False(t, cl.Locked)
 	assert.Equal(t, cl.UserAddress, address.String())
 }
```
timlind, line 88: Is there a unit test for run handler?

x/dispensation/keeper/distributionRecords.go
```diff
@@ -31,9 +31,12 @@ func (k Keeper) GetDistributionRecord(ctx sdk.Context, airdropName string, recip
...
 
-func (k Keeper) ExistsDistributionRecord(ctx sdk.Context, airdropName string, recipientAddress string, status types.DistributionStatus) bool {
-	key := types.GetDistributionRecordKey(status, airdropName, recipientAddress)
-	return k.Exists(ctx, key)
+func (k Keeper) ExistsDistributionRecord(ctx sdk.Context, airdropName string, recipientAddress string, status types.DistributionStatus, distributionType types.DistributionType) bool {
+	key := types.GetDistributionRecordKey(status, airdropName, recipientAddress, distributionType)
+	if k.Exists(ctx, key) {
+		return true
+	}
+	return false
```
timlind, line 39: Could rather be `return k.Exsists()`

x/dispensation/keeper/distributionRecords.go
```diff
@@ -159,5 +231,68 @@ func (k Keeper) GetRecords(ctx sdk.Context) *types.DistributionRecords {
 		k.cdc.MustUnmarshalBinaryBare(bytesValue, &dr)
 		res.DistributionRecords = append(res.DistributionRecords, &dr)
 	}
+	iterator = k.GetDistributionRecordsIterator(ctx, types.DistributionStatus_DISTRIBUTION_STATUS_FAILED)
+	for ; iterator.Valid(); iterator.Next() {
+		var dr types.DistributionRecord
+		bytesValue := iterator.Value()
+		k.cdc.MustUnmarshalBinaryBare(bytesValue, &dr)
+		res.DistributionRecords = append(res.DistributionRecords, &dr)
+	}
```
timlind, line 240: Do all records not share a common prefix regardless of status, that can be used here? Isn't it possible to have status = unspecified that isn't iterated here?

x/dispensation/types/records.go
```diff
@@ -28,8 +38,15 @@ func (dr DistributionRecord) Add(dr2 DistributionRecord) DistributionRecord {
 	return dr
 }
 
+func (dr DistributionRecord) DoesClaimExist() bool {
```
timlind, line 41: This function name doesn't seem to indicate what it's doing, it's not checking if a claim exists in the store.

x/dispensation/types/records.go
```diff
@@ -59,12 +72,14 @@ func GetDistributionStatus(status string) DistributionStatus {
 		return DistributionStatus_DISTRIBUTION_STATUS_COMPLETED
 	case "Pending":
 		return DistributionStatus_DISTRIBUTION_STATUS_PENDING
+	case "Failed":
+		return DistributionStatus_DISTRIBUTION_STATUS_FAILED
 	default:
 		return DistributionStatus_DISTRIBUTION_STATUS_UNSPECIFIED
```
timlind, line 78: Can consider returning a bool here as is done with GetClaimType

x/dispensation/types/records.go
```diff
@@ -87,3 +102,26 @@ func IsValidDistribution(distributionType string) (DistributionType, bool) {
 		return DistributionType_DISTRIBUTION_TYPE_UNSPECIFIED, false
 	}
 }
+func IsValidDistribution(distributionType string) (DistributionType, bool) {
```
timlind, line 105: should be called isValidDistributionType

x/dispensation/types/records.go
```diff
@@ -87,3 +102,26 @@ func IsValidDistribution(distributionType string) (DistributionType, bool) {
...
+	case "DISTRIBUTION_TYPE_LIQUIDITY_MINING":
+		return DistributionType_DISTRIBUTION_TYPE_LIQUIDITY_MINING, true
+	case "DISTRIBUTION_TYPE_VALIDATOR_SUBSIDY":
+		return DistributionType_DISTRIBUTION_TYPE_VALIDATOR_SUBSIDY, true
+	default:
+		return DistributionType_DISTRIBUTION_TYPE_UNSPECIFIED, false
+	}
+}
+
+func IsValidClaim(claimType string) (DistributionType, bool) {
```
timlind, line 118: Should be called IsValidClaimType() ... or even better, converted to enum then use claimType.Valid()

x/dispensation/client/cli/query.go
Jedi2002: They do pass in the words , which would be more intuitive to use compared to numbers 
```
func GetDistributionStatus(status string) DistributionStatus {
	switch status {
	case "Completed":
		return DistributionStatus_DISTRIBUTION_STATUS_COMPLETED
	case "Pending":
		return DistributionStatus_DISTRIBUTION_STATUS_PENDING
	case "Failed":
		return DistributionStatus_DISTRIBUTION_STATUS_FAILED
	default:
		return DistributionStatus_DISTRIBUTION_STATUS_UNSPECIFIED
	}
}
```

x/dispensation/keeper/distributionRecords.go
Jedi2002: No , records are in different prefixes . This was done to avoid unnessary iteration over Completed and Failed records when running dispesations .
The unspecified status is never set by us

x/dispensation/types/records.go
Jedi2002: Its called from distribute drops .  The although doesnt query the keeper , it certifies that the claim should exist . 
Alternatively I can rename this to something like DoesTypeSupportClaim()

x/dispensation/types/records.go
Jedi2002: I am still returning Pending records , if the status in unspecified (or anything other than the required ones ) . I can change it to return error to the  query if the status does not match the three prefixes specified

x/dispensation/types/records.go
Jedi2002: I changed the name , not sure if I understand the second option 
Claim and Distribution both use the same type , and Airdrop is not Valid for distribution.  The Valid function would need a parameter in that case .

cmd/sifnoded/cmd/migrate_test.go
```diff
@@ -1,49 +0,0 @@
-package cmd
```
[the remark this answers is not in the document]
Jedi2002: Was deleted by mistake , adding it back .

x/dispensation/handler_test.go
Jedi2002: Added .

x/dispensation/types/records.go
Jedi2002: changed it to return error if user queries with unknown status

timlind (approved):

timlind:
@Jedi2002 Does this include any new integration test code or updates required to those? Do we need to get work there ported as well?

PR #1542, opened 2021-06-24
Fix remaining issues in 0.42

iverc:
- Fix remaining warnings
- Fix unit tests & related issues
- Pull in docker fixes (integration tests fixees)


intl-man (approved):
LGTM.

x/dispensation/utils/parser_test.go
```diff
@@ -77,3 +77,43 @@ func TestParseOutput(t *testing.T) {
 	assert.NoError(t, err)
 	assert.Equal(t, len(outputs), count)
 }
+
+func TestSplitBetweenReciepients(t *testing.T) {
```
Jedi2002, line 81: We can remove this , its not relevent any more

Jedi2002 (commented):
Looks great ,  minor comment which should be addressed before merging

x/oracle/genesis.go
```diff
@@ -11,20 +11,21 @@ import (
...
-
-	var wl []sdk.ValAddress
-	for _, addr := range data.AddressWhitelist {
-		if len(strings.TrimSpace(addr)) == 0 {
-			continue
+	if data.AddressWhitelist != nil {
+		wl := make([]sdk.ValAddress, len(data.AddressWhitelist))
+		for i, entry := range data.AddressWhitelist {
+			if len(strings.TrimSpace(entry)) == 0 {
+				continue
```
timlind, line 18: This introduces a bug where wl has an zero value entry created by make() but not assigned here, which is why the code was using an unallocated slice.

timlind (changes requested):

timlind (approved):

PR #1552, opened 2021-06-26
Fix up ui-stack

ryardley:
1. Fix up how `sifnoded` was being called
2. Copied the app.toml and enabled the rest api and updated the cors settings to allow for unsafe cors 
3. Removes a few references to sifnodecli
4. Added this branch temporarily to the ui-stack build list so we can have a docker image to get testing against. 
5. Fixed up migration scripts to support new sifchain commands
6. Removed stuff I suspect is unnecessary 
7. Fixed up bug where docker image stable tag could not be made from branches with multiple slashes in their names


ryardley:
@timlind be great if you could let me know if there is anything I am missing here

timlind (approved):

PR #1559, opened 2021-06-29
Future/peggy2hardhat

banshee:


ElliotFriedman (approved):
LGTM! Still need to migrate test fixtures to use waffle and HRE.

PR #1596, opened 2021-07-09
Token migration

juniuszhou:
migrate the token symbol and its balance.


ElliotFriedman (changes requested):
Who is allowed to call this migrate function?

juniuszhou:
> Who is allowed to call this migrate function?
My understanding is the functions called by cosmos sdk when upgrade the network. they are not triggered by any account or transaction.

ElliotFriedman (approved):

ElliotFriedman:
Could you please add an actual naming doc if we used the coin metadata convention described in this doc: https://docs.google.com/document/d/1DFYQhiSEArzsDlDfL5GgTnjrAwpz7iS81nCkbhZYCX0/edit?ts=60feb9ef#

juniuszhou:
> Could you please add an actual naming doc if we used the coin metadata convention described in this doc: https://docs.google.com/document/d/1DFYQhiSEArzsDlDfL5GgTnjrAwpz7iS81nCkbhZYCX0/edit?ts=60feb9ef#

so the actual demon will be the hash ba6973c40816e27ad5019d6ae10c67ccb0eef1d67ac0bd2c17425a17680eba3f, will update it.

PR #1605, opened 2021-07-12
Denom whitelist module

timlind:


x/whitelist/keeper/keeper.go
```diff
@@ -0,0 +1,111 @@
...
+	wl := k.GetDenomWhitelist(ctx)
+
+	for i := range wl.DenomWhitelistEntries {
+		if wl.DenomWhitelistEntries[i].Denom == denom &&
+			wl.DenomWhitelistEntries[i] != nil {
+			return *wl.DenomWhitelistEntries[i]
+		}
+	}
+
+	return types.DenomWhitelistEntry{
```
Jedi2002, line 64: Cant we just return bool ?

Jedi2002: Oh is it becuase you want to use Units in the future ?

timlind: It's so that we can include a decimal value in the denom in future and use from clp.

proto/sifnode/whitelist/v1/tx.proto
```diff
@@ -0,0 +1,20 @@
...
+option go_package = "github.com/Sifchain/sifnode/x/whitelist/types";
+
+service Msg {
+  rpc UpdateWhitelist(MsgUpdateWhitelist) returns (MsgUpdateWhitelistResponse) {}
+}
+
+message MsgUpdateWhitelist {
+  string from = 1;
+  string denom = 2;
+  int64 units = 3;
```
timlind, line 15: rename to decimals

Jedi2002 (approved):

PR #1611, opened 2021-07-13
IBC wrapper

timlind:


x/ibctransfer/whitelisting.go
```diff
@@ -0,0 +1,95 @@
...
+
+		// The denomination used to send the coins is either the native denom or the hash of the path
+		// if the denomination is not native.
+		denomTrace := transfertypes.ParseDenomTrace(unprefixedDenom)
+		if denomTrace.Path != "" {
+			denom = denomTrace.IBCDenom()
+		}
+
+		return denom
+	}
```
Jedi2002, line 55: Cant we just return denom,bool . 
transfertypes.ReceiverChainIsSource is true "",true
transfertypes.ReceiverChainIsSource is false denom,false

And check for whitelist only in the second case .

Jedi2002: Assuming that we can allow anything that went from sifchain to come back

Jedi2002: Oh are we planning to keep non 18 decimal denoms in the whitelist too ? And use the same whitelist to get their units ?

x/ibctransfer/whitelisting_test.go
```diff
@@ -0,0 +1,52 @@
...
+	"testing"
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	transfertypes "github.com/cosmos/cosmos-sdk/x/ibc/applications/transfer/types"
+	channeltypes "github.com/cosmos/cosmos-sdk/x/ibc/core/04-channel/types"
+	"github.com/stretchr/testify/require"
+	tmproto "github.com/tendermint/tendermint/proto/tendermint/types"
+)
+
+func TestIsRecvPacketAllowed(t *testing.T) {
```
iverc, line 13: I think we could have more tests to cover all whitelisting functions, ideally success/failure case for each one

iverc (changes requested):

Jedi2002 (approved):

x/ibctransfer/whitelisting_test.go
timlind: Yup, will do once we integrate the whitelist keeper.

PR #1616, opened 2021-07-15
fixed run

Jedi2002:
- Fix all integration tests for dispensation module, add more checks
- Add more unit tests for dispensation


timlind (dismissed):
Nice.

I noticed that the Distribution.Runner field isn't being set when creating a distribution. Can you add this value in when creating a distribution?

It existed on the 39 branch and not on 42 so I added when working on the migrate-data code, but it's not being set on creation yet.

marshsif (approved):

iverc:
@timlind Could you please clarify where we do not set distribution runner?
Have you checked the `distpensation_envutils.py`? We have  `create_online_singlekey_txn_with_runner` and `create_offline_singlekey_txn_with_runner` methods that accept `runner_address`.

iverc:
> Nice.
> 
> I noticed that the Distribution.Runner field isn't being set when creating a distribution. Can you add this value in when creating a distribution?
> 
> It existed on the 39 branch and not on 42 so I added when working on the migrate-data code, but it's not being set on creation yet.

Done.

PR #1617, opened 2021-07-15
Crosschain fee rest api

juniuszhou:
add an interface for UI to get the crosschain fee configuration such as token symbol, gas fee for lock/burn


ElliotFriedman:
@juniuszhou Could we add an integration test in here so that we can show the response from the rest server coming back correctly?

juniuszhou:
> @juniuszhou Could we add an integration test in here so that we can show the response from the rest server coming back correctly?

will add it in integration test.

ElliotFriedman (approved):
LGTM!

PR #1621, opened 2021-07-18
Signature aggregation

juniuszhou:


proto/sifnode/ethbridge/v1/tx.proto
```diff
@@ -125,4 +126,14 @@ message MsgSetFeeInfo {
   ];
 }
 
-message MsgSetFeeInfoResponse {}
\ No newline at end of file
+message MsgSetFeeInfoResponse {}
+
+// MsgSignProphecy defines a message for sending signature for prophecy
+message MsgSignProphecy {
```
ElliotFriedman, line 132: @juniuszhou This data structure is incorrect. In order for this to work, the witness must sign off on the following data:

struct ClaimData {
        bytes cosmosSender;
        uint256 cosmosSenderSequence;
        address payable ethereumReceiver;
        address tokenAddress;
        uint256 amount;
        bool doublePeg;
        uint128 nonce;
}

The data should be concatenated using the abi.encode method before being signed. The data should be signed the order specified above. This data that needs to be signed off on can be viewed in the cosmos bridge contract on the future/peggy2 branch starting at line 155 where the struct is defined.

proto/sifnode/oracle/v1/types.proto
```diff
@@ -59,4 +60,12 @@ message CrossChainFeeConfig {
...
     (gogoproto.moretags) = "yaml:\"minimum_burn_cost\""
   ];
+}
+
+// ProphecySignatures include the ethereum addressed and signatures.
+message ProphecySignatures {
+  sifnode.oracle.v1.NetworkDescriptor network_descriptor = 1
+      [ (gogoproto.moretags) = "yaml:\"network_descriptor\"" ];
+  repeated string ethereum_address = 2;
+  repeated string signatures = 3;
```
ElliotFriedman, line 70: Should be named to ProphecySignature and signature, not signatures as each message relayed to sifchain will only contain the signature that witness has signed and not all of the signatures.

x/oracle/keeper/prophecy.go
```diff
@@ -0,0 +1,97 @@
...
+		prophecies = append(prophecies, prophecy)
+	}
+	return prophecies
+}
+
+// GetProphecy gets the entire prophecy data struct for a given id
+func (k Keeper) GetProphecy(ctx sdk.Context, prophecyID []byte) (types.Prophecy, bool) {
+
+	store := ctx.KVStore(k.storeKey)
+	bz := store.Get(append(types.ProphecyPrefix, prophecyID[:]...))
```
ElliotFriedman, line 28: Why do we have to do appending to get the array of prophecies stored?

ElliotFriedman (changes requested):

proto/sifnode/oracle/v1/types.proto
ElliotFriedman: Is signatures being used to define the data structure that all of the signatures for a single prophecy will be stored in?

ElliotFriedman:
What is the API for witnesses to hit to send these new signed prophecies in?

proto/sifnode/ethbridge/v1/tx.proto
juniuszhou: This message is used for relayer to send the signature to sifnode after find out the event of new prophecy. The prophecy id is the hash of claimData in my implementation. so relayer just sign off on the prophecy id.

proto/sifnode/oracle/v1/types.proto
juniuszhou: this data structure is not for message. keeper use it to store all signatures and other information about this prophecy.

x/oracle/keeper/prophecy.go
juniuszhou: it is to get the key of a prophecy, key includes the prophecy prefix and the prophecy id.

proto/sifnode/oracle/v1/types.proto
ElliotFriedman: Ok. Got it.

ElliotFriedman (changes requested):

x/ethbridge/types/msgs.go
```diff
@@ -375,6 +457,56 @@ func NewMsgSetFeeInfo(cosmosSender sdk.AccAddress, networkDescriptor oracletypes
...
+func (msg MsgSignProphecy) Route() string { return RouterKey }
+
+// NewMsgSignProphecy is a constructor function for MsgSignProphecy
+func NewMsgSignProphecy(cosmosSender string, networkDescriptor oracletypes.NetworkDescriptor, prophecyID []byte, ethereumAddress, signature string) MsgSignProphecy {
+	return MsgSignProphecy{
+		CosmosSender:      cosmosSender,
+		NetworkDescriptor: networkDescriptor,
+		ProphecyId:        prophecyID,
+		EthereumAddress:   ethereumAddress,
+		Signature:         signature,
```
ElliotFriedman, line 470: Does geth automatically return signatures as a string

x/ethbridge/keeper/msg_server.go
```diff
@@ -352,3 +386,45 @@ func (srv msgServer) SetFeeInfo(goCtx context.Context, msg *types.MsgSetFeeInfo)
 
 	return &types.MsgSetFeeInfoResponse{}, nil
 }
+
+func (srv msgServer) SignProphecy(goCtx context.Context, msg *types.MsgSignProphecy) (*types.MsgSignProphecyResponse, error) {
```
ElliotFriedman, line 390: This function needs to be permissioned and only allow whitelisted addresses for that network ID to submit prophecies.

ElliotFriedman: So we should take the network ID, get the list of whitelisted relayers for that network ID, check if the sender is in that list, if it is whitelisted, then the tx will succeed, if the sender is not in the whitelist, the tx will fail.

x/oracle/keeper/keeper.go
```diff
@@ -183,6 +160,77 @@ func (k Keeper) SetFeeInfo(ctx sdk.Context, networkDescriptor types.NetworkDescr
...
+				sdk.NewAttribute(types.AttributeKeyCosmosSenderSequence, strconv.FormatInt(int64(prophecyInfo.CosmosSenderSequence), 10)),
+				sdk.NewAttribute(types.AttributeKeyEthereumReceiver, prophecyInfo.EthereumReceiver),
+				// TODO get the token address from denom
+				sdk.NewAttribute(types.AttributeKeyTokenContractAddress, prophecyInfo.EthereumReceiver),
+				sdk.NewAttribute(types.AttributeKeyAmount, strconv.FormatInt(prophecyInfo.TokenAmount.Int64(), 10)),
+				sdk.NewAttribute(types.AttributeKeyDoublePeggy, strconv.FormatBool(prophecyInfo.DoublePeg)),
+				sdk.NewAttribute(types.AttributeKeyGlobalNonce, strconv.FormatInt(int64(prophecyInfo.GlobalNonce), 10)),
+				sdk.NewAttribute(types.AttributeKeycrossChainFee, strconv.FormatInt(prophecyInfo.CrosschainFee.Int64(), 10)),
+				sdk.NewAttribute(types.AttributeKeySignatures, strings.Join(prophecyInfo.Signatures, ",")),
+				sdk.NewAttribute(types.AttributeKeyEthereumAddresses, strings.Join(prophecyInfo.EthereumAddress, ",")),
```
ElliotFriedman, line 227: Why would we emit an event with an array of ethereum addresses? There can only be 1 receiver and 1 token address.

x/oracle/keeper/prophecy.go
```diff
@@ -0,0 +1,135 @@
...
+import (
+	"errors"
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+
+	"github.com/Sifchain/sifnode/x/oracle/types"
+)
+
+// ProphecyLiftTime is used to clean outdated prophecy info from keeper
+const ProphecyLiftTime = 1000
```
ElliotFriedman, line 12: Per https://docs.google.com/document/d/1U420zg7tD7mAr1hlT2y6_BaqMepbnUM_cFfLz7zEfwo/edit#heading=h.k2oo0395vd1o

The ProphecyLiftTime should be 520,000

x/oracle/keeper/prophecy.go
```diff
@@ -0,0 +1,135 @@
...
+	var prophecyInfo types.ProphecyInfo
+	currentHeight := uint64(k.currentHeight)
+
+	store := ctx.KVStore(k.storeKey)
+	iter := sdk.KVStorePrefixIterator(store, types.SignaturePrefix)
+	for ; iter.Valid(); iter.Next() {
+		k.cdc.MustUnmarshalBinaryBare(iter.Value(), &prophecyInfo)
+		if prophecyInfo.BlockNumber-currentHeight > ProphecyLiftTime {
+			storePrefix := append(types.SignaturePrefix, prophecyInfo.ProphecyId[:]...)
+			store.Delete(storePrefix)
```
ElliotFriedman, line 132: Here if we end up deleting, we should update the k.currentHeight to be the current block so that we do not continually scan every block after the first 1000 blocks.

ElliotFriedman (changes requested):

x/oracle/keeper/prophecy.go
juniuszhou: the currentHeight is keeping updated via following method for every new block
func (am AppModule) BeginBlock(ctx sdk.Context, block abci.RequestBeginBlock) {
	am.keeper.UpdateCurrentHeight(block.Header.Height)

x/oracle/keeper/keeper.go
juniuszhou: when a prophecy become completed, the relayer need get all signatures and all addresses, then relayer can call the smart contract with all data needed.

x/ethbridge/keeper/msg_server.go
juniuszhou: yes, should check it. will add it in keeper of oracle.

x/ethbridge/keeper/keeper.go
```diff
@@ -135,34 +140,40 @@ func (k Keeper) ProcessBurn(ctx sdk.Context, cosmosSender sdk.AccAddress, msg *t
...
 	err = k.bankKeeper.BurnCoins(ctx, types.ModuleName, coins)
 	if err != nil {
 		logger.Error("failed to burn locked coin.",
 			errorMessageKey, err.Error())
-		return err
+		return []byte{}, err
 	}
 
-	return nil
+	// TODO global sequence will be implemented in other feature
```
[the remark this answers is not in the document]
juniuszhou: will merge it.

x/ethbridge/keeper/msg_server.go
```diff
@@ -60,12 +82,7 @@ func (srv msgServer) Lock(goCtx context.Context, msg *types.MsgLock) (*types.Msg
 		sdk.NewEvent(
 			types.EventTypeLock,
 			sdk.NewAttribute(types.AttributeKeyNetworkDescriptor, strconv.FormatInt(int64(msg.NetworkDescriptor), 10)),
-			sdk.NewAttribute(types.AttributeKeyCosmosSender, msg.CosmosSender),
```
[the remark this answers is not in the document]
juniuszhou: in the relayer side, my implementation will be like that relayer just get the prophecy id from the event and then sign it, send the singature to sifnode with signProphecy message. relayer doesn't get the raw data and store these data. So when the prophecy completed, sifnode will emit the event with all signatures. relayer not process other signProphecy message, just the one change the prophecy status from pending to completed.

x/ethbridge/keeper/globalNonce.go
```diff
@@ -0,0 +1,49 @@
...
+	globalNonce := binary.LittleEndian.Uint64(value)
+
+	bs := make([]byte, 8)
+	binary.LittleEndian.PutUint64(bs, globalNonce+1)
+
+	return globalNonce
+}
+
+// ExistsGlobalNonce check if the global nonce exists
+func (k Keeper) ExistsGlobalNonce(ctx sdk.Context, prefix []byte) bool {
```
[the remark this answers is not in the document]
juniuszhou: the globalNonce always incremental, so the keeper store the next nonce. all number less than one in keeper are used.I think the data in the keeper will be in the world state, and sync up between different Cosmos nodes. not sure if any other single place we can use to store the global nonce.

x/ethbridge/keeper/globalNonce.go
```diff
@@ -0,0 +1,49 @@
...
+	"encoding/binary"
+
+	"github.com/Sifchain/sifnode/x/ethbridge/types"
+	oracletypes "github.com/Sifchain/sifnode/x/oracle/types"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+)
+
+// GetAndUpdateGlobalNonce get current global nonce and update it
+func (k Keeper) GetAndUpdateGlobalNonce(ctx sdk.Context, networkDescriptor oracletypes.NetworkDescriptor) uint64 {
+	prefix := k.GetGlobalNoncePrefix(ctx, networkDescriptor)
```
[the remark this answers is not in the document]
juniuszhou: yes, but we need use the keeper to store global nonce. keeper need key/value, then the network descriptor become the key by nature.

x/oracle/keeper/keeper.go
```diff
@@ -183,6 +160,87 @@ func (k Keeper) SetFeeInfo(ctx sdk.Context, networkDescriptor types.NetworkDescr
...
+func (k Keeper) ProcessSignProphecy(ctx sdk.Context, networkDescriptor types.NetworkDescriptor, prophecyID []byte, cosmosSender, ethereumAddress, signature string) error {
+	prophecy, ok := k.GetProphecy(ctx, prophecyID)
+	if !ok {
+		return types.ErrProphecyNotFound
+	}
+
+	whiteList := k.GetOracleWhiteList(ctx, types.NewNetworkIdentity(networkDescriptor))
+	power, ok := whiteList.WhiteList[cosmosSender]
+	if !ok {
+		return errors.New("message sender to sign prophecy not in the whitelist")
```
ElliotFriedman, line 183: @banshee Flagging this as a path we need to test on peggy2.0.

x/oracle/keeper/keeper.go
```diff
@@ -183,6 +160,87 @@ func (k Keeper) SetFeeInfo(ctx sdk.Context, networkDescriptor types.NetworkDescr
...
+
+	// verify the signature
+	publicKey, err := gethCrypto.Ecrecover(prophecyID, gethCommon.FromHex(signature))
+	if err != nil {
+		return err
+	}
+
+	ok = gethCrypto.VerifySignature(publicKey, prophecyID, []byte(signature))
+	if !ok {
+		return errors.New("incorrect signature")
```
ElliotFriedman, line 198: Let's change this error message to `incorrect ethereum signature` so that it is a more descriptive error.

x/oracle/keeper/keeper.go
```diff
@@ -183,6 +160,87 @@ func (k Keeper) SetFeeInfo(ctx sdk.Context, networkDescriptor types.NetworkDescr
...
+				sdk.NewAttribute(types.AttributeKeyAmount, strconv.FormatInt(prophecyInfo.TokenAmount.Int64(), 10)),
+				sdk.NewAttribute(types.AttributeKeyDoublePeggy, strconv.FormatBool(prophecyInfo.DoublePeg)),
+				sdk.NewAttribute(types.AttributeKeyGlobalNonce, strconv.FormatInt(int64(prophecyInfo.GlobalNonce), 10)),
+				sdk.NewAttribute(types.AttributeKeycrossChainFee, strconv.FormatInt(prophecyInfo.CrosschainFee.Int64(), 10)),
+				sdk.NewAttribute(types.AttributeKeySignatures, strings.Join(prophecyInfo.Signatures, ",")),
+				sdk.NewAttribute(types.AttributeKeyEthereumAddresses, strings.Join(prophecyInfo.EthereumAddress, ",")),
+			),
+		})
+	}
+	return err
```
ElliotFriedman, line 241: We should not return an error here if the call to `AppendValidatorToProphecy` failed due to the prophecy already being complete.

We need the fourth and final tx to sifnode to complete so that that relayers nonce will increment as we map networkdescriptor => relayer address => nonce.

x/oracle/keeper/prophecy.go
```diff
@@ -0,0 +1,135 @@
...
+
+	store.Set(storePrefix, k.cdc.MustMarshalBinaryBare(&prophecy))
+}
+
+// GetProphecyInfo return a prophecy's signatures
+func (k Keeper) GetProphecyInfo(ctx sdk.Context, prophecyID []byte) (types.ProphecyInfo, bool) {
+	var prophecySignatures types.ProphecyInfo
+	store := ctx.KVStore(k.storeKey)
+
+	// storePrefix := append(types.SignaturePrefix, prophecyID[:]...)
```
ElliotFriedman, line 57: Remove commented out code.

ElliotFriedman (changes requested):

x/oracle/keeper/keeper.go
juniuszhou: It returns an ErrProphecyFinalized, then in msg_server.go, the error will be ignored. SignProphecy in msg_server.go will continue to emit the EventTypeSignProphecy.

x/oracle/keeper/prophecy.go
juniuszhou: will remove it.

x/oracle/keeper/keeper.go
juniuszhou: accepted

cmd/ebrelayer/relayer/prophecyHandler.go
```diff
@@ -0,0 +1,82 @@
+package relayer
+
+import (
+	cosmosbridge "github.com/Sifchain/sifnode/cmd/ebrelayer/contract/generated/bindings/cosmosbridge"
+	"github.com/Sifchain/sifnode/cmd/ebrelayer/txs"
+	"github.com/Sifchain/sifnode/cmd/ebrelayer/types"
+)
+
+var currentGlobalNonce = uint64(0)
```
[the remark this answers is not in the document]
juniuszhou: We can get it but what's usage.
1. if the value got is less than global nonce in the new message, we still need handle the message. I can't see any difference with initiate it with zero.
2. if the value got is larger, we can avoid send the transaction to etheruem network again. but it should never happen in realistic environment.

cmd/ebrelayer/replay.go
```diff
@@ -104,22 +110,95 @@ func RunReplayCosmosCmd(_ *cobra.Command, args []string) error {
...
+	}
+
+	sugaredLogger := logger.Sugar()
+
+	// Initialize new Cosmos event listener
+	cosmosSub := relayer.NewCosmosSub(oracletypes.NetworkDescriptor(networkDescriptor),
+		privateKey, tendermintNode, web3Provider, contractAddress, nil, cliContext,
+		validatorMoniker, true, sugaredLogger)
+
+	cosmosSub.ReplayBurnLock(txFactory, fromBlock, toBlock)
```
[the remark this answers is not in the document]
juniuszhou: Agree, it is better to ask the caller to set block numbers. Will implement it.

cmd/ebrelayer/relayer/cosmos.go
```diff
@@ -416,48 +286,17 @@ func (sub CosmosSub) handleBurnLockMsg(
...
 
-	for i < maxRetries {
-		err = txs.RelayProphecyClaimToEthereum(
-			cosmosMsg,
-			sub.SugaredLogger,
-			client,
-			auth,
-			cosmosBridgeInstance,
-		)
+	txs.SignProphecyToCosmos(txFactory, signProphecy, sub.CliContext, sub.SugaredLogger)
```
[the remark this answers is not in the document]
juniuszhou: The new solution is to send transaction to cosmos, we do it before because the old solution need send transaction to Ethereum, we often have some traffic issue. and we didn't have retry for Sifchain transaction. I think it is safe to remove retry since cosmos network much more efficient.

x/ethbridge/keeper/msg_server.go
juniuszhou: the check added in keeper of oracle module.

ElliotFriedman (changes requested):

x/ethbridge/types/msgs.go
```diff
@@ -152,6 +167,71 @@ func (msg MsgBurn) GetSigners() []sdk.AccAddress {
...
+		// TODO need get the token address from token's symbol
+		msg.EthereumReceiver,
+		msg.Amount,
+		doublePeggy,
+		globalNonce,
+	)
+}
+
+// ComputeProphecyID compute the prophecy id
+func ComputeProphecyID(cosmosSender string, sequence uint64, ethereumReceiver string, tokenAddress string, amount sdk.Int,
```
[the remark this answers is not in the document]
juniuszhou: got it, will update the algorithm in sifnode.

cmd/ebrelayer/relayer/cosmosReplay.go
```diff
@@ -276,7 +276,43 @@ func (sub CosmosSub) getAllSignSigature(accAddress sdk.AccAddress, fromBlock int
...
+	}
+
+	// Initialize CosmosBridge instance
+	cosmosBridgeInstance, err := cosmosbridge.NewCosmosBridge(target, client)
+	if err != nil {
+		sub.SugaredLogger.Errorw("failed to get cosmosBridge instance.",
+			errorMessageKey, err.Error())
+		return nil, err
+	}
+	return cosmosBridgeInstance.LastNonceSubmitted(nil)
```
ElliotFriedman, line 298: The variable has a lowercase L to start. Have you verified this returns the proper number from the smart contract?

ElliotFriedman (changes requested):

cmd/ebrelayer/relayer/cosmosReplay.go
juniuszhou: but I can't access to the varialbe directly, must use the function with capital. didn't do any integration test yet.

cmd/ebrelayer/main.go
```diff
@@ -92,21 +92,21 @@ func buildRootCmd() *cobra.Command {
...
 //	initRelayerCmd
 func initRelayerCmd() *cobra.Command {
 	//nolint:lll
 	initRelayerCmd := &cobra.Command{
-		Use:     "init [networkDescriptor] [tendermintNode] [web3Provider] [bridgeRegistryContractAddress] [validatorMnemonic]",
+		Use:     "init [networkDescriptor] [tendermintNode] [web3Provider] [bridgeRegistryContractAddress] [validatorMnemonic] [signatureAggregator]",
 		Short:   "Validate credentials and initialize subscriptions to both chains",
-		Args:    cobra.ExactArgs(5),
-		Example: "ebrelayer init 1 tcp://localhost:26657 ws://localhost:7545/ 0x30753E4A8aad7F8597332E813735Def5dD395028 mnemonic --chain-id=peggy",
+		Args:    cobra.ExactArgs(6),
```
Brando753, line 108: Is there any reason to leave signatureAggregation as a required argument over a flag? Shouldn't the system default to 'true' unless a flag says otherwise?

juniuszhou: We just have one or two dedicated relayers to hangle the signature aggregation events. For not signature aggregation relayers, they just handle the burn/lock messages from sifnode, then they sign against the prophecy id, send the message back to Sifchain. After enought relayers sign for the same prophecy, the signature aggregation completed event emitted. then signature aggregation relayers get all signatures and other info, send a transaction to smart contract in Ethereum.

x/ethbridge/keeper/keeper.go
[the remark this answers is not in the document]
ElliotFriedman: That PR has been merged, can we rebase with the main future/peggy2 branch?

x/oracle/keeper/keeper.go
ElliotFriedman: The ethereum address array is the array of the signers?

cmd/ebrelayer/txs/relayToCosmos.go
```diff
@@ -71,3 +70,36 @@ func RelayToCosmos(factory tx.Factory, claims []*types.EthBridgeClaim, cliCtx cl
...
+func SignProphecyToCosmos(factory tx.Factory, signProphecy ethbridge.MsgSignProphecy, cliCtx client.Context, sugaredLogger *zap.SugaredLogger) {
+	var messages []sdk.Msg
+
+	messages = append(messages, &signProphecy)
+
+	sugaredLogger.Infow("RelayToCosmos building, signing, and broadcasting", "messages", messages)
+	// TODO this WithGas isn't correct
+	// TODO we need to investigate retries
+	// TODO we need to investigate what happens when the transaction has already been completed
+	err := tx.BroadcastTx(
```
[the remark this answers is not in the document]
ElliotFriedman: @banshee bump

cmd/ebrelayer/txs/relayToCosmos.go
```diff
@@ -71,3 +70,36 @@ func RelayToCosmos(factory tx.Factory, claims []*types.EthBridgeClaim, cliCtx cl
...
+		cliCtx,
+		factory.
```
ElliotFriedman, line 86: We'll just accept the changes of how this is done on the current develop branch as they have the necessary changes for 0.42.

cmd/ebrelayer/txs/relayToEthereum.go
```diff
@@ -13,93 +13,25 @@ import (
...
+	ethereumtypes "github.com/ethereum/go-ethereum/core/types"
 	"github.com/ethereum/go-ethereum/ethclient"
+
 	"go.uber.org/zap"
 )
 
 const (
 	// GasLimit the gas limit in Gwei used for transactions sent with TransactOpts
-	GasLimit            = uint64(500000)
+	GasLimit = uint64(500000)
```
ElliotFriedman, line 24: We need a higher gas limit here to allow for creation of the double pegged assets on destination chains. @daniportugalgit I would appreciate if you could post here how much it costs to create a new bridgetoken.

I think setting the gas limit at 2 million should be good and allow us to handle cases where bridgetokens are created.

cmd/ebrelayer/txs/relayToEthereum.go
```diff
@@ -185,4 +107,97 @@ func InitRelayConfig(
...
+	)
+
+	claimData := cosmosbridge.CosmosBridgeClaimData{
+		CosmosSender:         []byte(prophecyInfo.CosmosSender),
+		CosmosSenderSequence: big.NewInt(int64(prophecyInfo.CosmosSenderSequence)),
+		EthereumReceiver:     common.HexToAddress(prophecyInfo.EthereumReceiver),
+		TokenAddress:         common.HexToAddress(prophecyInfo.TokenSymbol),
+		Amount:               &prophecyInfo.TokenAmount,
+		DoublePeg:            prophecyInfo.DoublePeg,
+		Nonce:                big.NewInt(int64(prophecyInfo.GlobalNonce)),
```
ElliotFriedman, line 136: You'll also need to add the network descriptor here.

@daniportugalgit please take a look over this claim data and check if anything is missing.

cmd/ebrelayer/types/types.go
```diff
@@ -185,6 +192,30 @@ type EthereumBridgeClaim struct {
...
+	CosmosSender      sdk.ValAddress
+	NetworkDescriptor oracle.NetworkDescriptor
+	ProphecyID        []byte
+}
+
+// ProphecyInfo store all data needed for smart contract call
+type ProphecyInfo struct {
+	TokenAmount          big.Int
+	ProphecyID           []byte
+	EthereumAddresses    []string
```
ElliotFriedman, line 208: This should be renamed to `EthereumTokenAddress` or `TokenAddress`.

x/ethbridge/client/rest/rest.go
```diff
@@ -53,6 +53,15 @@ type burnOrLockEthReq struct {
 	CrosschainFee     sdk.Int      `json:"cross_chain_fee_amount" yaml:"cross_chain_fee_amount"`
 }
 
+type signProphecyReq struct {
+	BaseReq           rest.BaseReq `json:"base_req"`
+	NetworkDescriptor string       `json:"network_descriptor"`
+	CosmosSender      string       `json:"cosmos_sender"`
+	EthereumAddress   string       `json:"ethereum_receiver"`
```
ElliotFriedman, line 60: If EthereumAddress is the ethereum receiver address let's rename to `EthereumReceiverAddress`, if EthereumAddress is the token address, let's rename to `EthereumTokenAddress`

`EthereumAddress` on its own is not descriptive enough of what it is.

Also, the signProphecyRequest should also have the nonce on it.

x/ethbridge/keeper/msg_server.go
ElliotFriedman: Gotcha, the `ProcessSignProphecy` function checks that the sender is in fact whitelisted.

x/ethbridge/keeper/msg_server.go
```diff
@@ -356,3 +386,47 @@ func (srv msgServer) SetFeeInfo(goCtx context.Context, msg *types.MsgSetFeeInfo)
...
+		sdk.NewEvent(
+			sdk.EventTypeMessage,
+			sdk.NewAttribute(sdk.AttributeKeyModule, types.AttributeValueCategory),
+			sdk.NewAttribute(sdk.AttributeKeySender, msg.CosmosSender),
+		),
+		sdk.NewEvent(
+			types.EventTypeSignProphecy,
+			sdk.NewAttribute(types.AttributeKeyCosmosSender, msg.CosmosSender),
+			sdk.NewAttribute(types.AttributeKeyNetworkDescriptor, msg.NetworkDescriptor.String()),
+			sdk.NewAttribute(types.AttributeKeyProphecyID, string(msg.ProphecyId)),
```
ElliotFriedman, line 427: We should add an attribute that is the cosmos nonce of that message that just got signed.

x/ethbridge/types/msgs.go
```diff
@@ -152,6 +168,76 @@ func (msg MsgBurn) GetSigners() []sdk.AccAddress {
...
+	doublePeggy bool, globalNonce uint64, networkDescriptor oracletypes.NetworkDescriptor) []byte {
+
+	bytesTy, _ := abi.NewType("bytes", nil)
+	boolTy, _ := abi.NewType("bool", nil)
+	uint128Ty, _ := abi.NewType("uint128", nil)
+	uint256Ty, _ := abi.NewType("uint256", nil)
+	addressTy, _ := abi.NewType("address", nil)
+
+	arguments := abi.Arguments{
```
ElliotFriedman, line 197: @daniportugalgit Please take a look at this and double check that everything is in order per the new smart contract data structures.

x/oracle/keeper/keeper.go
```diff
@@ -183,6 +160,91 @@ func (k Keeper) SetFeeInfo(ctx sdk.Context, networkDescriptor types.NetworkDescr
...
+		ctx.EventManager().EmitEvents(sdk.Events{
+
+			sdk.NewEvent(
+				types.EventTypeProphecyCompleted,
+				sdk.NewAttribute(types.AttributeKeyProphecyID, string(prophecyID)),
+				sdk.NewAttribute(types.AttributeKeyNetworkDescriptor, prophecyInfo.NetworkDescriptor.String()),
+				sdk.NewAttribute(types.AttributeKeyCosmosSender, prophecyInfo.CosmosSender),
+				sdk.NewAttribute(types.AttributeKeyCosmosSenderSequence, strconv.FormatInt(int64(prophecyInfo.CosmosSenderSequence), 10)),
+				sdk.NewAttribute(types.AttributeKeyEthereumReceiver, prophecyInfo.EthereumReceiver),
+				// TODO get the token address from denom
```
ElliotFriedman, line 234: @Brando753 Once this PR is merged, please assist with pulling in metadata from the metadata module.

ElliotFriedman (changes requested):

x/ethbridge/keeper/msg_server.go
juniuszhou: will add this attribute

cmd/ebrelayer/txs/parser.go
```diff
@@ -98,53 +93,15 @@ func BurnLockEventToCosmosMsg(claimType types.Event, attributes []abci.EventAttr
 		key := string(attribute.GetKey())
 		val := string(attribute.GetValue())
 
+		fmt.Printf(" key is %v, value is %v\n", key, val)
+
 		// Set variable based on the attribute's key
 		switch key {
-		case types.CosmosSender.String():
```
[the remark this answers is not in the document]
juniuszhou: the relayer doesn't need all information like CosmosSender as before, becuase relayer not send the transaction to smart contract for BurnOrLock happened in Cosmos side. relayer just need know the prophecy id and sign it, then send message with signature back to Cosmos.

cmd/ebrelayer/main.go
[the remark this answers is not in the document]
juniuszhou: will split into two commands.

cmd/ebrelayer/txs/relayToEthereum.go
daniportugalgit: Apart from the `networkDescriptor`, the new claim data structure has 3 other new fields: `tokenName`, `tokenSymbol` and `tokenDecimals`. Here's the typed struct we will have on our smart contracts once my PR gets merged into peggy2:
```
struct ClaimData {
  bytes cosmosSender;
  uint256 cosmosSenderSequence;
  address payable ethereumReceiver;
  address tokenAddress;
  uint256 amount;
  bool doublePeg;
  uint128 nonce;
  uint256 networkDescriptor;
  string tokenName;
  string tokenSymbol;
  uint8 tokenDecimals;
}
```

@ElliotFriedman By the way, I _think_ we might save some gas by reordering those properties inside the struct (because of tight variable packing). It's just optimization, though, and we could leave it to some other time.

x/ethbridge/types/msgs.go
daniportugalgit: Although the ClaimData does have new fields, the ProphecyID doesn't.
This structure is correct.

x/ethbridge/client/rest/rest.go
juniuszhou: it is signer's address, will rename as EthereumSignerAddress, also add nonce in the request.

cmd/ebrelayer/types/types.go
juniuszhou: it is signer's ethereum address, rename to EthereumSignerAddresses

cmd/ebrelayer/txs/relayToEthereum.go
juniuszhou: sure, after the new smart contract merged. will generate new go source file and add these items.

cmd/ebrelayer/relayer/prophecyHandler.go
```diff
@@ -0,0 +1,82 @@
...
+	if currentGlobalNonce > prophecyInfo.GlobalNonce {
+		sub.SugaredLogger.Errorw("global nonce is invalid.",
+			"current global nonce", currentGlobalNonce,
+			"global nonce in message", prophecyInfo.GlobalNonce)
+		return
+	}
+
+	// buffer the prophecy
+	if prophecyInfo.GlobalNonce > currentGlobalNonce+1 {
+		prophecyInfoBuffer[prophecyInfo.GlobalNonce] = prophecyInfo
```
ElliotFriedman, line 31: Is this `prophecyInfoBuffer` ever used? I only see two references to it, once on creation and second here where things are added to it. I never see the buffered items read.

cmd/ebrelayer/txs/relayToEthereum.go
daniportugalgit: To create a new BridgeToken, we're currently using:
With 3 validators: 1,149,480 gas.
With 4 validators: 1,162,781 gas. (+13,301)
With 5 validators: 1,176,677 gas. (+13,896)
With 6 validators: 1,191,190 gas. (+14,513)
With 7 validators: 1,206,296 gas. (+15,106)
With 8 validators: 1,222,006 gas. (+15,710)

**Roughly**, starting from 3 validators, for each new validator we add, we'll spend an extra
`13,300 * numberOfNewValidators + 605 * (numberOfNewValidators - 1)` gas.

I agree that 2 million gas should be good. It would allow us to have up to 61 validators.

cmd/ebrelayer/relayer/prophecyHandler.go
juniuszhou: will remove it

cmd/ebrelayer/relayer/prophecyHandler.go
```diff
@@ -0,0 +1,264 @@
...
+		endBlock = currentBlock - ProphecyLifeTime
+
+	} else {
+		endBlock = 0
+	}
+
+	ctx := context.Background()
+	// if found the next one
+	found := false
+	for currentBlock > endBlock {
```
ElliotFriedman, line 215: `currentBlock` needs to be decremented at the end of this loop so that this is not an infinite loop.

ElliotFriedman (changes requested):

cmd/ebrelayer/relayer/prophecyHandler.go
```diff
@@ -0,0 +1,264 @@
...
+		// not go to priviouos block if found.
+		if found {
+			break
+		}
+		tmpBlock := int64(currentBlock)
+		block, err := client.BlockResults(ctx, &tmpBlock)
```
ElliotFriedman, line 221: Let's update this to use the new API to gather a whole bunch of data at once from sifnode instead of querying block by block.

ElliotFriedman (changes requested):

cmd/ebrelayer/relayer/prophecyHandler.go
juniuszhou: implemented.

cmd/ebrelayer/relayer/prophecyHandler.go
juniuszhou: loop on block number will be removed

cmd/ebrelayer/relayer/prophecyHandler.go
```diff
@@ -0,0 +1,192 @@
...
+	lastSubmittedNonce, err := GetLastNonceSubmitted(ethClient, cosmosBridgeAddress, sub.SugaredLogger)
+	if err != nil {
+		sub.SugaredLogger.Errorw("failed to get the last submitted nonce.",
+			errorMessageKey, err.Error())
+		return
+	}
+
+	prophecyInfoArray := GetAllProphciesCompleted(sub.TmProvider, sub.NetworkDescriptor, lastSubmittedNonce.Uint64()+1)
+
+	batches := (len(prophecyInfoArray) + 4) / 5
```
ElliotFriedman, line 93: I'm not sure I understand the formula we are using to calculate how many batches we have.

juniuszhou: yes, it calculate the batchs. if we length is 2 or 3, we still need process one batch.

x/ethbridge/client/rest/rest.go
```diff
@@ -55,6 +55,15 @@ type burnOrLockEthReq struct {
 	CrosschainFee     sdk.Int      `json:"cross_chain_fee_amount" yaml:"cross_chain_fee_amount"`
 }
 
+type signProphecyReq struct {
+	BaseReq               rest.BaseReq `json:"base_req"`
+	NetworkDescriptor     string       `json:"network_descriptor"`
+	CosmosSender          string       `json:"cosmos_sender"`
+	EthereumSignerAddress string       `json:"ethereum_signer_address"`
+	Signature             string       `json:"signature"`
+	ProphecyID            string       `json:"prophecy_id"`
```
ElliotFriedman, line 64: With the new data structure on the cosmos side, we are going to have to add more to this struct. We can do this after this PR is merged.

x/ethbridge/keeper/keeper.go
```diff
@@ -57,18 +62,19 @@ func (k Keeper) ProcessClaim(ctx sdk.Context, claim *types.EthBridgeClaim) (orac
...
-		symbol := fmt.Sprintf("%v%v", types.PeggedCoinPrefix, claim.Symbol)
-		k.AddPeggyToken(ctx, symbol)
-
-		coins = sdk.Coins{sdk.NewCoin(symbol, claim.Amount)}
+		coins = sdk.Coins{sdk.NewCoin(tokenMetadata.Symbol, claim.Amount)}
 		err = k.bankKeeper.MintCoins(ctx, types.ModuleName, coins)
 	case types.ClaimType_CLAIM_TYPE_BURN:
-		coins = sdk.Coins{sdk.NewCoin(claim.Symbol, claim.Amount)}
+		coins = sdk.Coins{sdk.NewCoin(tokenMetadata.Symbol, claim.Amount)}
 		err = k.bankKeeper.MintCoins(ctx, types.ModuleName, coins)
```
ElliotFriedman, line 78: Here we should transfer coins from the ethbridge module to the user instead of minting them coins.

juniuszhou: Agree, no need to mint for the burn case.

x/ethbridge/client/rest/rest.go
juniuszhou: Ok, will sync up the change after smart contract finalied.

x/ethbridge/keeper/keeper.go
```diff
@@ -75,7 +75,7 @@ func (k Keeper) ProcessSuccessfulClaim(ctx sdk.Context, claim *types.EthBridgeCl
...
 	case types.ClaimType_CLAIM_TYPE_BURN:
 		coins = sdk.Coins{sdk.NewCoin(tokenMetadata.Symbol, claim.Amount)}
-		err = k.bankKeeper.MintCoins(ctx, types.ModuleName, coins)
+		err = nil
```
ElliotFriedman, line 78: We can't just do nothing. We have to transfer from module to the account.

juniuszhou: code line 95	if err := k.bankKeeper.SendCoinsFromModuleToAccount(
will transfer the coin.

ElliotFriedman: Wouldn't this double spend coins in the CLAIM_TYPE_MINT because the user gets minted coins on line 75, then on line 95, we then transfer coins from the module to this user.

juniuszhou: mint just for module account, which get the minted token, then transfer to reciver account.

ElliotFriedman (approved):
LFG!

PR #1622, opened 2021-07-19
[ChainOps] Latest merge to master.

intl-man:


timlind (approved):

x/oracle/keeper/keeper.go
```diff
@@ -43,6 +44,27 @@ func (k Keeper) Logger(ctx sdk.Context) log.Logger {
...
 }
 
+func (k Keeper) GetProphecies(ctx sdk.Context) []types.Prophecy {
+	var prophecies []types.Prophecy
+	store := ctx.KVStore(k.storeKey)
+	iter := store.Iterator(nil, nil)
+	for ; iter.Valid(); iter.Next() {
+		key := iter.Key()
+		var dbProphecy types.DBProphecy
+		if bytes.Compare(key, types.AdminAccountPrefix) != 0 &&
```
juniuszhou, line 54: we can use the KVStorePrefixIterator to get all prophecies. current implemenation only works if there are just three prefix for oracle module.

timlind: @juniuszhou the prophecies haven't been set with any prefix.

intl-man: @juniuszhou - Good point. Yes, we can. Do you have an exact implementation in mind?

juniuszhou (approved):

PR #1626, opened 2021-07-19
Use Compound's system of renaming smart contracts to ...G(n + 1) when they have changes

banshee:
Restore smart contracts to their deployed state, and move the changed ones to names ending in ...G2.sol


ElliotFriedman:
Because of dependency issues and switching our smart contract development framework, we will not be going forward with this approach.

PR #1634, opened 2021-07-21
clp normalization using whitelist

Jedi2002:
The add-denoms-toWhitelist , gentx would be required to for the integration tests to pass.


x/clp/keeper/table_test.go
```diff
@@ -4,16 +4,51 @@ import (
...
+	m["cdai"] = 8
+	m["wbtc"] = 8
+	m["ceth"] = 8
+	m["renbtc"] = 8
+	m["cusdc"] = 8
+	m["husd"] = 8
+	m["ampl"] = 9
+	return m
+}
+func GetNormalizationFactorStub(denom string) (sdk.Dec, bool) {
```
timlind, line 33: We should test the actual implementation of this that exists in the clp keeper.

The test should insert register the denoms in the whitelist keeper so it can be fetched from there.

Can use a mock generator for whitelist keeper if needed: https://github.com/golang/mock

x/whitelist/keeper/keeper.go
```diff
@@ -115,3 +115,28 @@ func (k keeper) GetDenomWhitelist(ctx sdk.Context) types.DenomWhitelist {
 
 	return whitelist
 }
+
+func (k keeper) GetNormalizationFactor(ctx sdk.Context, denom string) (sdk.Dec, bool) {
```
timlind, line 119: GetNormalizationFactor is a CLP function so should be implemented there instead of on whitelist keeper

x/whitelist/keeper/keeper.go
```diff
@@ -115,3 +115,28 @@ func (k keeper) GetDenomWhitelist(ctx sdk.Context) types.DenomWhitelist {
...
+	wl := k.GetDenomWhitelist(ctx)
+	normalizationFactor := sdk.NewDec(1)
+	ok := false
+	nf := int64(0)
+	for i := range wl.DenomWhitelistEntries {
```
timlind, line 124: Why not use GetDenom()?

x/whitelist/utils/parser_test.go
```diff
@@ -0,0 +1,32 @@
+package utils
+
+import (
+	"encoding/json"
+	types2 "github.com/Sifchain/sifnode/x/whitelist/types"
```
timlind, line 5: remove alias.

x/clp/handler_test.go
```diff
@@ -409,14 +419,16 @@ func CalculateWithdraw(t *testing.T, keeper clpkeeper.Keeper, ctx sdk.Context, a
...
-	emitAmount, _, _, _, err := clpkeeper.SwapOne(assetSent, swapAmount, clptypes.GetSettlementAsset(), inPool)
+	normalizationFactor, adjustExternalToken := wk.GetNormalizationFactor(ctx, inPool.ExternalAsset.Symbol)
+	emitAmount, _, _, _, err := clpkeeper.SwapOne(assetSent, swapAmount, clptypes.GetSettlementAsset(), inPool, normalizationFactor, adjustExternalToken)
 	assert.NoError(t, err)
-	emitAmount2, _, _, _, err := clpkeeper.SwapOne(clptypes.GetSettlementAsset(), emitAmount, assetReceived, outPool)
+	normalizationFactor, adjustExternalToken = wk.GetNormalizationFactor(ctx, outPool.ExternalAsset.Symbol)
+	emitAmount2, _, _, _, err := clpkeeper.SwapOne(clptypes.GetSettlementAsset(), emitAmount, assetReceived, outPool, normalizationFactor, adjustExternalToken)
 	assert.NoError(t, err)
 	return emitAmount2
 }
```
timlind, line 434: Add test table for erroring on disabled tokens for all handlers.

timlind (changes requested):

denoms.json
```diff
@@ -0,0 +1,19 @@
+{
```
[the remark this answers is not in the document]
Jedi2002: Yes , I added this as a sample, should ideally be generated and removed my the integration tests

x/clp/keeper/msg_server.go
```diff
@@ -14,19 +15,19 @@ import (
 
 type msgServer struct {
 	Keeper
+	whitelistKeeper whitelisttypes.Keeper
```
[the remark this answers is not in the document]
Jedi2002: not sure hwat the second part means, I made some changes please check if thats what you ment as well .

x/clp/keeper/table_test.go
Jedi2002: That can be a different test , this test is pure. It just tests the mathematical  formulae, this file does not have anything related to the blockchain / state . 
We have a  counterpart for these on the frontend as well.

x/whitelist/keeper/keeper.go
Jedi2002: Moved it , but the I was under the impression GetNormalizationFactor might be moved to the peggy bridge and used in ibc_sifchain .
 CLP in general is a mathematical module , it should not be dealing with normlization etc .

timlind (approved):
LGTM

PR #1636, opened 2021-07-22
Remove ebrelayer changes from feature/ibc

banshee:
Leaving it as draft for now since it's not clear if we need these out of feature/ibc or feature/cosmos-0.42


banshee:
Doing this as a PR against feature/ibc since we don't want to disrupt the 42 release.

ElliotFriedman (approved):
LGTM!

PR #1637, opened 2021-07-22
Test and fix to avoid invalid coins errors

iverc:


x/dispensation/keeper/distribution.go
```diff
@@ -20,10 +20,10 @@ func (k Keeper) SetDistribution(ctx sdk.Context, ar types.Distribution) error {
 	return nil
 }
 
-func (k Keeper) GetDistribution(ctx sdk.Context, name string, distributionType types.DistributionType, authorisedRunner string) (*types.Distribution, error) {
```
banshee, line 23: I wouldn't bother to change this (but I wouldn't change the PR either, it's fine either way); both spellings are correct.  (we use the z here in the US and Canada, the rest of the world uses an s)

timlind (approved):
Good find.

PR #1639, opened 2021-07-23
global nonce of lock/burn for each network

juniuszhou:


x/ethbridge/keeper/globalNonce_test.go
```diff
@@ -0,0 +1,25 @@
...
+)
+
+func TestGetAndUpdateGlobalNonce(t *testing.T) {
+	var ctx, keeper, _, _, _, _, _, _ = test.CreateTestKeepers(t, 0.7, []int64{3, 3}, "")
+	networkDescriptor := oracletypes.NetworkDescriptor_NETWORK_DESCRIPTOR_ETHEREUM
+
+	// test the init value
+	zeroGlobalNonce := uint64(0)
+	globalNonce := keeper.GetAndUpdateGlobalNonce(ctx, networkDescriptor)
+	assert.Equal(t, globalNonce, zeroGlobalNonce)
```
ElliotFriedman, line 19: First global nonce should be 1, not 0.

ElliotFriedman (changes requested):

x/ethbridge/keeper/globalNonce.go
```diff
@@ -0,0 +1,48 @@
...
+	prefix := k.GetGlobalNoncePrefix(ctx, networkDescriptor)
+	store := ctx.KVStore(k.storeKey)
+
+	if !k.ExistsGlobalNonce(ctx, prefix) {
+		nextGlobalNonce := uint64(1)
+		bs := make([]byte, 8)
+		binary.LittleEndian.PutUint64(bs, nextGlobalNonce)
+
+		store.Set(prefix, bs)
+		return uint64(0)
```
ElliotFriedman, line 22: This should return 1, not 0.

x/ethbridge/keeper/globalNonce_test.go
juniuszhou: will update it.

ElliotFriedman (approved):

PR #1643, opened 2021-07-25
Token whitelist cli

timlind:


Jedi2002:
http://github.com/Sifchain/sifnode/blob/3f38f4ccfa4db60c1200953f7f2beb36b5f2be63/x/whitelist/keeper/msg_server.go#L18-L21 
might is not required, can still keep for additional safety

PR #1670, opened 2021-07-28
Reduce Fee for Create Dispensation

Jedi2002:


app/ante/ante.go
```diff
@@ -0,0 +1,85 @@
...
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
+	"github.com/cosmos/cosmos-sdk/x/auth/ante"
+	authkeeper "github.com/cosmos/cosmos-sdk/x/auth/keeper"
+	"github.com/cosmos/cosmos-sdk/x/auth/signing"
+	bankkeeper "github.com/cosmos/cosmos-sdk/x/bank/keeper"
+	"github.com/pkg/errors"
+)
+
+func NewAnteHandler(ak authkeeper.AccountKeeper, bk bankkeeper.Keeper, gasConsumer ante.SignatureVerificationGasConsumer, signModeHandler signing.SignModeHandler) sdk.AnteHandler {
```
alexanderbez, line 14: This isn't how I would do it. You're creating an anonymous ante-handler function based on the current tx. Instead you should create a custom ante-handler decorator. This  decorator would do two things:

1. Error based on lines 21-23
2. Would have custom gas/fee logic based on the msgs.

timlind (approved):

PR #1672, opened 2021-07-29
Added Pagination for lplist and getpools

Jedi2002:


x/clp/keeper/grpc_query.go
```diff
@@ -105,21 +107,25 @@ func (k Querier) GetLiquidityProviderList(c context.Context, req *types.Liquidit
...
+	lpList, pageRes, err := k.GetLiquidityProvidersForAssetPaginated(ctx, searchingAsset, req.Pagination)
+	if err != nil {
+		return nil, err
+	}
+
+	liquidityProviders := make([]*types.LiquidityProvider, len(lpList))
 
-	lpl := make([]*types.LiquidityProvider, len(lpList))
 	for i, lp := range lpList {
 		lp := lp
```
timlind, line 121: Do we want to make a copy here? All we want is to convert to pointers right?

x/clp/keeper/liquidityprovider.go
```diff
@@ -86,6 +90,31 @@ func (k Keeper) GetLiquidityProvidersForAsset(ctx sdk.Context, asset types.Asset
 	return lpList
 }
 
+func (k Keeper) GetLiquidityProvidersForAssetPaginated(ctx sdk.Context, asset types.Asset, pagination *query.PageRequest) ([]types.LiquidityProvider, *query.PageResponse, error) {
```
timlind, line 93: Could return pointers here to avoid that conversion in caller.

timlind (commented):
LGTM. Can be deployed to tesnet-042 for debugging. If possible deploy the branch without merging in case we revert.

x/clp/keeper/grpc_query.go
iverc: addressed it here #1673

marshsif (approved):

PR #1673, opened 2021-07-29
Fix lplist issues

iverc:
- DO NOT MERGE

### DIFF
--- a/x/clp/keeper/liquidityprovider.go
+++ b/x/clp/keeper/liquidityprovider.go
@@ -1,6 +1,8 @@
 package keeper
 
 import (
+	"fmt"
+
 	"github.com/Sifchain/sifnode/x/clp/types"
 	sdk "github.com/cosmos/cosmos-sdk/types"
 )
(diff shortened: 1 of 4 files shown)


timlind:
Marked as draft so we don't merge into develop yet, just needs to be deployed to tesnet-042 for now. Should maybe be merged into a common branch with https://github.com/Sifchain/sifnode/pull/1672

x/clp/keeper/liquidityprovider.go
```diff
@@ -71,18 +73,26 @@ func (k Keeper) DestroyLiquidityProvider(ctx sdk.Context, symbol string, lpAddre
...
-	var lpList []types.LiquidityProvider
+func (k Keeper) GetLiquidityProvidersForAsset(ctx sdk.Context, asset types.Asset) []*types.LiquidityProvider {
+	var lpList []*types.LiquidityProvider
 	iterator := k.GetLiquidityProviderIterator(ctx)
 	defer iterator.Close()
 	for ; iterator.Valid(); iterator.Next() {
 		var lp types.LiquidityProvider
 		bytesValue := iterator.Value()
+		if !(len(bytesValue) > 0) {
+			panic("GetLiquidityProvidersForAsset ==>> len(bytesValue) <= 0")
```
marshsif, line 84: Should we be panicking here? What if there are no LPs for a given asset, shouldn't that just return an error?

iverc: it is for testing purposes, the logic should be updated after fixing the issue

marshsif (approved):

x/clp/keeper/liquidityprovider.go
```diff
@@ -71,18 +73,26 @@ func (k Keeper) DestroyLiquidityProvider(ctx sdk.Context, symbol string, lpAddre
...
+		}
 		k.cdc.MustUnmarshalBinaryBare(bytesValue, &lp)
+		if lp.Asset == nil {
+			panic(fmt.Sprintf("GetLiquidityProvidersForAsset ==>> lp.Asset == nil, len(bytesValue): %d", len(bytesValue)))
```
intl-man, line 88: @iverc Should we be using panic like this? not log and return?

x/clp/keeper/liquidityprovider.go
intl-man, line 84: @iverc Should we be using panic like this? not log and return?

PR #1683, opened 2021-08-03
Bump tendermint and sdk

faddat:
These version bumps should enable smooth state sync and should not introduce breaking changes.


timlind:
Thank you @faddat!

This SDK upgrade has been bumped to v0.42.9 and merged into develop through #1711 along with some significant IBC bug fixes in the SDK.

PR #1691, opened 2021-08-05
Feature/metadata module 1671

Brando753:
- Updated protofiles to support the metadata structure and gRPC services.
- Created a metadata keeper to store and fetch token metadata structures
- Updated Claim handler to search for token metadata and save the metadata if it does not exist
- Created CLI to query metadata structure
- Created test cases for the metadata keeper


cmd/ebrelayer/types/types.go
```diff
@@ -44,6 +44,8 @@ func (d Event) String() string {
 type EthereumEvent struct {
 	To                    []byte
 	Symbol                string
+	Name                  string
+	Decimals              int32
```
ElliotFriedman, line 48: This event should contain the token address as well.

x/ethbridge/keeper/msg_server.go
```diff
@@ -146,31 +146,56 @@ func (srv msgServer) CreateEthBridgeClaim(goCtx context.Context, msg *types.MsgC
...
+
+	if !srv.Keeper.ExistsTokenMetadata(ctx, claim.DenomHash) {
+		// If this is the first time we see this token its not
+		// double pegged correct?
+		metadataMapping := types.TokenMetadataMapping{
+			DoublePegged:  false,
+			SourceAddress: claim.TokenContractAddress,
+		}
+		networkMapping := make(map[string]*types.TokenMetadataMapping)
+		networkMapping[claim.NetworkDescriptor.String()] = &metadataMapping
```
ElliotFriedman, line 167: Why is this needed?

x/ethbridge/types/claim.go
```diff
@@ -49,8 +58,48 @@ func (claim *EthBridgeClaim) GetProphecyID() []byte {
...
+	  * together to get this output:
+	  *    0xbF45BFc92ebD305d4C0baf8395c4299bdFCE9EA229wBTCWBTC
+	  *
+	  * Then, that data will be hashed with SHA256 to produce the following hash:
+	  *    179e6a6f8ab6efb5fa1f3992aef69f855628cfd27868a1be0525f40b456494ff
+	  *
+	**/
+	// No Prefix Yet....
+	// "{Network Descriptor}{ERC20 Token Address}{Decimals}{Token Name}{Token Symbol}"
+	denomHashedString := fmt.Sprintf("%d%s%d%s%s",
```
ElliotFriedman, line 93: Is this calculated the same everywhere?

x/ethbridge/types/claim.go
```diff
@@ -49,8 +58,48 @@ func (claim *EthBridgeClaim) GetProphecyID() []byte {
...
+		strings.ToLower(tokenContractAddress),
+		decimals,
+		strings.ToLower(tokenName),
+		strings.ToLower(symbol),
+	)
+
+	rawDenomHash := sha256.Sum256([]byte(denomHashedString))
+	denomHash := hex.EncodeToString(rawDenomHash[:])
+
+	return denomHash
```
ElliotFriedman, line 104: What happens if this denom begins with a non alphabetical character? Does sifnode blow up? If sifnode does blow up, how can we get around this limitation?

ElliotFriedman (changes requested):

x/ethbridge/types/claim.go
Brando753: Yes, both relayer and keeper use the same imported function.

cmd/ebrelayer/types/types.go
Brando753: It does, expand down it always included the field Token common.Address

x/ethbridge/keeper/msg_server.go
Brando753: This code is checking if a token that a claim is made against currently has its metadata stored in the metadata module. currently nothing else adds or checks this data, but its stored if not currently available. It was unclear to me how data would otherwise be stored into this module.

x/ethbridge/types/claim.go
Brando753: Good catch, I missed the regex requirement in the metadata module. Is there any reason not to just simply prefix the denomHash with an 's'?

> Based on research, newer versions of cosmos SDK allow up to 128 character symbols in the coins module and will need to match this regular expression [a-zA-Z][a-zA-Z0-9/]{2,127}. 

proposed change is: 
```
denomHash := "s" + hex.EncodeToString(rawDenomHash[:])
```

cmd/ebrelayer/types/types.go
```diff
@@ -62,6 +64,8 @@ func (e EthereumEvent) Equal(other EthereumEvent) bool {
 		e.From == other.From &&
 		bytes.Equal(e.To, other.To) &&
 		e.Symbol == other.Symbol &&
+		e.Name == other.Name &&
+		e.Decimals == other.Decimals &&
```
ElliotFriedman, line 68: We should assert that the token addresses are the same here.

x/ethbridge/keeper/msg_server.go
ElliotFriedman: I understand that we are putting the token metadata into the keeper if it doesn't already exist, there, my question is, why is the networkMapping needed and what is its usecase?

x/ethbridge/types/claim.go
ElliotFriedman: Adding a 's' prefix works for me. 

I would consider adding a 'sif' prefix instead, but either works.

ElliotFriedman (changes requested):

x/ethbridge/keeper/msg_server.go
Brando753: Network Mapping is defined in the Coin Metadata Conventions document; while it wasn't exactly clear to me how it was supposed to be used, I was trying to implement it according to this description:

> Denom => {
> 	decimals: 9
> 	name: “wBTC”
> 	symbol: “WBTC”
> 	tokenAddress: “0xbF45BFc92ebD305d4C0baf8395c4299bdFCE9EA2”
> 	networkDescriptor: “2”
> 		networkDescriptor => {
> doublePegged: true
> }
> }
> 
> The metadata included in the keeper will be the smart contract address, the decimals, the network descriptor, token name and symbol. When relayers submit a transaction to peg an asset into sifchain, they will send the token address, network descriptor, decimal amount, token name, and token symbol. Then, if that token has not already been pegged into sifchain, the token address, name, symbol, network descriptor and decimal amount will be stored in the metadata module. There will also be a nested mapping that maps a network descriptor to the smart contract address on the source chain.

ElliotFriedman: We only need to modify that mapping when we double peg an asset from sifchain to an external chain. This is a change that you will need to make to the lock function to support double pegging.

ElliotFriedman (approved):
LGTM!

ElliotFriedman:
@Brando753 It seems there are some merge conflicts here between your branch and future/peggy2.0. I would appreciate if you could resolve and then merge.

PR #1700, opened 2021-08-08
Updated hardhat functions and added a script to update contracts

banshee:
This is to support upgrading the contracts on testnets instead of just recreating them from scratch every time.


smart-contracts/hardhat.config.ts
```diff
@@ -1,24 +1,34 @@
...
         hardhat: {
             allowUnlimitedContractSize: false,
             chainId: 1,
             forking: {
-                url: alchemyUrl,
+                url: mainnetUrl,
                 blockNumber: 12865480,
             }
         },
+        ropsten: {
```
ElliotFriedman, line 28: We should add another config right here just called `mainnet` that uses the mainnnet private key and url.

smart-contracts/scripts/upgrade_contracts.ts
```diff
@@ -0,0 +1,44 @@
+import * as hardhat from "hardhat";
```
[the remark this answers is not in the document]
banshee: We're going to add the readme next to https://github.com/Sifchain/sifnode/blob/develop/smart-contracts/Deployment.md as Upgrade.md

smart-contracts/hardhat.config.ts
banshee: > We should add another config right here just called `mainnet` that uses the mainnnet private key and url.

done

ElliotFriedman (approved):
LGTM!

PR #1703, opened 2021-08-09
Migrate erowan to rowan (post ibc)

banshee:


smart-contracts/.run/upgrade_contracts.ts.run.xml
```diff
@@ -0,0 +1,5 @@
+<component name="ProjectRunConfigurationManager">
```
ElliotFriedman, line 1: Please remove

ElliotFriedman (changes requested):

smart-contracts/contracts/BridgeBank/BridgeBank.sol
```diff
@@ -289,6 +290,23 @@ contract BridgeBank is BankStorage,
...
+        uint256 amount,
+        address previousToken,
+        address newToken
+    );
+
+    function migrateERowan(
+        uint256 amount
+    ) public whenNotPaused {
+        address previousToken = controlledBridgeTokens["erowan"];
+        address newToken = controlledBridgeTokens["Rowan"];
```
ElliotFriedman, line 304: This makes the assumption that the symbol stored in the controlledBridgeTokens mapping is uppercased when it should be lowercased as all tokens are lowercased before stored in a mappinng.

smart-contracts/contracts/BridgeBank/BridgeBank.sol
```diff
@@ -289,6 +290,23 @@ contract BridgeBank is BankStorage,
         unlockFunds(_recipient, tokenAddress, symbol, _amount);
     }
 
+    event LogMigrateRowan(
+        address migrator,
```
ElliotFriedman, line 294: Please index the fields you think should be important to index.

I think `migrator` and `amount` at the very least should be indexed in this event.

ElliotFriedman (changes requested):

daniportugalgit:
Starting to look into this.

There's already a PR: https://github.com/Sifchain/sifnode/pull/1703/files#diff-f0ca71e3964e169e2598ddd80da5d8c2a173bee454b64b60abe6e0169322e6caR304-R308

I'll review it and take it to completion.

smart-contracts/contracts/BridgeBank/BridgeBank.sol
daniportugalgit: We can only have 3 indexed arguments at most and I wonder how we're going to use `amount` in our systems.
In principle, I would rather index `migrator`, `previousToken`, and `newToken`, because each user will be able to do only one migration (right?). We could query the events using `migrator` as key, instead of `amount`.

daniportugalgit:
We need to emit an event on different chains with the different native currency info.
If you send a lock tx on bsc and lock bnb, we need to custom fit a function that an admin can set the name and symbol of the native token for that smart contract so that we don’t default to eth.

smart-contracts/contracts/BridgeBank/BridgeBank.sol
banshee: I don't think that's true for controlledBridgeTokens.  There's no attempt to lower-case the symbol in useExistingBridgeToken; it takes the symbol exactly as it is by doing ```string memory _symbol = BridgeToken(_contractAddress).symbol();```.

PR #1711, opened 2021-08-10
Upgrade Cosmos SDK to v0.42.9

timlind:
Squashed commits:
[3710c7be] Upgrade SDK to v0.42.9
[2c0d6b88] Add capability module to begin blocker order.
[8c1ab4d4] Remove json output definition as it is defined in sdk now.
[d58108b7] version bumps
[ce375ae9] Bump tendermint and sdk

These version bumps should enable smooth state sync and should not introduce breaking changes.


timlind:
This has already been merged into develop through feature/ibc

PR #1712, opened 2021-08-10
Fix upgrade issues

iverc:


timlind:
Closing as a copy of #1711 that was merged into feature/ibc and then develop.

PR #1715, opened 2021-08-11
removed output flags

Jedi2002:


timlind:
Looks like this is already resolved on develop.

PR #1727, opened 2021-08-13
added store loaded

Jedi2002:


timlind (approved):
Test was breaking so I had a look at what gaia was doing since they're on the same version and test is passing now.

PR #1728, opened 2021-08-13
IBC High Precision Transfers

timlind:


iverc (approved):
I'd remove commented code if not needed, otherwise LGTM

x/ibctransfer/module.go
```diff
@@ -124,13 +132,14 @@ func (am AppModule) OnTimeoutPacket(ctx sdk.Context, packet types.Packet) (*sdk.
 	return am.cosmosAppModule.OnTimeoutPacket(ctx, packet)
```
timlind, line 132: I think this is meant to be cosmosAppModule.OnAcknowledgementPacket()

x/ibctransfer/keeper/keeper.go
```diff
@@ -0,0 +1,74 @@
+package keeper
```
timlind, line 1: Can rename this file to msg_server.go

x/ibctransfer/whitelisting.go
```diff
@@ -33,7 +34,7 @@ func OnRecvPacketWhiteListed(
 		ctx.EventManager().EmitEvent(
 			sdk.NewEvent(
 				transfertypes.EventTypePacket,
-				sdk.NewAttribute(sdk.AttributeKeyModule, types.ModuleName),
+				sdk.NewAttribute(sdk.AttributeKeyModule, types.ModuleName), // is this ethbridge on purpose?
```
timlind, line 37: No. Likely an incorrect autoimport.

x/ibctransfer/whitelisting.go
```diff
@@ -45,8 +46,97 @@ func OnRecvPacketWhiteListed(
 			Events: ctx.EventManager().Events().ToABCIEvents(),
 		}, acknowledgement.GetBytes(), nil
 	}
+	// get result of transfer receive
+	recvResult, resBytes, err := sdkAppModule.OnRecvPacket(ctx, packet)
+	// if no error and packet is returning and needs conversion: convert
+	if err != nil && IsRecvPacketReturning(packet, data) && checkRecvConvert(ctx, whitelistKeeper, packet, data) {
```
timlind, line 52: sdkAppModule.OnRecvPacket doesn't ever return an error, it converts it to an error acknowledgement packet in the returned bytes.

Will need to copy the module.OnRecvPacket code here instead of calling it, so that we can do the call to the keeper.OnRecvPacket() to get the error.

x/ibctransfer/whitelisting.go
```diff
@@ -45,8 +46,97 @@ func OnRecvPacketWhiteListed(
...
+		recvResult, err = sendConvertRecvDenom(ctx, ibcToken, convToken, bankKeeper, data)
+
+	}
+
+	// otherwise return
+	return recvResult, resBytes, err
+}
+
+func checkRecvConvert(ctx sdk.Context, whitelistKeeper tokenregistrytypes.Keeper,
+	packet channeltypes.Packet, data transfertypes.FungibleTokenPacketData) bool {
```
timlind, line 64: Let's comment and name this really well.

x/ibctransfer/whitelisting.go
```diff
@@ -45,8 +46,97 @@ func OnRecvPacketWhiteListed(
...
+	denom := GetMintedDenomFromPacket(packet, data)
+	// get token registry entry for received token
+	registryEntry := whitelistKeeper.GetIBCDenom(ctx, denom)
+	return registryEntry.IbcDenom != "" && registryEntry.Decimals > registryEntry.IbcDecimals
+}
+
+func convertRecvDenom(ctx sdk.Context, whitelistKeeper tokenregistrytypes.Keeper,
+	packet channeltypes.Packet, data transfertypes.FungibleTokenPacketData) (sdk.Coin, sdk.Coin) {
```
timlind, line 72: Also comment and name well.

x/ibctransfer/whitelisting.go
```diff
@@ -45,8 +46,97 @@ func OnRecvPacketWhiteListed(
...
+	// registryEntry := whitelistKeeper.GetIBCDenom(ctx, denom)
+	// // check if registry entry has an IBC decimal field
+	// // calculate conversion
+	// po := registryEntry.Decimals - registryEntry.IbcDecimals
+	// decAmount := sdk.NewDecFromInt(sdk.NewIntFromUint64(data.Amount))
+	// convAmountDec := IncreasePrecision(sdk.NewDecFromBigInt(decAmount.BigInt()), po)
+	// convAmount := sdk.NewIntFromBigInt(convAmountDec.RoundInt().BigInt())
+
+	// convToken := sdk.NewCoin(registryEntry.Denom, convAmount)
+	// // decode the receiver address
```
timlind, line 99: What's the story with commented code?

x/tokenregistry/keeper/keeper.go
```diff
@@ -71,6 +71,21 @@ func (k keeper) GetDenom(ctx sdk.Context, denom string) types.RegistryEntry {
 	}
 }
 
+func (k keeper) GetIBCDenom(ctx sdk.Context, ibcdenom string) types.RegistryEntry {
```
timlind, line 74: Needs a more descriptive name and comment. It get's confusing with ordinary denom / ibc usage, when this is specific to denom conversions for ibc.

x/tokenregistry/keeper/keeper.go
```diff
@@ -114,15 +129,20 @@ func (k keeper) SetDenomWhitelist(ctx sdk.Context, wl types.Registry) {
 }
 
 func (k keeper) GetDenomWhitelist(ctx sdk.Context) types.Registry {
-	var whitelist types.Registry
-	store := ctx.KVStore(k.storeKey)
-	bz := store.Get(types.WhitelistStorePrefix)
+	return types.DefaultRegistry()
```
timlind, line 132: Upgrade handler is working now so we can install registry dynamically.

x/tokenregistry/types/expected_keepers.go
```diff
@@ -10,6 +10,7 @@ type Keeper interface {
 	SetAdminAccount(ctx sdk.Context, adminAccount sdk.AccAddress)
 	IsDenomWhitelisted(ctx sdk.Context, denom string) bool
 	GetDenom(ctx sdk.Context, denom string) RegistryEntry
+	GetIBCDenom(ctx sdk.Context, denom string) RegistryEntry
```
timlind, line 13: As mentioned, needs clearer name.

x/ibctransfer/module.go
iverc: Isn't it correct?

AustinoBombino (dismissed):

x/ibctransfer/keeper/msg_server.go
```diff
@@ -0,0 +1,73 @@
...
+	// get token registry entry for sent token
+	registryEntry := srv.tokenRegistryKeeper.GetDenom(ctx, msg.Token.Denom)
+	// check if registry entry has an IBC decimal field
+	if registryEntry.IbcDenom != "" && registryEntry.Decimals > registryEntry.IbcDecimals {
+		// calculate the conversion difference and reduce precision
+		po := registryEntry.Decimals - registryEntry.IbcDecimals
+		decAmount := sdk.NewDecFromInt(msg.Token.Amount)
+		convAmountDec := ReducePrecision(decAmount, po)
+
+		convAmount := sdk.NewIntFromBigInt(convAmountDec.RoundInt().BigInt())
```
Jedi2002, line 45: It might be of value to use TruncateInt() in this case , to be extra careful, we never mint more

x/ibctransfer/keeper/msg_server.go
```diff
@@ -0,0 +1,73 @@
...
+			return nil, err
+		}
+		// send coins from module account to address
+		err = srv.bankKeeper.SendCoinsFromModuleToAccount(ctx, types.ModuleName, sender, sdk.NewCoins(convToken))
+		if err != nil {
+			return nil, err
+		}
+		msg.Token = convToken
+	}
+
```
Jedi2002, line 66: Should we emit the minted amount in events ?

x/ibctransfer/keeper/msg_server.go
```diff
@@ -0,0 +1,73 @@
...
+	return srv.sdkMsgServer.Transfer(goCtx, msg)
```
Jedi2002, line 67: Cant we just return MsgTransferResponse{},nil ?

x/ibctransfer/keeper/msg_server.go
Jedi2002: Also , the events should have different type to differntiate from sdk events

x/ibctransfer/whitelisting.go
```diff
@@ -5,48 +5,149 @@ import (
...
+) (sdk.Coin, sdk.Coin) {
+	// get token registry entry for received denom
+	denom := GetMintedDenomFromPacket(packet, data)
+	registryEntry := whitelistKeeper.GetRegistryEntry(ctx, denom)
+	// get the token amount from the packet data
+	decAmount := sdk.NewDecFromInt(sdk.NewIntFromUint64(data.Amount))
+	// calculate the conversion difference and increase precision
+	po := registryEntry.Decimals - registryEntry.IbcDecimals
+	convAmountDec := IncreasePrecision(decAmount, po)
+	convAmount := sdk.NewIntFromBigInt(convAmountDec.RoundInt().BigInt())
```
Jedi2002, line 101: Should be changed in case we decided to move to truncate .

AustinoBombino: I think he's suggesting that we use Truncate() instead of RoundInt(). 

This makes me think of another issue, if a user sends 1000000000010000000 rowan we:
- calculate convToken to be 10000000000 urowan
- send 1000000000010000000 rowan from the user to the module
- mint 10000000000 urowan
- send 10000000000 urowan via ibc
Then if the user sends the same tokens back we:
- calculate convToken to be 1000000000000000000 rowan
- send 10000000000 urowan from user to the module
- send 1000000000000000000 rowan from the module to the user
- burn 10000000000 urowan
So the user lost 10000000 rowan. I think we need to reduce precision on the send, then increase precision on that amount 
and use that for the amount that's deducted from the user.

Does this make sense to you guys?
@timlind @Jedi2002 @iverc

x/ibctransfer/keeper/msg_server.go
timlind: +1

x/ibctransfer/whitelisting.go
Jedi2002: Yeah that makes sense , 
I was actually suggesting that we always round down ,so that the network never looses an tokens as an additional safety .

AustinoBombino: Yeah I agree we should truncate. And on top of that we should determine the amount of tokens to deduct based on the truncated amount.

timlind: 👍🏻

iverc: yeah, rounding down was discussed a couple of times, so instead of bankers rounding we always round down so we never send more that needed / available. Fixing that logic would require updating test tables since expected values are calculated based on bankers rounding logic

timlind:
The new properties on token registry are a bit confusing for me. We already have a concept of an ibc denom being any denom that is an ibc denom. Considering a clearer and more future proof naming / concepts, original thought was each denom would have a “source denom” where the funds are always stored for that denom unit. Then we’d have a “destination denom” used for a specific bridge ie dest_denom_ibc. The actual precision conversion would be determined by looking up the difference between those two units. The new lookup function is currently named GetRegistryEntry which is just wrong, should be something like GetEntryByIBCDest(destDenom), but I don’t think we need that function at all if we store source and destination properties and just lookup with GetDenom. Also, we don’t / won't enforce any uniqueness on those fields in the white list so could potentially have a whitelist entry that doesn't behave as expected if two tokens have the same ibc_denom value, I don’t think we’d have that problem with the approach of using src_denom and dest_denom_ibc properties instead.

AustinoBombino:
@timlind Yeah I agree about the ibcdenom naming, didn't make that connection when I made it. The `GetRegistryEntry()` name tripped me up today when making changes. I originally named it `GetIBCDenom()` and only added it because we need to be able to lookup by `dest_denom_ibc` when processing received packets because we don't have `src_denom` in the packet. I'm confused about how we would use `GetDenom()` with only `dest_denom_ibc`.

x/ibctransfer/keeper/transfer_test.go
```diff
@@ -0,0 +1,123 @@
...
+// constructs a send from chainA to chainB on the established channel/connection
+// and sends the same coin back from chainB to chainA.
+func (suite *TransferTestSuite) TestHandleMsgTransfer() {
+	// setup between chainA and chainB
+	clientA, clientB, connA, connB := suite.coordinator.SetupClientConnections(suite.chainA, suite.chainB, exported.Tendermint)
+	channelA, channelB := suite.coordinator.CreateTransferChannels(suite.chainA, suite.chainB, connA, connB, channeltypes.UNORDERED)
+	//	originalBalance := suite.chainA.App.BankKeeper.GetBalance(suite.chainA.GetContext(), suite.chainA.SenderAccount.GetAddress(), sdk.DefaultBondDenom)
+	timeoutHeight := clienttypes.NewHeight(0, 110)
+
+	coinToSendToB := sdk.NewCoin(sdk.DefaultBondDenom, sdk.NewInt(100))
```
AustinoBombino, line 43: We'll need to add cases here with denominations that will trigger precision conversion

mccallofthewild:
@timlind @AustinoBombino The term **_IBC denom_** denotes the format `ibc/DENOMHASH`. So the name `dest_denom_ibc` implies that it stores the IBC denom for the counterparty ledger. In reality, it stores the _**base denom**_ of the counterparty ledger, specifically when exporting.

The `src` and `dest` prefixes should be limited to the scope of packets, as they depend on the context of a given transfer. 
i.e. a Sifchain channel is the `src` when sending packets, but it is the `dest` when receiving packets. 

The term **_counterparty_** succinctly clarifies this and is used consistently throughout the [spec](https://github.com/cosmos/ibc/search?l=Markdown&p=2&q=counterparty) and whitepaper.

More apt properties may be:
```
    {
      "decimals": "18",
      "denom": "cdai",
      "base_denom": "",
      "channel": "",
      "counterparty_channel": "",
      "counterparty_base_denom": "udai",
      "counterparty_decimals": "10",
      ...
    },
   ```

File: x/ibctransfer/keeper/msg_server_test.go [hunk 16]
```diff
@@ -0,0 +1,316 @@
...
+			err = app.BankKeeper.AddCoins(appCtx, sender, initCoins)
+			require.NoError(t, err)
+
+			err = keeper.PrepareToSendConvertedCoins(sdk.WrapSDKContext(appCtx), tt.args.msg, tokenDeduction, tokensConverted, app.BankKeeper)
+			require.NoError(t, err)
+		})
+	}
+}
+
+func TestOnTimeoutPacketConvert(t *testing.T) {
```
timlind, line 217: This should be moved to top level module directory into ontimeout_test.go

app/test_helpers.go
```diff
@@ -72,6 +85,85 @@ func Setup(isCheckTx bool) *SifchainApp {
 	return app
 }
 
+// SetupWithGenesisValSet initializes a new SifApp with a validator set and genesis accounts
+// that also act as delegators. For simplicity, each validator is bonded with a delegation
+// of one consensus engine unit (10^6) in the default token of the sifapp from first genesis
+// account. A Nop logger is set in SifApp.
+func SetupWithGenesisValSet(t *testing.T,
```
timlind, line 92: @iverc can these test helpers also move to your other PR?

x/tokenregistry/types/events.go
```diff
@@ -0,0 +1,14 @@
...
+
+const (
+	EventTypeConvertTransfer  = "converted_transfer"
+	EventTypeConvertReceived  = "converted_received_packet"
+	AttributeKeySentAmount    = "sent_amount"
+	AttributeKeySentDenom     = "sent_denom"
+	AttributeKeyPacketAmount  = "packet_amount"
+	AttributeKeyPacketDenom   = "packet_denom"
+	AttributeKeyConvertAmount = "converted_amount"
+	AttributeKeyConvertDenom  = "converted_denom"
```
timlind, line 13: All these should move to ibctransfer module.

x/tokenregistry/types/genesis.go
```diff
@@ -33,7 +33,13 @@ func GetGenesisStateFromAppState(marshaler codec.JSONMarshaler, appState map[str
...
+			{IsWhitelisted: true, Denom: "ibc/896F0081794734A2DBDF219B7575C569698F872619C43D18CC63C03CFB997257", BaseDenom: "atom", SrcChannel: "channel-0", DestChannel: "channel-86", Path: "transfer/channel-0", DisplayName: "ATOM", ExternalSymbol: "ATOM", Decimals: 18},
+			{IsWhitelisted: true, Denom: "ibc/48E40290A494F271890BCFC867EB0940D8A6205DD94750C8EA71750480D65BA9", BaseDenom: "akt", SrcChannel: "channel-1", DestChannel: "channel-12", Path: "transfer/channel-1", DisplayName: "AKT", ExternalSymbol: "AKT", Decimals: 18},
+			{IsWhitelisted: true, Denom: "ibc/0F3C9D893A0ADE5738E473BB1A15C44D9715568E0C005D33A02495B444E15225", BaseDenom: "ncat", SrcChannel: "channel-2", DestChannel: "channel-12", Path: "transfer/channel-2", DisplayName: "NCAT", ExternalSymbol: "NCAT", Decimals: 18},
+			{IsWhitelisted: true, Denom: "ibc/E0B9629F3DF557C3412F12F6EFE3DACB28B4A30627A27697B6CFAD03A3DE0C96", BaseDenom: "dvpn", SrcChannel: "channel-3", DestChannel: "channel-16", Path: "transfer/channel-3", DisplayName: "dVPN", ExternalSymbol: "dVPN", Decimals: 18},
+			{IsWhitelisted: true, Denom: "rowan", Decimals: 18, IbcDenom: "urowan", IbcDecimals: 10},
```
timlind, line 42: Can we remove these and still have tests work right now?

x/ibctransfer/onsend_ack.go
```diff
@@ -0,0 +1,95 @@
...
+		)
+	// if acknowledgement error then a refund was processed so we must check if conversion is necessary
+	case *channeltypes.Acknowledgement_Error:
+		ctx.EventManager().EmitEvent(
+			sdk.NewEvent(
+				transfertypes.EventTypePacket,
+				sdk.NewAttribute(transfertypes.AttributeKeyAckError, resp.Error),
+			),
+		)
+		// TODO: Always refund, not only if sender of ack is source.
```
timlind, line 73: Should this conversion logic only apply when sender is source? Does the same apply in the other direction? Is this a requirement we need in the code or just something the registry indicates?

AustinoBombino: The sdk's refund logic is triggered above at:
`	if err := sdkTransferKeeper.OnAcknowledgementPacket(ctx, packet, data, ack); err != nil {
		return nil, err
	}`
So a refund is already triggered regardless of whether sender is source. But we need to check for conversion if sender is source because the tokens that were refunded by the sdk function may have been converted tokens. The only tokens that we convert are those that originate on Sifchain.

Not sure what you mean by this:

> Is this a requirement we need in the code or just something the registry indicates?

AustinoBombino (approved):

PR #1735, opened 2021-08-15
v0.9.3-ibc

timlind:
Includes test cases for SDK panics.

The upgrade itself does not have test code as that would require more refactoring.

This should be run locally before deploying.


timlind (commented):
Tested locally (without previous error state). Upgrade works and new stores is initialised with data.

PR #1744, opened 2021-08-17
Feature/metadata admin add

Brando753:
Added support for Add/Delete of IBC token metadata.
     - proto files updated for new requests
     - tx messages updated to support transactions for changing state
     - cli interface updated to support add/delete commands
     - TokenMetadata renamed to TokenMetadataSearch
     - Unit tests added
    TO DO:
     - Integration test that supports add/delete command line

Things to verify by reviewers:

- Is the admin auth properly checked? It was modeled after the UpdateWhiteListValidator command which I believe also required admin privileges but I am unsure if the auth is handled properly


x/ethbridge/client/cli/metadata.go
```diff
@@ -36,3 +43,90 @@ func GetCmdGetTokenMetadata() *cobra.Command {
...
+			}
+
+			cosmosSender, err := sdk.AccAddressFromBech32(args[0])
+			if err != nil {
+				return err
+			}
+
+			tokenName := args[1]
+
+			tokenSymbol := args[2]
```
ElliotFriedman, line 66: We should probably check both tokenName and tokenSymbol to ensure the strings are not empty.

x/ethbridge/client/cli/metadata.go
```diff
@@ -36,3 +43,90 @@ func GetCmdGetTokenMetadata() *cobra.Command {
...
+				return errors.New("Error parsing tokenAddress invalid format must be a hex address")
+			}
+
+			tokenAddress := types.NewEthereumAddress(tokenAddressRaw)
+
+			tokenDecimalsRaw, err := strconv.ParseInt(args[4], 10, 32)
+			if err != nil {
+				return errors.New("Error parsing token decimals, must be base 10 number")
+			}
+			tokenDecimals := int32(tokenDecimalsRaw)
```
ElliotFriedman, line 79: Let's add a check here to ensure that the tokenDecimals are a positive number.

x/ethbridge/keeper/metadata.go
```diff
@@ -33,6 +47,48 @@ func (k Keeper) AddTokenMetadata(ctx sdk.Context, metadata types.TokenMetadata)
...
+		logger.Error("Token is not IBC cannot use this keeper")
+		return ""
+	}
+
+	if !k.oracleKeeper.IsAdminAccount(ctx, cosmosSender) {
+		logger.Error("cosmos sender is not admin account.")
+		return ""
+	}
+
+	denom := k.AddTokenMetadata(ctx, metadata)
```
ElliotFriedman, line 62: `AddTokenMetadata` should be an internal method to this keeper, so that only functions within this file can call it.

x/ethbridge/keeper/metadata.go
```diff
@@ -33,6 +47,48 @@ func (k Keeper) AddTokenMetadata(ctx sdk.Context, metadata types.TokenMetadata)
 	return denomHash
 }
 
+func (k Keeper) AddIBCTokenMetadata(ctx sdk.Context, metadata types.TokenMetadata, cosmosSender sdk.AccAddress) string {
+	logger := k.Logger(ctx)
+	if !IsIBCToken(metadata.Name) {
+		logger.Error("Token is not IBC cannot use this keeper")
```
ElliotFriedman, line 53: Change error message from: `Token is not IBC cannot use this keeper` to `Token is not IBC, cannot modify metadata manually`

x/ethbridge/keeper/msg_server.go
```diff
@@ -372,3 +372,52 @@ func (srv msgServer) SetFeeInfo(goCtx context.Context, msg *types.MsgSetFeeInfo)
...
+		return nil, err
+	}
+
+	account := srv.Keeper.accountKeeper.GetAccount(ctx, cosmosSender)
+	if account == nil {
+		logger.Error("account is nil.", "CosmosSender", msg.CosmosSender)
+		return nil, sdkerrors.Wrap(sdkerrors.ErrInvalidAddress, msg.CosmosSender)
+	}
+
+	if !IsIBCToken(msg.Metadata.Name) {
```
ElliotFriedman, line 391: We can remove this check as we already have a check that checks this in the keeper.

ElliotFriedman (changes requested):

x/ethbridge/keeper/metadata.go
Brando753: This method needs to be called outside this file however. For example the metadata_test unit tests will be unable to run if we do not export this method. Let me know if you still want me to make the changes.

ElliotFriedman: This function only needs to be called by the metadata_test externally to this file. Because of this, we should make the function internal so that there are no mix ups where people call this when they shouldn't be able to.

ElliotFriedman (approved):

PR #1747, opened 2021-08-17
Future/peggy2optimized

daniportugalgit:
What's in this PR:
- Gas optimization: moved things around, changed the way we declare some variables, joined functions that needed `for` loops
- A new function to **batch** update the Eth Whitelist: `batchUpdateEthWhiteList(address[],bool[])`
- `intermediateNonce` in `batchSubmitProphecyClaimAggregatedSigs(bytes32[],ClaimData[],SignatureData[][])`
- New tests to measure gas optimization (they are in test_txSignatureAggregation.js)
- New tests to guarantee batchUpdateEthWhiteList works
- Helper to print colored logs (in helpers/helpers.js)
- Added `tokenName`, `tokenSymbol`, and `tokenDecimals` to `ProphecyID`

```
Results:
Gas Cost With 4 Validators
~~~~~~~~~~~~
LOCK:
Original: 173978
Current : 160942
Diff    : -13036 (7.49%)
      ✓ should allow us to check the cost of submitting a prophecy claim lock
~~~~~~~~~~~~
MINT:
Original: 179749
Current : 176877
Diff    : -2872 (1.60%)
      ✓ should allow us to check the cost of submitting a prophecy claim mint
~~~~~~~~~~~~
DoublePeg :: New BridgeToken:
Original: 1162769
Current : 1121426
Diff    : -41343 (3.56%)
      ✓ should allow us to check the cost of creating a new BridgeToken
~~~~~~~~~~~~
BATCH, regarding previous implementation:
Original: 328673
Current : 301089
Diff    : -27584 (8.39%)
---
3 Batched claims VS single claim:
Original: 482826
Current : 301089
Diff    : -181737 (37.64%)
      ✓ should allow us to check the cost of submitting a batch prophecy claim lock
~~~~~~~~~~~~
DoublePeg :: 3 Batched claims VS single claim:
Original: 3364278
Current : 3177610
Diff    : -186668 (5.55%)
      ✓ should allow us to check the cost of batch creating new BridgeTokens
~~~~~~~~~~~~
Batch Whitelist :: 3 Batched VS single:
Original: 173325
Current : 114217
Diff    : -59108 (34.10%)
      ✓ should allow us to check the cost of adding a token to Eth Whitelist

### CODE REVIEW

File: smart-contracts/contracts/CosmosBridge.sol [hunk 1]
```diff
@@ -117,37 +117,40 @@ contract CosmosBridge is CosmosBridgeStorage, Oracle {
         address signer,
         bytes32 hashDigest,
         uint8 _v,
-		bytes32 _r,
-		bytes32 _s
+		    bytes32 _r,
```
daniportugalgit, line 120: These whitespace changes actually did not happen. I'm not sure why they're showing here.

smart-contracts/contracts/CosmosBridge.sol
```diff
@@ -117,37 +117,40 @@ contract CosmosBridge is CosmosBridgeStorage, Oracle {
...
+		    bytes32 _s
     ) private pure returns (bool) {
-		bytes32 messageDigest = keccak256(abi.encodePacked("\x19Ethereum Signed Message:\n32", hashDigest));
-		return signer == ecrecover(messageDigest, _v, _r, _s);
+		    bytes32 messageDigest = keccak256(abi.encodePacked("\x19Ethereum Signed Message:\n32", hashDigest));
```
daniportugalgit, line 123: Same thing here: there were no changes in whitespaces.

smart-contracts/contracts/CosmosBridge.sol
```diff
@@ -179,35 +188,36 @@ contract CosmosBridge is CosmosBridgeStorage, Oracle {
...
     }
 
-    // Essentially, each tx to the submitProphecyClaimAggregatedSigs needs to include a
-    // networkDescriptor or other unique identifier that only the smart contract on that chain has.
-    // If the chain identifier in the smart contract and the chain identifier in the message do not
-    // match, the transaction should fail.
     function _submitProphecyClaimAggregatedSigs(
         bytes32 hashDigest,
-        ClaimData calldata claimData,
+        ClaimData memory claimData,
```
ElliotFriedman, line 218: This should be calldata to conserve gas. If it is memory, then it will be copied from calldata to memory which will cost more.

ElliotFriedman (changes requested):

smart-contracts/contracts/CosmosBridge.sol
daniportugalgit: I know, but I tried a lot of different approaches to solve the `stack too deep` error and this was the only one that worked.
Local scopes wouldn't help here, as you mentioned in our meeting. Passing the struct instead of separate values didn't work either. I even tried to move some verifications to different places, to no avail.
After the gas optimization, we're still spending less than we were before.

smart-contracts/contracts/CosmosBridge.sol
```diff
@@ -232,48 +242,29 @@ contract CosmosBridge is CosmosBridgeStorage, Oracle {
...
-            tokenAddress = sourceAddressToDestinationAddress[claimData.tokenAddress];
+        if(claimData.doublePeg) {
+            if(!_isManagedToken(claimData.tokenAddress)) {
+              // if we are double pegging AND we don't control the token, we deploy a new smart contract
+              tokenAddress = _createNewBridgeToken(
+                  claimData.tokenSymbol,
+                  claimData.tokenName,
+                  claimData.tokenAddress,
+                  claimData.tokenDecimals,
+                  claimData.networkDescriptor
```
ElliotFriedman, line 262: Let's add a new field `cosmosDenom` that will be a part of all bridgetokens.

There could be downstream implications on pre-existing bridgetokens from doing this.

smart-contracts/contracts/CosmosBridge.sol
```diff
@@ -179,35 +188,36 @@ contract CosmosBridge is CosmosBridgeStorage, Oracle {
...
     }
 
     function submitProphecyClaimAggregatedSigs(
         bytes32 hashDigest,
         ClaimData calldata claimData,
         SignatureData[] calldata signatureData
     ) external {
+        uint256 previousNonce = lastNonceSubmitted;
+        require(previousNonce + 1 == claimData.nonce, "INV_ORD");
+
```
ElliotFriedman, line 209: As discussed on call, let's move line 213 to 209 before the call to `_submitProphecyClaimAggregatedSigs `.

smart-contracts/contracts/CosmosBridge.sol
```diff
@@ -179,35 +188,36 @@ contract CosmosBridge is CosmosBridgeStorage, Oracle {
         require(sigs.length == claims.length, "INV_CLM_LEN");
         require(sigs.length == signatureData.length, "INV_SIG_LEN");
 
+        uint256 intermediateNonce = lastNonceSubmitted;
+        require(intermediateNonce + 1 == claims[0].nonce, "INV_ORD");
+
         for (uint256 i = 0; i < sigs.length; i++) {
+            intermediateNonce++;
```
ElliotFriedman, line 195: Same comment as `submitProphecyClaimAggregatedSigs`, let's enforce ordering in this loop by checking claim data's nonce is equal to the intermediate nonce. If it isn't correct, revert.

smart-contracts/contracts/CosmosBridge.sol
```diff
@@ -179,35 +188,36 @@ contract CosmosBridge is CosmosBridgeStorage, Oracle {
         require(sigs.length == claims.length, "INV_CLM_LEN");
         require(sigs.length == signatureData.length, "INV_SIG_LEN");
 
+        uint256 intermediateNonce = lastNonceSubmitted;
+        require(intermediateNonce + 1 == claims[0].nonce, "INV_ORD");
+
```
ElliotFriedman, line 193: Move line 199 to line 193 and instead of assigning to intermediate nonce, just increment by claims.length. This would remove the need to have a reentrancy lock.

ElliotFriedman (changes requested):

smart-contracts/contracts/CosmosBridge.sol
daniportugalgit: As we talked about on our call, I'll create a new branch to implement the `cosmosDenom`

smart-contracts/contracts/CosmosBridge.sol
```diff
@@ -107,47 +110,53 @@ contract CosmosBridge is CosmosBridgeStorage, Oracle {
...
+                "INV_SIG"
+            );
+
+            pow += getValidatorPower(validator.signer);
+
             for (uint256 j = i + 1; j < _validators.length; j++) {
-                if (_validators[i].signer == _validators[j].signer) {
-                    return true;
+                if (validator.signer == _validators[j].signer) {
+                    return (true, pow);
```
ElliotFriedman, line 156: I'm not seeing the part where this function returns false if it ends up going down that path. We should explicitly rather than implicitly return in the case we see no duplicates.

ElliotFriedman (changes requested):

PR #1751, opened 2021-08-18
Future/peggy2optimized native symbols

daniportugalgit:
I branched off of future/peggy2optimized to do this because I don't think we can go back, it changed a lot of things.

This PR has everything from that branch, plus the ability to **set the native token's name and symbol during the initialization of BridgeBank**.

### DIFF
--- a/smart-contracts/contracts/CosmosBridge.sol
+++ b/smart-contracts/contracts/CosmosBridge.sol
@@ -88,14 +88,17 @@ contract CosmosBridge is CosmosBridgeStorage, Oracle {
     }
 
     function getProphecyID(
-        bytes calldata cosmosSender,
+        bytes memory cosmosSender,
         uint256 cosmosSenderSequence,
         address payable ethereumReceiver,
         address tokenAddress,
         uint256 amount,
         bool doublePeg,
         uint128 nonce,
-        uint256 _networkDescriptor
+        uint256 _networkDescriptor,
+        string memory tokenName,
+        string memory tokenSymbol,
+        uint8 tokenDecimals
     ) public pure returns (uint256) {
         return uint256(
             keccak256(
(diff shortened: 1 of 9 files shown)


smart-contracts/contracts/CosmosBridge.sol
```diff
@@ -107,47 +110,53 @@ contract CosmosBridge is CosmosBridgeStorage, Oracle {
...
-    // this is unfortunately the best we can do to ensure no duplicate validators are calling
-    // it is possible to build a hashmap in memory, but I'm unsure of how much that saves and
-    // it would require some pretty low level work for this very simple function
-    // Alternatively, cast addresses to UINT's and possibly do some bitwise operations
-    // to ensure there are no duplicate numbers
-    function findDup(SignatureData[] calldata _validators) public pure returns (bool) {
+    function getSignedPowerAndFindDup(
+        SignatureData[] calldata _validators,
+        bytes32 hashDigest
+    ) private view returns(bool, uint256 pow) {
```
ElliotFriedman, line 136: Let's remove the boolean being returned here and instead just revert if a false would have been returned.

smart-contracts/contracts/CosmosBridge.sol
```diff
@@ -232,48 +242,29 @@ contract CosmosBridge is CosmosBridgeStorage, Oracle {
...
-            require(verifyNetworkDescriptor(claimData.networkDescriptor), "INV_NET_DESC");
+            require(_verifyNetworkDescriptor(claimData.networkDescriptor), "INV_NET_DESC");
         }
         
-        // ensure there are no duplicate signers
-        require(
-            !findDup(signatureData), "DUP_SIGNER"
-        );
+        (bool dup, uint256 pow) = getSignedPowerAndFindDup(signatureData, hashDigest);
+        require(!dup, "DUP_SIGNER");
```
ElliotFriedman, line 249: Let's remove this require statement and change `getSignedPowerAndFindDup` to only return a uint.

ElliotFriedman (changes requested):

ElliotFriedman:
Let's pivot to a different architecture where instead of storing these things in storage and needing them in initialization of things, we just hardcode the native symbol to something like "evm_native".

pandaring2you:
hey @daniportugalgit - this task still valid?

daniportugalgit:
It actually is, Jordan. The change requested by Elliot was done somewhere,
but it probably stayed in a branch that was abandoned for some reason. I'll
bring it to my peggy2-blocklist branch.

daniportugalgit:
We'll deal with this task in a different branch and PR

PR #1757, opened 2021-08-20
Design spec of `localnet` tool

wojciech-sif:
`localnet` is a tool allowing us to set up the entire local ecosystem required to build and test new features.

### DIFF
--- a/tools/localnet/design.md
+++ b/tools/localnet/design.md
@@ -1,0 +1,93 @@
+# LOCALNET
+
+`localnet` is a tool allowing us to set up the entire local ecosystem required to build and test new features.
+
+## Motivation
+
+By providing new features and integrating new blockchains, more and more funds are collected in our ecosystem. It means that
+responsibility of dev team grows every day. That's why we need good tools helping us to build our confidence that new
+feature works well before pull request is even created by the author.
+
+Before IBC era, testing new features locally was pretty easy. The only required component was a single `sifnode`.
+Now, after releasing IBC it is much more complicated, and many interconnected components are needed.
+For ease of use it seems reasonable to create a tool starting and managing all of them.
+
+## Requirements
+
+To be able to play with all th features of `sifnode` (including IBC) we need these components:
+- at least two `sifnodes` running two separate chains
+- relayer between them
+- local eth testnet
+- peggy between eth and sifchain
+- block explorer for every chain
+
+
+Every time we add new supported blockchain `localnet` is extended by adding testnet, relayer  and block explorer.
+
+### Sifchain
+
+To support sifchains, a single `sifnode` per chain is started. `sifnode` should be configured to not mine empty blocks,
+to make output in block explorer more relevant and easier to inspect. IIRC there is an option for this in tendermint but
+3 years ago when I tested it, consensus was failing because of that. Maybe it has been fixed, have to test it again.
+
+It would be nice to implement some logs filter in `sifnode` to prevent regular `tendermint` consensus-related logs
+from being printed in `localnet` unless they are errors or panics. By doing this, relevant logs printed by our app are
+more meaningful and take less space on screen.
+
+[`hermes`](https://hermes.informal.systems/) is configured to relay tokens between two sifchains.
+
+[`Big Dipper`](https://github.com/forbole/big-dipper) is used for exploring blocks.
+
+`sifnode` and `hermes` are single-file pieces of software. Both may be run locally easily, so they don't require
+containers. On the other hand `Big Dipper` requires frontend and `mongo` instance, that's why it's better to execute them
+in containers (ready to use images are available there).
+
+### Ethereum
+
+Single [`geth`](https://github.com/ethereum/go-ethereum) instance is used to create ethereum testnet. It's a single executable so may be used directly,
+without container. 
+
+`peggy` is started to maintain token transmission between ethereum and sifchain. Have to discuss it with peggy
+team.
+
+[`BlockScout`](https://github.com/blockscout/blockscout) block explorer is started for ethereum. Again, it's better to containerize it.
+
+## Tools
+
+### tmate
+
+[`tmate`](https://tmate.io/) is a great tool based on [`tmux`](https://github.com/tmux/tmux/) allowing us to run many
+applications in a single terminal window. Keyboard shortcuts are used to navigate between started apps so all the logs
+may be inspected manually. `tmate` is an extension over `tmux`, supporting remote access to sessions created by others.
+
+So instead of running many terminals, one per component, all of them may be started in a single one making our life easier.
+At the same time, `tmate` provides the list of all running apps, marking those which exited (due to panic for ex.).
+
+### podman
+
+`podman` is a docker replacement with 100% compatible CLI interface. The difference is that `podman` doesn't require
+daemon and supports containers created in user-space (no `root` required). It's much lighter than `docker` and
+world is moving toward this.
+
+## CLI interface
+
+`localnet` exposes these commands:
+
+- `localnet install` - downloads all the required components, unpacks, builds them etc.
+- `localnet update` - updates components
+- `localnet start` - starts all the components in `tmate` session
+- `localnet stop` - stops `tmate` session, all components are stopped but may be (re)started again - state is preserved  between restarts  
+- `localnet destroy` - drops all the state generated/collected by the components (blocks, keys, etc.)
+
+## Block explorers
+
+Each block explorer has statically assigned IP in `127.0.0.0/24` with frontend listening on port `8080` making it easy to bookmark each of them.
+
+## Future work
+
+Once stuff described above is done, we could think about using this environment to run integration tests.
+By investing reasonable effort, tool could be created for sending transactions to different blockchains (sifchain, ethereum etc.)
+and querying them to check if result is correct.
+
+We could also use `localnet` in more complex scenarios like starting old version of `sifnode` then stopping it and replacing with a new one.
+It would allow us to check if state is correctly maintained between upgrades. 
--- a/tools/localnet/design.md
+++ b/tools/localnet/design.md
@@ -64,11 +64,13 @@ may be inspected manually. `tmate` is an extension over `tmux`, supporting remot
 So instead of running many terminals, one per component, all of them may be started in a single one making our life easier.
 At the same time, `tmate` provides the list of all running apps, marking those which exited (due to panic for ex.).
 
-### podman
+### docker / podman
 
 `podman` is a docker replacement with 100% compatible CLI interface. The difference is that `podman` doesn't require
 daemon and supports containers created in user-space (no `root` required). It's much lighter than `docker` and
 world is moving toward this.
+
+Both `docker` and `podman` are supported by `localnet`.  
 
 ## CLI interface
 
--- a/tools/localnet/design.md
+++ b/tools/localnet/design.md
@@ -21,14 +21,15 @@ To be able to play with all th features of `sifnode` (including IBC) we need the
 - peggy between eth and sifchain
 - block explorer for every chain
 
+User selects which blockchains are started.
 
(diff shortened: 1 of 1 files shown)


tools/localnet/design.md
```diff
@@ -0,0 +1,93 @@
...
+
+To be able to play with all th features of `sifnode` (including IBC) we need these components:
+- at least two `sifnodes` running two separate chains
+- relayer between them
+- local eth testnet
+- peggy between eth and sifchain
+- block explorer for every chain
+
+
+Every time we add new supported blockchain `localnet` is extended by adding testnet, relayer  and block explorer.
```
marshsif, line 25: we'll probably want to create some parameterized config, to set which blockchain(s) the localnet should include, otherwise it could get unwieldy if we have 5+. Also wondering how we address block history for a given chain?

marshsif (commented):
we've needed things like this for a while, but haven't had the bandwidth. Maybe now we can begin to improve our overall dev.

tools/localnet/design.md
wojciech-sif: > we'll probably want to create some parameterized config, to set which blockchain(s) the localnet should include, otherwise it could get unwieldy if we have 5+

yes, I thought about it, we may provide some CLI flags: `--with-blockchain-ethereum`, `--with-blockchain-<blockchain-name>`

> Also wondering how we address block history for a given chain?

What do you mean? Every blockchain is a separate "database" stored in separate directory and accessed by querying its node

Jedi2002:
This looks like a great start , Ill try to list down my thoughts 
- The networks should be multi-node ,with a config for the number of nodes . This would allow us to test consensus and app hash errors. So SifChain-Id1 (4 validators + 1 full node) and SifChain-Id2 (4 validators + 1 full node) 
- It would be usefull to run these nodes in containers to avoid issues with ports etc ( Would be easier to maintain config )
- Would be nice to have optional componets , for example we dont need to run the peggy setup to run any tests that we have for sifnode. Same for multiple chains and ibc relayers.
- Should be possible to start a chain with state data from another chain in a straightforward way. 
- Explorer seem to me like a good to have option ,and not super mandtory , but I might have missed some usecases . Wallet integration might be good though .
- Not sure if running with empty blocks off is such a good idea , as it wont emulate the real netowork ( we can always keep a track of non empty blocks separetely if required  )

wojciech-sif:
@Jedi2002 

> It would be usefull to run these nodes in containers to avoid issues with ports etc

You need different IP/port pair for each node anyway. It doesn't matter if it's started directly or in container

> Should be possible to start a chain with state data from another chain in a straightforward way.

Do you mean another local chain or devnet/testnet? Taking data from devnet would be a superpower but it means our laptops have to sync 100s GBs of data

> Explorer seem to me like a good to have option ,and not super mandtory , but I might have missed some use cases

If I want to play with some feature it's simpler IMO to check result in nice UI rather than anylising JSON manually in terminal

> Not sure if running with empty blocks off is such a good idea , as it wont emulate the real netowork

I would like them to not appear in block explorer (for ease of inspection). Maybe there is an option to filter them out in Big Dipper. Will check it. If it's not possible then probably option for turning them off (on by default)  will be added.

wojciech-sif:
@jedi2002 please place your comments on specific, related lines. It's easier to discuss each thing in dedicated thread.

Jedi2002:
@wojciech-sif  , my formatting skills are average at best , Not sure if this makes the discussion easier 😄 
 
>> It would be usefull to run these nodes in containers to avoid issues with ports etc
 
>>> You need different IP/port pair for each node anyway. It doesn't matter if it's started directly or in container

>>>> Yes definitly we would need ports , but managing the port mappings would easier in a docker file .Also,the docker file would efficient if we want to replicate issues in some other enviorment.

---

>> Should be possible to start a chain with state data from another chain in a straightforward way.
 >>> Do you mean another local chain or devnet/testnet? Taking data from devnet would be a superpower but it means our laptops have to sync 100s GBs of data
>>>>Not sure if we are talking about the same things here, I was referring to exporting the state of a running network in a json file ,and starting a new network with a pre populated state. You dont need to run a node to sync up , you should be able to extract the export through one of the existing nodes , just need access to sifnoned process 

---
>> Explorer seem to me like a good to have option ,and not super mandtory , but I might have missed some use cases
 >>> If I want to play with some feature it's simpler IMO to check result in nice UI rather than anylising JSON manually in terminal
>>>> I was imagining this to setup to be oriented towards tests ,and we would be doing all of this programitically. It would be easier to parse json into structs . And add assertions
---
>> Not sure if running with empty blocks off is such a good idea , as it wont emulate the real netowork
>>> I would like them to not appear in block explorer (for ease of inspection). Maybe there is an option to filter them out in Big Dipper. Will check it. If it's not possible then probably option for turning them off (on by default) will be added.
>>>> Hmm , again to my earlier point , in case we are doing this programitically , filtering out empty vs non empty blocks is a trivial task .
---

tools/localnet/design.md
```diff
@@ -0,0 +1,95 @@
...
+more meaningful and take less space on screen.
+
+[`hermes`](https://hermes.informal.systems/) is configured to relay tokens between two sifchains.
+
+[`Big Dipper`](https://github.com/forbole/big-dipper) is used for exploring blocks.
+
+`sifnode` and `hermes` are single-file pieces of software. Both may be run locally easily, so they don't require
+containers. On the other hand `Big Dipper` requires frontend and `mongo` instance, that's why it's better to execute them
+in containers (ready to use images are available there).
+
```
timlind, line 45: Both Sifnode and relayers usually store directories in the user's home folder, so storing this on container can be useful to create isolation.

tools/localnet/design.md
```diff
@@ -0,0 +1,95 @@
...
+## Tools
+
+### tmate
+
+[`tmate`](https://tmate.io/) is a great tool based on [`tmux`](https://github.com/tmux/tmux/) allowing us to run many
+applications in a single terminal window. Keyboard shortcuts are used to navigate between started apps so all the logs
+may be inspected manually. `tmate` is an extension over `tmux`, supporting remote access to sessions created by others.
+
+So instead of running many terminals, one per component, all of them may be started in a single one making our life easier.
+At the same time, `tmate` provides the list of all running apps, marking those which exited (due to panic for ex.).
```
timlind, line 65: Suggest checking out the UI yarn stack setup here.

tools/localnet/design.md
```diff
@@ -0,0 +1,95 @@
...
+
+`localnet` exposes these commands:
+
+- `localnet install` - downloads all the required components, unpacks, builds them, installs/updates them
+- `localnet start` - starts all the components in `tmate` session
+  - `--with-chain=<blockchain>` adds `blockchain` to the setup
+  - `--with-explorers` starts block explorers for each blockchain
+- `localnet stop` - stops `tmate` session, all components are stopped but may be (re)started again - state is preserved  between restarts  
+- `localnet destroy` - drops all the state generated/collected by the components (blocks, keys, etc.)
+
```
timlind, line 83: Would like to be using Docker compose.

tools/localnet/design.md
```diff
@@ -0,0 +1,95 @@
...
+### tmate
+
+[`tmate`](https://tmate.io/) is a great tool based on [`tmux`](https://github.com/tmux/tmux/) allowing us to run many
+applications in a single terminal window. Keyboard shortcuts are used to navigate between started apps so all the logs
+may be inspected manually. `tmate` is an extension over `tmux`, supporting remote access to sessions created by others.
+
+So instead of running many terminals, one per component, all of them may be started in a single one making our life easier.
+At the same time, `tmate` provides the list of all running apps, marking those which exited (due to panic for ex.).
+
+### podman
```
timlind, line 67: Is podman Linux only?

tools/localnet/design.md
```diff
@@ -0,0 +1,95 @@
...
+
+`podman` is a docker replacement with 100% compatible CLI interface. The difference is that `podman` doesn't require
+daemon and supports containers created in user-space (no `root` required). It's much lighter than `docker` and
+world is moving toward this.
+
```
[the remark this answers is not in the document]
wojciech-sif: We may support both and use any of them depending on what is installed. I prefer `podman` because fedora dropped support for `docker`

tools/localnet/design.md
wojciech-sif: But there is always a CLI flag to modify the path to the config/data dir. I don't see much sense in running software in container for testing purposes because of one directory

tools/localnet/design.md
wojciech-sif: WDYM?

tools/localnet/design.md
wojciech-sif: Using docker composer implies that all the apps are containerized which I want to avoid. It's faster to just compile `sifnoded` and run it directly without building docker image, which takes time.

tools/localnet/design.md
wojciech-sif: There's no podman for Windows, there are some tricks to run it on macos. Ok, so we will suport both docker and podman. It's easy.

wojciech-sif:
@Jedi2002 please create separate thread for each topic you are discussing because it's impossible to follow comments this way.
Please bind each topic to the related line in the file

khdegraaf:
Just a note that this significantly overlaps both historical and ongoing efforts at improvement from the Test engineering team, the Front End team, and the Peggy team.  It would likely be useful to sync up and discuss these efforts.  See also the document posted on #eng-team-testing on this subject.

wojciech-sif:
@khdegraaf I had great discussion with Jure about this. We have set our goals and will join our efforts to prepare a great tool for all the teams involved.

tools/localnet/design.md
59023g: Hey - https://github.com/Sifchain/sifnode/blob/develop/ui/README.md#docker-registry - It's a pretty comprehensive workflow that creates and uploads docker images to our ghcr repo. We use it for compatibility testing and ABI contracts in our repo, but I imagine there's a lot of overlap. 

Workflow: 
https://github.com/Sifchain/sifnode/blob/develop/.github/workflows/ui-stack.yml https://github.com/Sifchain/sifnode/runs/3407361543

intl-man:
> @khdegraaf I had great discussion with Jure about this. We have set our goals and will join our efforts to prepare a great tool for all the teams involved.

Would appreciate it if we (ChainOps) could be involved here. There are many intricacies which other teams are probably not aware of. Not only that, but whatever direction we take must be consistent with what we're running across the board (sorry, `podman` is nothing we run internally so I question why it's needed). I agree with @khdegraaf,  it smells like a massive duplication of effort.

timlind:
> > @khdegraaf I had great discussion with Jure about this. We have set our goals and will join our efforts to prepare a great tool for all the teams involved.
> 
> Would appreciate it if we (ChainOps) could be involved here. There are many intricacies which other teams are probably not aware of. Not only that, but whatever direction we take must be consistent with what we're running across the board (sorry, `podman` is nothing we run internally so I question why it's needed). I agree with @khdegraaf, it smells like a massive duplication of effort.

Agreed @intl-man. This should be an effort the team leads direct and led by @khdegraaf and their existing work.

For now this spec is an exercise from our side and a good opportunity for @wojciech-sif to get familiar with our setup and more comfortable with own. This definitely seems to something that would be considered falling under the category of what the testing team highlighted as frequently occurring "duplicate work" / personal hacks in their doc regarding the current setup and future plans.

We're going to hold off making any changes in the repo around this spec until we get to collaborating on the plan @khdegraaf is leading. Looking forward to that.

wojciech-sif:
@intl-man @timlind 

I agree, there are 4 parties involved in the process: devops, testers, developers and managers. I had a great discussion with @jzvikart about requirements and ideas he has already collected. Now, I'm preparing a very basic MVP to be able to present those ideas. I want to run demo for all the interested parties next week and then listen to opinions and get the list of next 3 hottest features being expected.

PR #1760, opened 2021-08-20
Future/peggy2optimized cosmos denom

daniportugalgit:
Include the `cosmosDenom` in the flow.

- [x] Add one more field into the `claimData` struct: `cosmosDenom`
- [x] BridgeToken should know its denom (interfaces in bridgeBank, and the constructor in bridgetoken)
- [x] BridgeBank should map token addresses to the cosmos denoms
- [x] BridgeBank: when you burn a token, it pulls the native denom out of the token and informs it in the event.
- [x] Special function in BridgeBank to burn old assets: if the token doesn't have a denom, revert

All tasks should include unit tests.

EDIT: adding more tasks:
- [x] Add a natspec comment at the start of each function
  - [x] BridgeBank
  - [x] CosmosBank
  - [x] CosmosBankStorage
  - [x] EthereumBankStorage
  - [x] BankStorage
  - [x] BridgeToken
  - [x] CosmosWhiteList
  - [x] CosmosWhiteListStorage
  - [x] EthereumWhiteList
  - [x] Pausable
  - [x] PauserRole
  - [x] CosmosBridge
  - [x] CosmosBridgeStorage
  - [x] Oracle
  - [x] OracleStorage
  - [x] Valset
  - [x] ValsetStorage
  - [x] BridgeRegistry
- [x] `LogBurn` event should include the cosmos denom
- [x] Make a setter function to change the `cosmosDenom` on the bridge token by the owner which is the bridgebank
- [x] Review gas optimization and check if we can do something differently
- [x] Improve visibility of functions as reported by Slither
- [x] Solve shadowing of variables (this point is on hold and being discussed)
- [x] Solve any other PR comments for the very first comments I got with the draft PR
- [x] Add a function to AddExistingBridgeTokens in bulk
- [x] Add a function `forceSetBridgeTokenDenom`, and its companion `batchForceSetBridgeTokenDenom`
- [x] Make the IbcToken become the new BridgeToken


smart-contracts/contracts/CosmosBridge.sol
```diff
@@ -189,25 +206,20 @@ contract CosmosBridge is CosmosBridgeStorage, Oracle {
...
+        // update the nonce
+        lastNonceSubmitted = claimData.nonce;
+
         _submitProphecyClaimAggregatedSigs(hashDigest, claimData, signatureData);
     }
 
-    // Essentially, each tx to the submitProphecyClaimAggregatedSigs needs to include a
-    // networkDescriptor or other unique identifier that only the smart contract on that chain has.
-    // If the chain identifier in the smart contract and the chain identifier in the message do not
-    // match, the transaction should fail.
```
ElliotFriedman, line 198: Let's add a natspec comment at the start of each function.

smart-contracts/test/helpers/testFixture.js
```diff
@@ -120,7 +125,17 @@ function initState({
   const sender = web3.utils.utf8ToHex("sif1nx650s8q9w28f2g3t9ztxyg48ugldptuwzpace");
   const state = {
     constants: {
-      zeroAddress: ZERO_ADDRESS
+      zeroAddress: ZERO_ADDRESS,
+      cosmos: {
+        denom: {
+          none: "",
+          rowan: "rowanCosmosDenomHere",
```
ElliotFriedman, line 132: We can just put "rowan" for the value here.

smart-contracts/test/test_txSignatureAggregation.js
```diff
@@ -195,16 +219,267 @@ describe("Gas Cost Tests", function () {
...
+        console.log("~~~~~~~~~~~~\nTotal: ", Number(receipt.gasUsed));
+      }
 
       const newlyCreatedTokenAddress = await state.cosmosBridge.sourceAddressToDestinationAddress(state.token1.address);
       expect(newlyCreatedTokenAddress).to.be.equal(expectedAddress);
     });
+
+    it("should allow us to check the cost of submitting a batch prophecy claim lock", async function () {
+      // Add tokens 2 and 3 into white list
+      await batchAddTokensToEthWhitelist(state, [state.token2.address, state.token3.address])
```
ElliotFriedman, line 238: After we call this batch add, let's do an explicit check where we call the contract and ensure they are whitelisted.

smart-contracts/contracts/CosmosBridge.sol
```diff
@@ -296,17 +291,18 @@ contract CosmosBridge is CosmosBridgeStorage, Oracle {
...
-        string calldata name,
+        string memory symbol,
+        string memory name,
         address sourceChainTokenAddress,
         uint8 decimals,
-        uint256 _networkDescriptor
-    ) internal returns(address) {
+        uint256 _networkDescriptor,
+        string memory cosmosDenom
+    ) internal returns(address tokenAddress) {
```
ElliotFriedman, line 305: Seems like we are shadowing a declaration on line 313 with this return value. Instead of doing `returns(address tokenAddress)` we can just do `returns(address)` as that's the style we use across the codebase.

smart-contracts/contracts/CosmosBridge.sol
```diff
@@ -179,7 +192,11 @@ contract CosmosBridge is CosmosBridgeStorage, Oracle {
         require(sigs.length == claims.length, "INV_CLM_LEN");
         require(sigs.length == signatureData.length, "INV_SIG_LEN");
 
+        uint256 intermediateNonce = lastNonceSubmitted;
+        lastNonceSubmitted += claims.length;
```
ElliotFriedman, line 196: Let's merge this code in on the other branch so this change doesn't show on this PR.

smart-contracts/contracts/CosmosBridge.sol
```diff
@@ -107,47 +112,54 @@ contract CosmosBridge is CosmosBridgeStorage, Oracle {
...
-    // it would require some pretty low level work for this very simple function
-    // Alternatively, cast addresses to UINT's and possibly do some bitwise operations
-    // to ensure there are no duplicate numbers
-    function findDup(SignatureData[] calldata _validators) public pure returns (bool) {
+    function getSignedPowerAndFindDup(
+        SignatureData[] calldata _validators,
+        bytes32 hashDigest
+    ) private view returns(bool, uint256 pow) {
         for (uint256 i = 0; i < _validators.length; i++) {
+            SignatureData memory validator = _validators[i];
```
ElliotFriedman, line 141: This change might be applied to another PR, but we don't want to copy this struct to memory if we don't have to as that costs gas. Unless we are hitting stack too deep errors by not copying to memory, we shouldn't do it.

ElliotFriedman (changes requested):

smart-contracts/contracts/BridgeBank/BridgeBank.sol
```diff
@@ -500,10 +516,6 @@ contract BridgeBank is BankStorage,
         // decimals defaults to 18 if call to decimals fails
         uint8 decimals = getDecimals(tokenAddress);
 
-        // Get name and symbol
-        string memory name = getName(tokenAddress);
-        string memory symbol = getSymbol(tokenAddress);
```
ElliotFriedman, line 505: We should call out to the bridgetoken here and get the cosmos denom and then modify the `LogBurn` message to emit the cosmos denom in the message.

smart-contracts/contracts/BridgeBank/BridgeToken.sol
```diff
@@ -12,13 +12,15 @@ import "@openzeppelin/contracts/access/Ownable.sol";
 contract BridgeToken is ERC20Burnable, Ownable {
```
ElliotFriedman, line 12: Let's add a BridgeToken interface here that has all of the public functions on it so that the bridgebank can understand how to pull the cosmosDenom.

ElliotFriedman (changes requested):

smart-contracts/contracts/CosmosBridge.sol
daniportugalgit: Yes, so here's an interesting result:

- If I use `returns(address)`, declare `tokenAddress` in the scope of the function, and then return it 'manually' at the end of the scope, Hardhat tells me that we're spending 1.467.248 gas.

- If I use `returns(address tokenAddress)`, DON'T redeclare `tokenAddress` in the scope of the function, and DON'T return it 'manually', Hardhat tells me that we're spending 1.467.256; it's almost the same amount of gas than before.

- BUT, if I shadow the declaration of tokenAddress by having it both in `returns(address tokenAddress)` AND redeclared in the scope of the function, then Hardhat tells me we're spending 1.429.398 (a 2.65% improvement).

It *is* weird. Should we consider that Hardhat is mistakenly reporting gas usage? I don't understand why we're saving gas by doing this convoluted thing, it sure looks like a mistake.

smart-contracts/contracts/BridgeBank/BridgeBank.sol
```diff
@@ -194,51 +244,80 @@ contract BridgeBank is BankStorage,
...
-     * @dev: Creates a new BridgeToken
-     *
-     * @param contractAddress: The new BridgeToken's address
+    /**
+     * @notice Allows the owner to add `contractAddress` as an existing BridgeToken
+     * @dev Adds the token to Cosmos Whitelist
+     * @param contractAddress The token's address
+     * @return New value of if token is in whitelist (true)
      */
     function addExistingBridgeToken(
```
ElliotFriedman, line 308: We should add a function to AddExistingBridgeTokens in bulk.

smart-contracts/contracts/BridgeBank/BridgeBank.sol
```diff
@@ -257,18 +336,22 @@ contract BridgeBank is BankStorage,
...
+     * @param amount How much will be burned
      */
     function burn(
         bytes calldata recipient,
         address token,
         uint256 amount
     ) external validSifAddress(recipient) onlyCosmosTokenWhiteList(token) whenNotPaused {
+        // revert if the token doesn't have a denom
+        string memory denom = getDenom(token);
+        require(keccak256(abi.encodePacked(denom)) != keccak256(abi.encodePacked("")), "INV_DENOM");
```
ElliotFriedman, line 353: Not sure we should hard assert this. I think it would be better to just ensure that things are in the whitelist

smart-contracts/contracts/BridgeBank/BridgeBank.sol
```diff
@@ -347,12 +444,35 @@ contract BridgeBank is BankStorage,
...
-     * @param recipient: bytes representation of destination address.
-     * @param token: token address in origin chain (0x0 if ethereum)
-     * @param amount: value of deposit
+    /**
+     * @dev Fetches the denom of a token by address
+     * @param token The bridgeTokens's address
+     * @return The bridgeTokens's denom or ''
+     */
+    function getDenom(address token) private returns (string memory) {
+        string memory denom = contractDenom[token];
```
ElliotFriedman, line 453: Would be great to see if we could pack all the things, cosmos denom, token name, symbol and decimals into a single struct for gas savings.

ElliotFriedman (changes requested):

smart-contracts/contracts/BridgeBank/BridgeBank.sol
daniportugalgit: Ok, I've just tested it. In this specific case, we can't save gas by packing these variables because of their types.
We have 3 `string` and one `uint8`, so we'll always use 4 storage slots no matter what. Gas costs stay the same in most cases and operations end up being more expensive whenever we only need to query/set fewer variables.
I'm dropping this one unless you feel we should discuss it further, ok?

smart-contracts/contracts/BridgeBank/BridgeBank.sol
daniportugalgit: Is there a situation where a token that's in the whitelist doesn't have a denom?
If it's added 'manually' through `addExistingBridgeToken`, it might not have a denom (nothing enforces it right now).

daniportugalgit: As discussed via Slack:
"In the case that the tokens do not have a cosmosDenom, but are still in the whitelist, they should be burned"
I'll implement it.

smart-contracts/contracts/BridgeBank/BridgeBank.sol
daniportugalgit: As discussed via Slack:
"Sounds good, we don’t need to move to using a struct"

smart-contracts/contracts/BridgeBank/BridgeBank.sol
```diff
@@ -574,4 +717,23 @@ contract BridgeBank is BankStorage,
 
         emit LogUnlock(recipient, token, amount);
     }
+
+    function setBridgeTokenDenom(
+      address _token, string memory _denom
+    ) public onlyOperator returns (bool) {
```
ElliotFriedman, line 723: There should just be a function called ```forceSetBridgeTokenDenom``` which anyone can call that automatically sets things, and another separate function that does what this function does.

This function should change to onlyOwner.

smart-contracts/contracts/BridgeBank/BridgeBank.sol
```diff
@@ -574,4 +717,23 @@ contract BridgeBank is BankStorage,
...
+      contractDenom[_token] = _denom;
+      return BridgeToken(_token).setDenom(_denom);
+    }
+
+    function batchSetBridgeTokenDenom(
```
ElliotFriedman, line 728: Same comment, remove the string from the parameter to this function, and remove the permissions such that anyone can call the function.

smart-contracts/contracts/BridgeBank/BridgeToken.sol
```diff
@@ -8,25 +8,44 @@ import "@openzeppelin/contracts/access/Ownable.sol";
...
     function decimals() public override view returns (uint8) {
         return _decimals;
     }
+
+    /**
+     * @notice Sets the cosmosDenom
+     * @param denom The new cosmos denom
+     * @return true if the operation succeeds
+     */
+    function setDenom(string calldata denom) external onlyOwner returns (bool) {
```
ElliotFriedman, line 47: Looks like you need to rebase off the latest peggy2.0 branch to resolve this and make this contract use RBAC.

ElliotFriedman (changes requested):

smart-contracts/contracts/BridgeBank/BridgeToken.sol
daniportugalgit: Actually, the new token was IbcToken.sol instead of BridgeToken.sol.
I'll bring the changes from IbcToken.sol to BridgeToken.sol and adjust the tests.

smart-contracts/contracts/BridgeBank/BridgeBank.sol
```diff
@@ -275,6 +370,9 @@ contract BridgeBank is BankStorage,
         // decimals defaults to 18 if call to decimals fails
         uint8 decimals = getDecimals(token);
 
+        // Denom defaults to "" (empty string) if call to cosmosDenom fails
+        string memory denom = getDenom(token);
```
ElliotFriedman, line 374: We will need to add the migration code to an IbcToken for the new rowan contract so that we can migrate from erowan to rowan and have the new rowan contract play nice with this system.

smart-contracts/contracts/BridgeBank/BridgeBank.sol
```diff
@@ -484,43 +623,47 @@ contract BridgeBank is BankStorage,
...
         // burn tokens
-        tokenToTransfer.burnFrom(
-            msg.sender,
-            tokenAmount
-        );
+        tokenToTransfer.burnFrom(msg.sender, tokenAmount);
+
+        // revert if the token doesn't have a denom
+        string memory denom = getDenom(tokenAddress);
+        require(keccak256(abi.encodePacked(denom)) != keccak256(abi.encodePacked("")), "INV_DENOM");
```
ElliotFriedman, line 646: Add an if statement here that checks if the bridgetoken address is equal to the hardcoded erowan token address, if it is, then don't do that require statement and assign the denom to `rowan`, else go into a block that has the require statement in it.

ElliotFriedman: This will only be able to be tested off of a mainnet fork.

smart-contracts/contracts/BridgeBank/BridgeBank.sol
```diff
@@ -484,43 +623,47 @@ contract BridgeBank is BankStorage,
...
 
         // decimals defaults to 18 if call to decimals fails
         uint8 decimals = getDecimals(tokenAddress);
 
-        // Get name and symbol
-        string memory name = getName(tokenAddress);
-        string memory symbol = getSymbol(tokenAddress);
```
ElliotFriedman, line 505: The alternative to the above solution is to get the token symbol and do a string compare using keccak256, and if the strings equal each other, then don't require the denom to be not empty.

smart-contracts/contracts/BridgeBank/BridgeBank.sol
```diff
@@ -546,17 +689,17 @@ contract BridgeBank is BankStorage,
             amount,
             lockBurnNonce,
             decimals,
-            symbol,
```
ElliotFriedman, line 549: Let's remove lines 680 and 681, this will save us on gas as we already have this stuff on the stack when emitting the log event. The reason this will result in a gas savings is due to not having to read from storage to fetch the name and symbol.

smart-contracts/contracts/BridgeBank/BridgeBank.sol
```diff
@@ -574,4 +717,65 @@ contract BridgeBank is BankStorage,
...
+     * @notice Changes the denom of `_token` to `_denom`
+     * @dev Will set the denom both in this contract AND in the token itself
+     * @param _token Address of the BridgeToken
+     * @param _denom The Cosmos denom to be applied
+     * @return true if the operation succeeded
+     */
+    function setBridgeTokenDenom(
+      address _token, string memory _denom
+    ) public onlyOwner returns (bool) {
+      contractDenom[_token] = _denom;
```
ElliotFriedman, line 731: Let's package up the logic for this function into a function called `_ setBridgeTokenDenom`, then make all the parameters calldata and the function private.

Have `batchSetBridgeTokenDenom` call that function inside it's loop.

ElliotFriedman: Then we can make this function external and cut down on gas costs.

smart-contracts/contracts/BridgeBank/BridgeBank.sol
```diff
@@ -574,4 +717,65 @@ contract BridgeBank is BankStorage,
...
+    }
+
+    /**
+     * @notice Sets in this contract the denom of `_token`
+     * @dev Will fetch the denom from `_token` and register it in this contract
+     * @dev Anyone may call this function
+     * @param _token Address of the BridgeToken
+     * @return true if the operation succeeded
+     */
+    function forceSetBridgeTokenDenom(address _token) external returns (bool) {
```
ElliotFriedman, line 761: Let's enforce that the address is in the cosmos whitelist before allowing any setting to happen. So we can just pull in `onlyCosmosTokenWhitelist` modifier here to enforce this check.

smart-contracts/contracts/BridgeBank/BridgeBank.sol
```diff
@@ -574,4 +717,65 @@ contract BridgeBank is BankStorage,
...
+    /**
+     * @notice Sets in this contract the denom of a list of BridgeTokens     * @dev Will fetch the denom from each token and register it in this contract
+     * @dev Will fetch the denom from each token and register it in this contract
+     * @dev Anyone may call this function
+     * @param _tokens List of address of BridgeTokens
+     * @return true if the operation succeeded
+     */
+    function batchForceSetBridgeTokenDenom(address[] memory _tokens) external returns (bool) {
+      for (uint256 i = 0; i < _tokens.length; i++) {
+        contractDenom[_tokens[i]] = BridgeToken(_tokens[i]).cosmosDenom();
```
ElliotFriedman, line 776: Let's add a require statement here asserting that each asset about to be updated is inside of the cosmos token whitelist before setting the contractDenom mapping.

smart-contracts/contracts/CosmosBridge.sol
```diff
@@ -179,35 +271,46 @@ contract CosmosBridge is CosmosBridgeStorage, Oracle {
...
     function submitProphecyClaimAggregatedSigs(
         bytes32 hashDigest,
         ClaimData calldata claimData,
         SignatureData[] calldata signatureData
     ) external {
+        uint256 previousNonce = lastNonceSubmitted;
+        require(previousNonce + 1 == claimData.nonce, "INV_ORD");
+
+        // update the nonce
+        lastNonceSubmitted = claimData.nonce;
```
ElliotFriedman, line 298: Let's create a test where an ERC20 token tries to pull a reentrancy attack on this code and it receives an `INV_ORD` error upon reentry.

smart-contracts/contracts/CosmosBridge.sol
```diff
@@ -232,48 +336,30 @@ contract CosmosBridge is CosmosBridgeStorage, Oracle {
...
         if (!claimData.doublePeg) {
-            require(verifyNetworkDescriptor(claimData.networkDescriptor), "INV_NET_DESC");
+            require(_verifyNetworkDescriptor(claimData.networkDescriptor), "INV_NET_DESC");
         }
         
-        // ensure there are no duplicate signers
-        require(
-            !findDup(signatureData), "DUP_SIGNER"
-        );
+        (bool dup, uint256 pow) = getSignedPowerAndFindDup(signatureData, hashDigest);
```
ElliotFriedman, line 342: Let's remove the `dup` variable and just add a require statement inside of the `getSignedPowerAndFindDup` function.

smart-contracts/contracts/CosmosBridge.sol
```diff
@@ -292,21 +378,42 @@ contract CosmosBridge is CosmosBridgeStorage, Oracle {
...
+     * @param sourceChainTokenAddress Address of the token on its original chain
+     * @param decimals The number of decimals this token has
+     * @param _networkDescriptor Unique identifier of the network
+     * @param cosmosDenom The token's Cosmos denom
+     * @return tokenAddress : The address of the newly deployed BridgeToken
+     */
     function _createNewBridgeToken(
-        string calldata symbol,
-        string calldata name,
+        string memory symbol,
```
ElliotFriedman, line 410: If we can make this into calldata, we should.

Let's do some gas benchmarking here and see how much things cost using memory.

smart-contracts/contracts/CosmosBridge.sol
```diff
@@ -292,21 +378,42 @@ contract CosmosBridge is CosmosBridgeStorage, Oracle {
         );
     }
 
-    function _isManagedToken(address tokenAddress) private returns(bool) {
+    /**
+     * @dev Verifies if `tokenAddress` is a known token
+     * @param tokenAddress The address of the token
+     * @return Boolean: is `tokenAddress` a known token?
+     */
+    function _isManagedToken(address tokenAddress) private view returns(bool) {
```
ElliotFriedman, line 386: Let's change this function name to `_isDoublePeggedToken` as this is really checking to see if an asset is double pegged or not.

smart-contracts/contracts/OracleStorage.sol
```diff
@@ -18,27 +22,27 @@ contract OracleStorage {
...
     */
-    mapping(uint256 => mapping(address => bool)) public hasMadeClaim;
+    mapping(uint256 => mapping(address => bool)) private hasMadeClaim;
 
     /**
-    * @notice mapping of prophecyid to check if it has been redeemed
+    * @dev {DEPRECATED}
     */
-    mapping(uint256 => bool) public prophecyRedeemed;
+    mapping(uint256 => bool) private prophecyRedeemed;
```
ElliotFriedman, line 37: Is this used anywhere in the current or legacy code?

If not, let's remove.

ElliotFriedman: If this data structure is not in the current code off of develop, we can delete it altogether and will just need to readjust the storage padding as a result of this.

ElliotFriedman (changes requested):

smart-contracts/test/testBridgeBank.js
```diff
@@ -169,4 +170,241 @@ describe("Test Bridge Bank", function () {
...
+      // expect token2's denom to NOT be registered in BridgeBank
+      let denomInBridgeBank = await state.bridgeBank.contractDenom(state.token2.address);
+      expect(denomInBridgeBank).to.be.equal("");
+
+      // add token 2 as BridgeToken
+      await state.bridgeBank.connect(owner).addExistingBridgeToken(state.token2.address);
+
+      // userOne calls forceSetBridgeTokenDenom
+      await expect(state.bridgeBank.connect(userOne)
+        .forceSetBridgeTokenDenom(state.token2.address))
```
ElliotFriedman, line 375: Let's add a test where we try to call `forceSetBridgeTokenDenom` with a non cosmos whitelisted asset, and assert that this fails.

smart-contracts/contracts/Mocks/ReentrancyToken.sol
```diff
@@ -0,0 +1,66 @@
...
+        sigData[1] = sig2;
+
+        CosmosBridge.SignatureData memory sig3;
+        sig3.signer = 0x90F79bf6EB2c4f870365E785982E1f101E93b906;
+        sig3._v = 27;
+        sig3._r = 0x48321cc08333eb832c4797a276d317f74b636fda5db8f7ba92604931fbe0f2a8;
+        sig3._s = 0x76ca07a2ec6ca237ede8b9ce4574b22bc7e4dec941978bc2fd62853ba28a8d63;
+        sigData[2] = sig3;
+
+        // doesn't revert, but user doesn't get the funds either
```
ElliotFriedman, line 59: Nice one

smart-contracts/contracts/CosmosBridge.sol
daniportugalgit: It costs 319 Gas for `symbol`, 319 Gas for `name`, and 587 Gas for `cosmosDenom`.
We can only change one of them to calldata, so I chose `cosmosDenom`.

smart-contracts/contracts/OracleStorage.sol
```diff
@@ -31,11 +31,6 @@ contract OracleStorage {
     */
     mapping(uint256 => mapping(address => bool)) private hasMadeClaim;
 
-    /**
-    * @dev {DEPRECATED}
-    */
-    mapping(uint256 => bool) private prophecyRedeemed;
```
ElliotFriedman: Was this data structure not used in the current mainnet deployment of peggy?

daniportugalgit: It wasn't.
It's not in the `develop` smart contracts, nor in the deployments folder jsons.

daniportugalgit:
@ElliotFriedman I believe this PR is done. It would be great if we could merge it into future/peggy2.
Then, if we need other functionalities, we could start a new branch and work on them separately.

ElliotFriedman (approved):

PR #1783, opened 2021-08-24
relayer send claim in order

juniuszhou:
the relayer will send the claim to sifnode according to lockBurnNonce in the ethereum events. also change how the relayer works. Before the PR, relayer listen to the events and process them. Now, relayer wake up every minute, identify the unprocessed events by compare lockBurnNonce from smart contract and sifnode.

### DIFF
--- a/cmd/ebrelayer/relayer/ethereum.go
+++ b/cmd/ebrelayer/relayer/ethereum.go
@@ -22,13 +22,17 @@ import (
 	sdk "github.com/cosmos/cosmos-sdk/types"
 	"github.com/ethereum/go-ethereum"
 	"github.com/ethereum/go-ethereum/accounts/abi"
+	"github.com/ethereum/go-ethereum/accounts/abi/bind"
 	"github.com/ethereum/go-ethereum/common"
 	ctypes "github.com/ethereum/go-ethereum/core/types"
+	"github.com/ethereum/go-ethereum/ethclient"
 	"github.com/syndtr/goleveldb/leveldb"
+	tmClient "github.com/tendermint/tendermint/rpc/client/http"
 	tmclient "github.com/tendermint/tendermint/rpc/client/http"
 	"go.uber.org/zap"
 
 	"github.com/Sifchain/sifnode/cmd/ebrelayer/contract"
+	bridgebank "github.com/Sifchain/sifnode/cmd/ebrelayer/contract/generated/bindings/bridgebank"
 	"github.com/Sifchain/sifnode/cmd/ebrelayer/txs"
 	"github.com/Sifchain/sifnode/cmd/ebrelayer/types"
 	ethbridge "github.com/Sifchain/sifnode/x/ethbridge/types"
@@ -37,8 +41,7 @@ import (
 
 const (
 	transactionInterval = 10 * time.Second
-	trailingBlocks      = 50
-	ethLevelDBKey       = "ethereumLastProcessedBlock"
+	ethereumWakeupTimer = 60
 )
 
 // EthereumSub is an Ethereum listener that can relay txs to Cosmos and Ethereum
@@ -90,7 +93,7 @@ func NewEthereumSub(
 	}
 }
 
-// Start an Ethereum chain subscription
+// Start an Ethereum event handler
 func (sub EthereumSub) Start(txFactory tx.Factory, completionEvent *sync.WaitGroup) {
 	defer completionEvent.Done()
 	time.Sleep(time.Second)
@@ -104,8 +107,16 @@ func (sub EthereumSub) Start(txFactory tx.Factory, completionEvent *sync.WaitGro
 		return
 	}
 	defer ethClient.Close()
+
 	sub.SugaredLogger.Infow("Started Ethereum websocket with provider:",
 		"Ethereum provider", sub.EthProvider)
+
+	tmClient, err := tmClient.New(sub.TmProvider, "/websocket")
+	if err != nil {
+		sub.SugaredLogger.Errorw("failed to initialize a sifchain client.",
+			errorMessageKey, err.Error())
+		return
+	}
 
 	networkID, err := ethClient.NetworkID(context.Background())
 	if err != nil {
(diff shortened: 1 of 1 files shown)


cmd/ebrelayer/main.go
```diff
@@ -365,10 +365,10 @@ func RunInitWitnessCmd(cmd *cobra.Command, args []string) error {
 func replayEthereumCmd() *cobra.Command {
 	//nolint:lll
 	replayEthereumCmd := &cobra.Command{
-		Use:     "replayEthereum [tendermintNode] [web3Provider] [bridgeRegistryContractAddress] [validatorMoniker] [fromBlock] [toBlock] [sifFromBlock] [sifEndBlock]",
+		Use:     "replayEthereum [tendermintNode] [web3Provider] [bridgeRegistryContractAddress] [validatorMoniker]",
 		Short:   "replay missed ethereum events",
-		Args:    cobra.ExactArgs(8),
-		Example: "replayEthereum tcp://localhost:26657 ws://localhost:7545/ 0x30753E4A8aad7F8597332E813735Def5dD395028 validator 100 200 100 200 --chain-id=peggy",
+		Args:    cobra.ExactArgs(4),
+		Example: "replayEthereum tcp://localhost:26657 ws://localhost:7545/ 0x30753E4A8aad7F8597332E813735Def5dD395028 validator --chain-id=peggy",
```
ElliotFriedman, line 371: Please add the --from and --symbol_translator path/to/file to this command as --from is needed and symbol translation is needed when we are moving bridged assets from ethereum to sifchain and vise versa.

ElliotFriedman (changes requested):

ElliotFriedman:
The architecture here is not something we will be moving forward with so we will do the merge of develop into future/peggy2 and then we can have a discussion of how to architect this.

ElliotFriedman:
Specifically, we should be recycling logic from the current relayer implementation to query events on ethereum, then filter out events from that data that we have already processed based off of their lock burn nonce.

cmd/ebrelayer/relayer/ethereum.go
```diff
@@ -127,177 +137,162 @@ func (sub EthereumSub) Start(txFactory tx.Factory, completionEvent *sync.WaitGro
...
+func (sub EthereumSub) CheckNonceAndProcess(txFactory tx.Factory,
+	networkID *big.Int,
+	ethClient *ethclient.Client,
+	tmClient *tmclient.HTTP,
+	bridgeBankAddress common.Address,
+	bridgeBankContractABI abi.ABI) {
+	// get current block height
+	currentBlock, err := ethClient.HeaderByNumber(context.Background(), nil)
+	if err != nil {
+		return
```
smartyalgo, line 168: It looks like we're failing silently if these calls fails. Is this the intended behaviour?

juniuszhou: log added if call failed.

cmd/ebrelayer/relayer/ethereum.go
```diff
@@ -130,17 +143,49 @@ func (sub EthereumSub) Start(txFactory tx.Factory, completionEvent *sync.WaitGro
...
 	if err != nil {
-		sub.SugaredLogger.Errorw("failed to subscribe new head.",
+		sub.SugaredLogger.Errorw("failed to get the current block from ethereum client",
 			errorMessageKey, err.Error())
 		return
 	}
-	defer subHead.Unsubscribe()
+	currentBlockHeight := currentBlock.Number
 
+	if currentBlockHeight.Cmp(big.NewInt(trailingBlocks)) <= 0 {
```
ElliotFriedman, line 182: Please add comment ```If current block is less than 50, just return.```

ElliotFriedman:
Remove all levelDB logic and solely rely on the blockchain and events to catch up.

ElliotFriedman:
We can query the topics by messages each sifchain relayer has not run by querying the topics like so https://ethereum.stackexchange.com/questions/16313/how-can-i-view-event-logs-for-an-ethereum-contract

ElliotFriedman:
We should remove all querying by block height and instead query based on the lock burn nonce in the bridge bank transactions.

cmd/ebrelayer/main.go
```diff
@@ -150,8 +150,6 @@ func RunInitRelayerCmd(cmd *cobra.Command, args []string) error {
 		return errors.Errorf("invalid [ETHEREUM_PRIVATE_KEY] environment variable")
 	}
 
-	// TODO peggy2merge
```
ElliotFriedman, line 153: If we haven't fixed, let's leave these comments in the codebase.

pandaring2you:
@Brando753 and @banshee to review this week

banshee (changes requested):

cmd/ebrelayer/relayer/cosmos.go
```diff
@@ -94,114 +88,110 @@ func (sub CosmosSub) Start(txFactory tx.Factory, completionEvent *sync.WaitGroup
...
-	data, err := sub.DB.Get([]byte(cosmosLevelDBKey), nil)
-	if err != nil {
-		log.Println("Error getting the last cosmos block from level db", err)
-		lastProcessedBlock = 0
-	} else {
-		lastProcessedBlock = new(big.Int).SetBytes(data).Int64()
-	}
-
+	// start the timer
+	t := time.NewTicker(time.Second * cosmosWakeupTimer)
```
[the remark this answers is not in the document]
juniuszhou: yes, we should consider the overlap. loop and a sleep between actions is bettwe. will remove the ticker.

cmd/ebrelayer/relayer/cosmos.go
```diff
@@ -94,114 +88,110 @@ func (sub CosmosSub) Start(txFactory tx.Factory, completionEvent *sync.WaitGroup
...
 	for {
 		select {
+		// Handle any errors
 		case <-quit:
 			log.Println("we receive the quit signal and exit")
 			return
+		case <-t.C:
+			sub.CheckNonceAndProcess(txFactory, client)
```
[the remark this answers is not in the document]
juniuszhou: will use the sleep and loop.

cmd/ebrelayer/relayer/cosmos.go
```diff
@@ -94,114 +88,108 @@ func (sub CosmosSub) Start(txFactory tx.Factory, completionEvent *sync.WaitGroup
...
-			sub.SugaredLogger.Infow("new sifchain block witnessed")
+	// get lock burn nonce and start block number from cosmos
+	globalNonce, blockNumber, err := sub.GetGlobalNonceBlockNumberFromCosmos(sub.NetworkDescriptor, sub.ValidatorName)
+	if err != nil {
+		sub.SugaredLogger.Errorw("failed to get the lock burn nonce from cosmos rpc",
+			errorMessageKey, err.Error())
+		return
+	}
+
+	ctx, cancel := context.WithTimeout(context.Background(), time.Second)
```
banshee: Waiting for one second seems likely to fail in production, even if it works in tests

cmd/ebrelayer/relayer/ethereum.go
```diff
@@ -129,243 +139,225 @@ func (sub EthereumSub) Start(txFactory tx.Factory, completionEvent *sync.WaitGro
...
-	return false
 }
 
 // Replay the missed events
-func (sub EthereumSub) Replay(txFactory tx.Factory, fromBlock int64, toBlock int64, cosmosFromBlock int64, cosmosToBlock int64, symbolTranslator *symbol_translator.SymbolTranslator) {
-	log.Printf("ethereum replay for %d block to %d block\n", fromBlock, toBlock)
-
-	bridgeClaims := sub.getAllClaims(cosmosFromBlock, cosmosToBlock)
-	log.Printf("found out %d bridgeClaims\n", len(bridgeClaims))
+func (sub EthereumSub) Replay(txFactory tx.Factory, symbolTranslator *symbol_translator.SymbolTranslator) {
```
banshee, line 314: Do we still need Replay?

juniuszhou: I think replay is unnecessary, will create other pr to remove it completely.

PR #1787, opened 2021-08-25
hotfix for decimal delta

juniuszhou:
the hotfix use the token delta map to fix the decimal mismatch issue between ethereum and sifnode for all IBC tokens.


ElliotFriedman:
We need to add this same logic to the replay scripts as well so things don't get borked if we try to replay.

ElliotFriedman:
Closing as we chose to go with another solution.

PR #1790, opened 2021-08-25
Rpc lock burn nonce

juniuszhou:
sifnode provide a rpc to relayer querying its lock burn nonce in the ethbridge keeper.


ElliotFriedman (approved):
LGTM!

PR #1816, opened 2021-08-31
Merge develop smart-contracts/deployments into peggy2

banshee:
We need to pick up sifchain-1, so might as well merge in the whole directory.


ElliotFriedman (approved):
LGTM

PR #1817, opened 2021-08-31
Update Readme, Add Mainnet JSON for Whitelisting

ElliotFriedman:


smart-contracts/src/ibcMatchingTokens.ts
```diff
@@ -1,78 +1,111 @@
...
 }
 
 async function attachIbcToken(bridgeBank: BridgeBank, token: IbcToken) {
-    return await bridgeBank.addExistingBridgeToken(token.address)
+  return await bridgeBank.addExistingBridgeToken(token.address);
 }
 
-export async function processTokenData(bridgeBank: BridgeBank, filename: string, container: DependencyContainer) {
-    const fileContents = fs.readFileSync(filename, {encoding: "utf8"})
+export async function processTokenData(
```
iverc, line 15: Do we need some tests around process token / build token logic?

iverc (approved):

smart-contracts/src/ibcMatchingTokens.ts
banshee: Yes, but this is a PR for work-in-progress, so that's fine.

PR #1820, opened 2021-08-31
osmosis test net

timlind:


scripts/ibc/tokenregistration/sifchain-testnet-1/crypto-org.json
```diff
@@ -2,7 +2,7 @@
   "entries": [
     {
       "is_whitelisted": true,
-      "decimals": "6",
+      "decimals": "8",
```
timlind, line 5: There's no IBC hash but decimals should be 8 as is in generate script

https://www.notion.so/sifchain/IBC-Tokens-78f3c309612c41ea970c06eef32ee045

PR #1823, opened 2021-09-01
Add Current & Upcoming Reward Program Types to DistributionType Enum

mccallofthewild:
Summary:
* Updates `DistributionType` with types for upcoming reward programs
* Adds `DISTRIBUTION_TYPE_LIQUIDITY_MINING_IBC_V1`, `DISTRIBUTION_TYPE_LIQUIDITY_MINING_AKASH_V1`, and `DISTRIBUTION_TYPE_LIQUIDITY_MINING_PEGGY_V1`

Optimally, we may want to shift toward an individual `RewardProgramType` that points to specific reward programs rather than categorizing them, as `DISTRIBUTION_TYPE_LIQUIDITY_MINING` does.

If there is not currently bandwidth to make this happen, custom distribution types as implemented are a fairly straightforward solution.

Additional steps may be required to properly implement this addition, but the implemented changes provide an outline of what updates are needed.


mccallofthewild:
Closing due to inactivity.

PR #1850, opened 2021-09-09
Add whitelist script using new hardhat functions

banshee:
We need some whitelist tools for the new whitelisting procedures.


daniportugalgit (approved):
LGTM!

PR #1853, opened 2021-09-10
Feature | New Liquidity Provider Data Endpoint

iverc:


x/clp/keeper/grpc_query.go
```diff
@@ -84,6 +85,52 @@ func (k Querier) GetLiquidityProvider(c context.Context, req *types.LiquidityPro
...
+		return nil, status.Error(codes.InvalidArgument, fmt.Sprintf("page size greater than max %d", MaxPageLimit))
+	}
+
+	ctx := sdk.UnwrapSDKContext(c)
+	addr, err := sdk.AccAddressFromBech32(req.LpAddress)
+	if err != nil {
+		return nil, err
+	}
+
+	assetList, _, err := k.Keeper.GetAssetsForLiquidityProviderPaginated(ctx, addr, &query.PageRequest{Limit: MaxPageLimit})
```
Jedi2002, line 109: Why do we need to send the MaxPageLimit always ? 

Cant we , change it to 
`if req.Pagination.Limit > MaxPageLimit {
     req.Pagination.Limit = MaxPageLimit
}`

and then use req.Pagination.Limit here

x/clp/keeper/keeper_test.go
```diff
@@ -149,3 +151,42 @@ func TestKeeper_GetModuleAccount(t *testing.T) {
...
 }
+
+func TestKeeper_GetLiquidityProviderData(t *testing.T) {
+	ctx, app := test.CreateTestAppClp(false)
+	clpKeeper := app.ClpKeeper
+	tokens := []string{"cada", "cbch", "cbnb", "cbtc", "ceos", "ceth", "ctrx", "cusdt"}
+	pools, lpList := test.GeneratePoolsAndLPs(clpKeeper, ctx, tokens)
+	lpaddr, err := sdk.AccAddressFromBech32(lpList[0].LiquidityProviderAddress)
+	require.NoError(t, err)
+	assetList, _, err := clpKeeper.GetAssetsForLiquidityProviderPaginated(ctx, lpaddr, &query.PageRequest{Limit: math.MaxUint64})
```
Jedi2002, line 162: might be a good idea to iterate over the pages in the test to verify expected behaviour . 
Something like this 
https://github.com/Sifchain/sifnode/blob/0370b7b86b7a6f169731fbbc692e3a0b63cf2c2c/x/dispensation/keeper/distributionRecords_test.go#L132-L144

x/clp/keeper/grpc_query.go
iverc: 👍

x/clp/keeper/keeper_test.go
iverc: 👍

timlind (changes requested):

proto/sifnode/clp/v1/types.proto
```diff
@@ -38,3 +38,9 @@ message LiquidityProvider {
 }
 
 message WhiteList { repeated string validator_list = 1; }
+
+message LiquidityProviderData {
+  sifnode.clp.v1.LiquidityProvider liquidity_provider = 1;
+  string native_asset_balance = 2;
+  string external_asset_balance = 3;
```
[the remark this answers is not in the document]
iverc: @timlind LiquidityProvider object has Asset object inside which has Symbol of that asset.

timlind (approved):

PR #1870, opened 2021-09-14
witness lock burn nonce support

juniuszhou:
Sifnode record the lock burn nonce for witness node and require the witness node send the transaction for sign prophecy in order. In the eblayer side, the level db will be removed.

### DIFF
--- a/x/oracle/keeper/keeper.go
+++ b/x/oracle/keeper/keeper.go
@@ -203,6 +203,17 @@ func (k Keeper) ProcessSignProphecy(ctx sdk.Context, networkDescriptor types.Net
 		return err
 	}
 
+	prophecyInfo, ok := k.GetProphecyInfo(ctx, prophecyID)
+	if !ok {
+		return errors.New("prophecy info not available in keeper")
+	}
+
+	// check the order of witness lock burn nonce
+	lastLockBurnNonce := k.GetWitnessLockBurnNonce(ctx, networkDescriptor, valAddr)
+	if lastLockBurnNonce != 0 && lastLockBurnNonce+1 != prophecyInfo.GlobalNonce {
+		return errors.New("witness node not send the signature in order")
+	}
+
 	oldStatus := prophecy.Status
 
 	newStatus, err := k.AppendValidatorToProphecy(ctx, networkDescriptor, prophecyID, valAddr)
@@ -215,13 +226,11 @@ func (k Keeper) ProcessSignProphecy(ctx sdk.Context, networkDescriptor types.Net
 		return err
 	}
 
+	// update witness's lock burn nonce
+	k.SetWitnessLockBurnNonce(ctx, networkDescriptor, valAddr, prophecyInfo.GlobalNonce)
+
 	// emit the event when status from pending to success
 	if oldStatus == types.StatusText_STATUS_TEXT_PENDING && newStatus == types.StatusText_STATUS_TEXT_SUCCESS {
-		prophecyInfo, ok := k.GetProphecyInfo(ctx, prophecyID)
-		if !ok {
-			return errors.New("prophecy info not available in keeper")
-		}
-
 		ctx.EventManager().EmitEvents(sdk.Events{
 
 			sdk.NewEvent(
--- a/x/oracle/keeper/witnessLockBurnNonce.go
+++ b/x/oracle/keeper/witnessLockBurnNonce.go
@@ -1,0 +1,43 @@
+package keeper
+
+import (
+	"bytes"
+	"encoding/binary"
+
+	"github.com/Sifchain/sifnode/x/oracle/types"
+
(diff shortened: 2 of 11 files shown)


x/oracle/keeper/witnessLockBurnNonce.go
```diff
@@ -0,0 +1,43 @@
...
+	}
+
+	bz := store.Get(key)
+	return binary.BigEndian.Uint64(bz)
+}
+
+// getWitnessLockBurnNoncePrefix return storage prefix
+func (k Keeper) getWitnessLockBurnNoncePrefix(networkDescriptor types.NetworkDescriptor, valAccount sdk.ValAddress) []byte {
+	bytebuf := bytes.NewBuffer([]byte{})
+	_ = binary.Write(bytebuf, binary.BigEndian, networkDescriptor)
```
banshee, line 40: Don't we have to check if Write returns an error?

banshee (approved):

x/oracle/keeper/keeper.go
```diff
@@ -215,13 +226,12 @@ func (k Keeper) ProcessSignProphecy(ctx sdk.Context, networkDescriptor types.Net
 		return err
 	}
 
+	// update witness's lock burn nonce
+	k.SetWitnessLockBurnNonce(ctx, networkDescriptor, valAddr, prophecyInfo.GlobalNonce)
+
 	// emit the event when status from pending to success
+	// TODO - what happens here if we don't match StatusText_STATUS_TEXT_PENDING and StatusText_STATUS_TEXT_SUCCESS?
```
iverc, line 233: @juniuszhou @banshee I think we should make it crystal clear what happens in every case:
- old = unspecified, new = pending (+fail/success noops)
- old = pending, new = failure
- old = pending, new = success

x/oracle/keeper/witnessLockBurnNonce.go
iverc: @juniuszhou yeah, let's handle Write error case

x/oracle/keeper/witnessLockBurnNonce_test.go
```diff
@@ -0,0 +1,38 @@
...
+	testAddress       = "cosmosvaloper1mnfm9c7cdgqnkk66sganp78m0ydmcr4pn7fqfk"
+	testLockBurnNonce = uint64(10)
+	testInitNonce     = uint64(0)
+)
+
+func TestSetWitnessLockBurnNonce(t *testing.T) {
+	var ctx, _, _, _, keeper, _, _, _ = test.CreateTestKeepers(t, 0.7, []int64{3, 3}, "")
+	testCosmosAddress, err := sdk.ValAddressFromBech32(testAddress)
+	require.NoError(t, err)
+
```
iverc, line 24: can we add lockBurnNonce check before we set it?

x/oracle/keeper/witnessLockBurnNonce_test.go
```diff
@@ -0,0 +1,38 @@
...
+	lockBurnNonce := keeper.GetWitnessLockBurnNonce(ctx, testNetwork, testCosmosAddress)
+	assert.Equal(t, lockBurnNonce, testLockBurnNonce)
+}
+
+func TestGetWitnessLockBurnNonce(t *testing.T) {
+	var ctx, _, _, _, keeper, _, _, _ = test.CreateTestKeepers(t, 0.7, []int64{3, 3}, "")
+	testCosmosAddress, err := sdk.ValAddressFromBech32(testAddress)
+	require.NoError(t, err)
+
+	lockBurnNonce := keeper.GetWitnessLockBurnNonce(ctx, testNetwork, testCosmosAddress)
```
iverc, line 36: another test case for `getWitnessLockBurnNoncePrefix` could be useful

x/oracle/keeper/keeper.go
juniuszhou: As the comment explains, the EventTypeProphecyCompleted just emitted once, and only happened if the status changed from pending to success. will add comment in code.
old = unspecified, new = pending  the prophecy just created, not emit the event
old = unspecified, new = success no such path
old = pending, new = failure no failed status for prophecy anymore
old = pending, new = success the only case we will emit the event
old = success, new = success not emit the same event twice

x/oracle/keeper/witnessLockBurnNonce_test.go
juniuszhou: will add the case.

x/oracle/keeper/witnessLockBurnNonce_test.go
juniuszhou: will add the test case.

x/oracle/keeper/witnessLockBurnNonce.go
juniuszhou: will add the error handler.

PR #1874, opened 2021-09-15
Feature/mint testnet

AustinoBombino:


version
```diff
@@ -1 +1 @@
-0.9.5
\ No newline at end of file
+0.9.6-rc.2
```
intl-man, line 1: Not sure we should be sending out a release candidate to TestNet? Should we not release it as 0.9.6 (and 0.9.7 for decimal precision?).

app/setup_handlers.go
```diff
@@ -5,13 +5,46 @@ import (
 	tokenregistrytypes "github.com/Sifchain/sifnode/x/tokenregistry/types"
 	storetypes "github.com/cosmos/cosmos-sdk/store/types"
 	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/cosmos/cosmos-sdk/x/mint"
+	minttypes "github.com/cosmos/cosmos-sdk/x/mint/types"
 	"github.com/cosmos/cosmos-sdk/x/upgrade/types"
 )
 
 const upgradeNameV095 = "0.9.5"
```
intl-man, line 13: Can we get rid of this?

intl-man (changes requested):

marshsif (dismissed):
looks Great!

intl-man (dismissed):

app/setup_handlers.go
```diff
@@ -1,34 +1,38 @@
...
-	tokenregistrytypes "github.com/Sifchain/sifnode/x/tokenregistry/types"
 	storetypes "github.com/cosmos/cosmos-sdk/store/types"
 	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/cosmos/cosmos-sdk/x/mint"
+	minttypes "github.com/cosmos/cosmos-sdk/x/mint/types"
 	"github.com/cosmos/cosmos-sdk/x/upgrade/types"
 )
 
-const upgradeNameV095 = "0.9.5"
+const upgradeNameV096 = "0.9.6"
```
marshsif, line 11: we don't need to do this now, but shouldn't the const name be more generic, without the version variable in it? (ie,. upgradeNameVersion)

app/setup_handlers.go
```diff
@@ -1,34 +1,38 @@
...
+	SetupHandlersForMint(app)
 }
 
-func SetupHandlersForV095(app *SifchainApp) {
-	app.UpgradeKeeper.SetUpgradeHandler("0.9.5", func(ctx sdk.Context, plan types.Plan) {
-		app.Logger().Info("Running upgrade handler for " + upgradeNameV095 + " with new store " + tokenregistrytypes.StoreKey)
-		// Install initial token registry entries for non-ibc tokens.
-		tokenregistrymigrations.Init(ctx, app.TokenRegistryKeeper)
+func SetupHandlersForMint(app *SifchainApp) {
+	app.UpgradeKeeper.SetUpgradeHandler("0.9.6", func(ctx sdk.Context, plan types.Plan) {
```
marshsif, line 18: we don't need this now, but shouldn't the first argument refer to the const rather than a hard-coded version string?

app/setup_handlers.go
intl-man: Good idea @marshsif - we should maybe think about a better way to handle migrations etc....

app/setup_handlers.go
intl-man: @marshsif Yes, it should.

marshsif (approved):

PR #1877, opened 2021-09-15
Develop whitelist gas prices

daniportugalgit:
- Explicitly sets the gas price for the bulk_set_whitelist;
- Changes the way we set .env variables (works well for the bulk_set_whitelist script)


ElliotFriedman (approved):
Tested against forked mainnet and it works!

PR #1880, opened 2021-09-16
Delete Whitelist.md

musnit:
This is deprecated right? Let's remove it if so.


daniportugalgit (approved):
Yes, Whitelist.md is deprecated.
PR approved.

PR #1882, opened 2021-09-16
Feature/enable multiple coins

Jedi2002:


x/dispensation/types/keys.go
```diff
@@ -23,10 +23,11 @@ const (
 	QuerierRoute       = ModuleName
 	DefaultParamspace  = ModuleName
 	MaxRecordsPerBlock = 10
-	TokenSupported     = "rowan"
 )
 
 var (
+	// TODO : Refactor to genesis state
```
iverc, line 29: @Jedi2002 should we address this TODO here?

Jedi2002: Good point , We can address this as a separate PR though. 
@marshsif  can you comment wether this feature , to support coins other than rowan is required or not .

PR #1884, opened 2021-09-16
Update Parser to Send Rowan to Sifchain with nonce 4,000,000,000

ElliotFriedman:


cmd/ebrelayer/txs/parser.go
```diff
@@ -61,14 +62,14 @@ func EthereumEventToEthBridgeClaim(valAddr sdk.ValAddress, event types.EthereumE
 	amount := sdk.NewIntFromBigInt(event.Value)
 
 	// Nonce type casting (*big.Int -> int)
-	nonce := int(event.Nonce.Int64())
+	nonce := int(4000000000)
```
banshee, line 65: We should have a comment here explaining why we're using a large int for a nonce

cmd/ebrelayer/txs/parser.go
```diff
@@ -61,14 +62,14 @@ func EthereumEventToEthBridgeClaim(valAddr sdk.ValAddress, event types.EthereumE
...
 
 	// Package the information in a unique EthBridgeClaim
 	witnessClaim.EthereumChainId = int64(chainID)
 	witnessClaim.BridgeContractAddress = bridgeContractAddress.String()
 	witnessClaim.Nonce = int64(nonce)
 	witnessClaim.TokenContractAddress = tokenContractAddress.String()
-	witnessClaim.Symbol = symbol
+	witnessClaim.Symbol = "rowan"
```
banshee, line 72: Why is this hardcoded to rowan now?

banshee (changes requested):

PR #1886, opened 2021-09-17
[ChainOps] Latest merge to master

intl-man:


app/app.go
```diff
@@ -395,9 +409,10 @@ func NewSifApp(
 	// CanWithdrawInvariant invariant.
 	app.mm.SetOrderBeginBlockers(
 		capabilitytypes.ModuleName,
+		upgradetypes.ModuleName,
```
Jedi2002, line 412: Any specific reason why we need change the order ? As in move upgrade up ? 
I cant foresee a problem with it , but the ordering has caused issues during upgrade before

marshsif (approved):
This looks good to me.

iverc (approved):

app/app.go
iverc: right, better to keep the order if it was working correctly

iverc (approved):

PR #1893, opened 2021-09-17
Remove the whitelist being needed to lock tokens in the bridgebank

daniportugalgit:
"Remove the whitelist being needed to lock tokens in the bridgebank.
Instead, add a check that whenever a token is locked, it is not in the cosmoswhitelist, and if it is, revert."


ElliotFriedman (approved):
LGTM!

PR #1894, opened 2021-09-17
tokenDetails: only allow symbols without spaces and/or special characters

daniportugalgit:


daniportugalgit:
Will complete this task in another PR, joining whitelist scripts together

PR #1897, opened 2021-09-18
Fix cli tests

iverc:


iverc:
Raised new PR based on `develop` #1898

PR #1903, opened 2021-09-21
Reduce ibc tx gas usage

iverc:


kingpinXD (commented):
LGTM

proto/sifnode/tokenregistry/v1/types.proto
```diff
@@ -20,39 +20,37 @@ enum Permission {
...
+  name of denomination unit of this token that should appear on counterparty chain when this unit is exported.
+  If empty, the denom is exported as is.
+  Generally this will only be used to map a high precision (unit_denom) to a lower precision,
+  to overcome the current uint64 limit on the packet level.
+  i.e rowan -> microrowan
+  i.e microrowan -> microrowan
+*/
 
 message RegistryEntry {
-  bool is_whitelisted = 1;
```
timlind, line 25: A change like this is breaking and needs to be reverted: the field number always needs to stay the same. 

Also why is IsWhitelisted being removed?

When calling GetDenom() how does the caller know whether the token is on the whitelist or not?

x/clp/types/expected_keepers.go
```diff
@@ -33,7 +33,7 @@ type AuthKeeper interface {
 }
 
 type TokenRegistryKeeper interface {
-	GetDenom(ctx sdk.Context, denom string) tokenregistryTypes.RegistryEntry
-	IsDenomWhitelisted(ctx sdk.Context, denom string) bool
-	CheckDenomPermissions(ctx sdk.Context, denom string, permissions []tokenregistryTypes.Permission) bool
+	GetDenom(registry tokenregistryTypes.Registry, denom string) *tokenregistryTypes.RegistryEntry
```
timlind, line 36: +1 on returning pointer, IsWhitelisted becomes less necessary in this case.

x/clp/types/expected_keepers.go
```diff
@@ -33,7 +33,7 @@ type AuthKeeper interface {
...
+	CheckDenomPermissions(entry *tokenregistryTypes.RegistryEntry, permissions []tokenregistryTypes.Permission) bool
+	GetDenomWhitelist(ctx sdk.Context) tokenregistryTypes.Registry
```
timlind, line 38: Should be called Registry not Whitelist

proto/sifnode/tokenregistry/v1/types.proto
```diff
@@ -20,39 +20,37 @@ enum Permission {
...
-  int64 decimals = 2;
-  string denom = 3;
-  string base_denom = 4;
-  string path = 5;
+  uint32 decimals = 1;
+  string denom = 2;
+  string base_denom = 3;
+  string unit_denom = 4;
+  string ibc_transfer_port = 5;
```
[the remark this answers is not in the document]
timlind: This field doesn't represent the port, it represents the `ibc_trace_path`.

The transfer port will always be `transfer` so there's no need for an `ibc_transfer` field.

proto/sifnode/tokenregistry/v1/types.proto
```diff
@@ -20,39 +20,37 @@ enum Permission {
...
   string ibc_channel_id = 6;
   string ibc_counterparty_channel_id = 7;
-  string display_name = 8;
```
timlind, line 32: Why is display_name being removed?

proto/sifnode/tokenregistry/v1/types.proto
```diff
@@ -20,39 +20,37 @@ enum Permission {
...
-  string display_symbol = 9;
```
timlind, line 33: Why is display_symbol being removed?

proto/sifnode/tokenregistry/v1/types.proto
```diff
@@ -20,39 +20,37 @@ enum Permission {
...
-  string network = 10;
```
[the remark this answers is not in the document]
timlind: Why is network being removed?

proto/sifnode/tokenregistry/v1/types.proto
```diff
@@ -20,39 +20,37 @@ enum Permission {
...
-  string address = 11;
-  string external_symbol = 12;
```
timlind, line 36: Why is external_symbol being removed?

proto/sifnode/tokenregistry/v1/types.proto
iverc: @timlind we use `reserved` to make sure that removed field numbers can't be reused.
I believe that is_whitelisted field is redundant, as long as record is in the registry -  it is whitelisted. We return pointer and use nil check instead.

x/clp/types/expected_keepers.go
iverc: @timlind isWhitelisted is replaced with nil check now

proto/sifnode/tokenregistry/v1/types.proto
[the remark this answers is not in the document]
iverc: @timlind All removed fields are empty for all records. Only a couple of entries have display_name/symbol populated and also unused. In general I think that additional info like that is being used mainly for display purposes and should not be stored in the registry. I'd only store fields that are being used for the conv logic (denom, unit denom, decimals, counterparty_chain_id, and both channel ids) to reduce memory & gas usage, and improve performance

x/ibctransfer/onrecv_packet.go
```diff
@@ -3,36 +3,67 @@ package ibctransfer
...
+			Events: ctx.EventManager().Events().ToABCIEvents(),
+		}, acknowledgement.GetBytes(), nil
+	}
+	// Get the denom that will be minted by sdk transfer module,
+	// so that it can be converted to the denom it should be stored as.
+	// For a native token that has been returned, this will just be a base_denom,
+	// which will be on the whitelist.
+	mintedDenom := helpers.GetMintedDenomFromPacket(packet, data)
+	registry := whitelistKeeper.GetDenomWhitelist(ctx)
+	mintedDenomEntry := whitelistKeeper.GetDenom(registry, mintedDenom)
```
timlind, line 39: GetDenom needs to return an err if value returned could be nil

proto/sifnode/tokenregistry/v1/types.proto
```diff
@@ -35,24 +56,9 @@ message RegistryEntry {
   string address = 11;
   string external_symbol = 12;
   string transfer_limit = 13;
+  uint32 decimals = 14;
```
timlind, line 59: Please leave decimal field as is.

x/ibctransfer/onrecv_packet.go
```diff
@@ -3,36 +3,67 @@ package ibctransfer
...
 	transfertypes "github.com/cosmos/cosmos-sdk/x/ibc/applications/transfer/types"
 	channeltypes "github.com/cosmos/cosmos-sdk/x/ibc/core/04-channel/types"
-
-	sctransfertypes "github.com/Sifchain/sifnode/x/ibctransfer/types"
-	tokenregistrytypes "github.com/Sifchain/sifnode/x/tokenregistry/types"
 )
 
-// OnRecvPacketMaybeConvert will receive a transfer (after whitelisting is checked),
-// and if the receive was successful,
-// it will be converted into the unit_denom of the denom minted by the IBC transfer module.
```
timlind, line 17: Doc comments should stay or be updated.

x/ibctransfer/onrecv_timeout.go
```diff
@@ -47,18 +48,21 @@ func OnTimeoutMaybeConvert(
...
 	)
-	// if needs conversion, convert and send
-	if ShouldConvertRefundCoins(ctx, whitelistKeeper, packet, data) {
-		incomingCoins, finalCoins := GetConvForRefundCoins(ctx, whitelistKeeper, packet, data)
-		err := ExecConvForRefundCoins(ctx, incomingCoins, finalCoins, bankKeeper, packet, data)
-		if err != nil {
-			return nil, err
+	denom := data.Denom
+	registry := whitelistKeeper.GetRegistry(ctx)
+	denomEntry, err := whitelistKeeper.GetEntry(registry, denom)
```
timlind, line 53: Avoid err == nil checks.

x/ibctransfer/onrecv_timeout_test.go
```diff
@@ -74,33 +66,27 @@ func TestOnTimeoutPacketConvert_Source(t *testing.T) {
...
-
-			err = keeper.PrepareToSendConvertedCoins(sdk.WrapSDKContext(ctx), tt.args.msg, tokenDeduction, tokensConverted, app.BankKeeper)
+			err = helpers.PrepareToSendConvertedCoins(sdk.WrapSDKContext(ctx), tt.args.msg, tokenDeduction, tokensConverted, app.BankKeeper)
 			require.NoError(t, err)
 			require.Equal(t, tokensConverted.String(), app.BankKeeper.GetBalance(ctx, sender, tokensConverted.Denom).String())
 			require.Equal(t, tt.args.msg.Token.Sub(tokenDeduction).String(), app.BankKeeper.GetBalance(ctx, sender, tt.args.msg.Token.Denom).String())
-
 			// Simulate send with SDK stub.
-			sdkSentDenom, err := testhelpers.SendStub(ctx, app.TransferKeeper, app.BankKeeper, tokensConverted, sender, tt.args.msg.SourcePort, tt.args.msg.SourceChannel)
+			sdkSentDenom, _ := testhelpers.SendStub(ctx, app.TransferKeeper, app.BankKeeper, tokensConverted, sender, tt.args.msg.SourcePort, tt.args.msg.SourceChannel)
```
timlind, line 87: assert error value

x/ibctransfer/onrecv_timeout.go
```diff
@@ -3,6 +3,7 @@ package ibctransfer
 import (
 	"fmt"
 
+	"github.com/Sifchain/sifnode/x/ibctransfer/helpers"
```
timlind, line 6: Is this filename change necessary and accurate? The original name was chosen to be less ambiguous.

x/tokenregistry/client/cli/query.go
```diff
@@ -54,21 +54,15 @@ func GetCmdQueryEntries() *cobra.Command {
...
+	var flagIbcCounterpartyChannelID = "token_ibc_counterparty_channel_id"
+	var flagIbcCounterpartyChainID = "token_ibc_counterparty_chain_id"
+	var flagIbcCounterpartyDenom = "token_ibc_counterparty_denom"
 	var flagDecimals = "token_decimals"
-	var flagDisplayName = "token_display_name"
-	var flagDisplaySymbol = "token_display_symbol"
-	var flagExternalSymbol = "token_external_symbol"
-	var flagTransferLimit = "token_transfer_limit"
-	var flagNetwork = "token_network"
-	var flagAddress = "token_address"
```
timlind, line 71: These fields should remain.

x/tokenregistry/client/cli/query.go
```diff
@@ -168,9 +138,9 @@ func GetCmdGenerateEntry() *cobra.Command {
 			// generate denom if path is also provided.
 			// override the IBC generation with --denom if specified explicitly.
 			// otherwise fallback to base_denom
-			if ibcChannelId != "" {
+			if ibcChannelID != "" {
```
timlind, line 141: This should proabably change to if ibcPath != "" and use the path instead of ibcChannelID.

If ibcChannelID is specified alone without ibcPath, then it should be used to set path and denom = ibc hash.

We should also introduce validation checks on RegistryEntry that are run here.

x/ibctransfer/keeper/msg_server.go
```diff
@@ -33,29 +33,34 @@ var _ sdktransfertypes.MsgServer = msgServer{}
...
+	}
+	// check if registry entry has an IBC counterparty conversion to process
 	if registryEntry.IbcCounterpartyDenom != "" && registryEntry.IbcCounterpartyDenom != registryEntry.Denom {
-		sendAsRegistryEntry := srv.tokenRegistryKeeper.GetDenom(sdk.UnwrapSDKContext(goCtx), registryEntry.IbcCounterpartyDenom)
-		if sendAsRegistryEntry.Decimals != 0 && registryEntry.Decimals > sendAsRegistryEntry.Decimals {
-			token, tokenConversion := ConvertCoinsForTransfer(goCtx, msg, registryEntry, sendAsRegistryEntry)
-			if token.Amount.Equal(sdk.NewInt(0)) || tokenConversion.Amount.Equal(sdk.NewInt(0)) {
+		sendAsRegistryEntry, err := srv.tokenRegistryKeeper.GetEntry(registry, registryEntry.IbcCounterpartyDenom)
+		if err != nil {
+			fmt.Println(err.Error())
```
timlind, line 54: Use ctx.Logger() to log

x/tokenregistry/client/cli/query.go
```diff
@@ -162,18 +162,15 @@ func GetCmdGenerateEntry() *cobra.Command {
...
-			var path string
 			var denom string
 			// base_denom is required.
 			// generate denom if path is also provided.
 			// override the IBC generation with --denom if specified explicitly.
 			// otherwise fallback to base_denom
-			if ibcChannelId != "" {
-				// normalise path slashes before generating hash (do this in MsgRegister.ValidateBasic as well)
-				path = "transfer/" + ibcChannelId
+			if ibcPath != "" {
```
timlind, line 170: Right now, the scripts expect to use --token_channel_id, so we need to keep support for that, if ibc_path isn't specified (it should then just be set as it previously was).

PR #1914, opened 2021-09-23
Future/devenv rebased

Brando753:
Initial Phase of DevEnv along with some changes to peggy2 as we continue to try and get Ceth transfers working


smart-contracts/contracts/BridgeBank/IbcToken.sol
```diff
@@ -0,0 +1,62 @@
+// SPDX-License-Identifier: Apache-2.0
```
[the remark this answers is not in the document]
daniportugalgit: @smartyalgo Correct. Once this PR gets merged in, I'll open a new PR to deal with IbcToken/BridgeToken.

Just for some context: we created this temporary IbcToken so that Michael and Brandon could run tests quickly. In `future/peggy2` the IbcToken doesn't exist anymore (it's called BridgeToken).

smart-contracts/contracts/CosmosBridge.sol
```diff
@@ -41,7 +41,7 @@ contract CosmosBridge is CosmosBridgeStorage, Oracle {
      */
     event LogNewBridgeTokenCreated(
         uint8 decimals,
-        uint256 indexed sourcechainId,
+        int32 indexed sourcechainId,
```
daniportugalgit, line 44:
```suggestion
        int32 indexed _networkDescriptor,
```

test/integration/src/py/test_utilities.py
```diff
@@ -215,20 +230,22 @@ def get_password(network_definition_file_json):
...
-         f"--ethereum_address {transfer_request.ethereum_address}",
-         f"--json_path {transfer_request.solidity_json_path}",
-         private_element,
-         symbol_element,
-         network_element]
-    )
-    result = run_yarn_command(command_line)
-    return int(result["balanceWei"])
+    w3 = Web3(Web3.WebsocketProvider(ethereum_websocket))
+    return w3.eth.get_balance()
```
banshee, line 234: This probably won't work - the original method gets the balance of the account defined in transfer_request.ethereum_symbol, and the new one gets the account balance of the first key that w3.eth knows about (I think?)

banshee (approved):

test/integration/src/py/test_utilities.py
Brando753: This was one of those things we need to refactor, it gets the balance of the first key as you stated rather then a specified key. Does seem to work for that key though.

PR #1917, opened 2021-09-24
Peggy2 code review comments; don't merge this (it's a draft)

ElliotFriedman:


cmd/ebrelayer/relayer/cosmosReplay.go
```diff
@@ -0,0 +1,201 @@
...
+		log.Printf("failed to get the account address, error as %s\n", err)
+		return
+	}
+
+	ProphecyClaims := sub.getAllSignSigature(client, accAddr, fromBlock, toBlock)
+
+	sub.ReplayCosmosBurnLockWithBlocks(txFactory, client, accAddr, ProphecyClaims, fromBlock, toBlock)
+}
+
+// ReplayCosmosBurnLockWithBlocks replay the missed burn lock events from cosmos
```
ElliotFriedman, line 53: This comment needs to actually describe what is going on in this function as the function itself is confusing.

cmd/ebrelayer/relayer/prophecyHandler.go
```diff
@@ -0,0 +1,193 @@
...
+	lastSubmittedNonce, err := GetLastNonceSubmitted(ethClient, cosmosBridgeAddress, sub.SugaredLogger)
+	if err != nil {
+		sub.SugaredLogger.Errorw("failed to get the last submitted nonce.",
+			errorMessageKey, err.Error())
+		return
+	}
+
+	prophecyInfoArray := GetAllProphciesCompleted(sub.TmProvider, sub.NetworkDescriptor, lastSubmittedNonce.Uint64()+1)
+
+	batches := (len(prophecyInfoArray) + 4) / 5
```
ElliotFriedman, line 94: We need to be clearer on why these numbers are needed. We should use a modulo operator here instead of just add 4.

Is this code supposed to round up @juniuszhou?

cmd/ebrelayer/txs/relayToEthereum.go
```diff
@@ -11,93 +11,29 @@ import (
...
-	ctypes "github.com/ethereum/go-ethereum/core/types"
+	ethereumtypes "github.com/ethereum/go-ethereum/core/types"
 	"github.com/ethereum/go-ethereum/ethclient"
 	"go.uber.org/zap"
 )
 
 const (
 	// GasLimit the gas limit in Gwei used for transactions sent with TransactOpts
-	GasLimit = uint64(1800000)
+	GasLimit = uint64(2000000)
```
ElliotFriedman, line 24: Tx batch size should be limited to 1 in initial release of peggy2.

cmd/ebrelayer/relayer/prophecyHandler.go
```diff
@@ -0,0 +1,193 @@
...
+		sub.SugaredLogger.Errorw("failed to get cosmosBridge instance.",
+			errorMessageKey, err.Error())
+		return false
+	}
+
+	maxRetries := 5
+	i := 0
+
+	for i < maxRetries {
+		err = txs.RelayBatchProphecyCompletedToEthereum(
```
ElliotFriedman, line 142: We should only send 1 tx across the bridge at a time to start, we can move to batching in the relayer later.

Instead of calling RelayBatch, just call `RelayProphecyCompletedToEthereum` instead.

This means that we can remove the logic to turn data into batches earlier on in this function.

cmd/ebrelayer/relayer/prophecyHandler.go
```diff
@@ -0,0 +1,198 @@
...
+	lastSubmittedNonce, err := GetLastNonceSubmitted(ethClient, cosmosBridgeAddress, sub.SugaredLogger)
+	if err != nil {
+		sub.SugaredLogger.Errorw("failed to get the last submitted nonce.",
+			errorMessageKey, err.Error())
+		return
+	}
+
+	prophecyInfoArray := GetAllProphciesCompleted(sub.TmProvider, sub.NetworkDescriptor, lastSubmittedNonce.Uint64()+1)
+
+	batches := (len(prophecyInfoArray) + 4) / 5
```
ElliotFriedman, line 94: Remove all batching logic for now and send tx's 1 by 1.

cmd/ebrelayer/replay.go
```diff
@@ -78,94 +79,164 @@ func RunReplayEthereumCmd(cmd *cobra.Command, args []string) error {
...
 
-// RunReplayCosmosCmd executes initRelayerCmd
-func RunReplayCosmosCmd(cmd *cobra.Command, args []string) error {
-	// Validate and parse arguments
-	if len(strings.Trim(args[0], "")) == 0 {
-		return errors.Errorf("invalid [tendermint-node]: %s", args[0])
-	}
-	tendermintNode := args[0]
+// RunReplayCosmosBurnLockCmd replay missed burn lock events from cosmos
+func RunReplayCosmosBurnLockCmd(cmd *cobra.Command, args []string) error {
```
ElliotFriedman, line 83: Let's use cobra and flags instead of grabbing positional arguments from the CLI.

cmd/ebrelayer/replay.go
```diff
@@ -78,94 +79,164 @@ func RunReplayEthereumCmd(cmd *cobra.Command, args []string) error {
...
+		validatorMoniker, sugaredLogger)
+
+	cosmosSub.ReplayCosmosBurnLock(txFactory)
+
+	return nil
+}
+
+// RunReplayCosmosSignatureAggregationCmd replay all missed signature aggregation completed events
+func RunReplayCosmosSignatureAggregationCmd(cmd *cobra.Command, args []string) error {
+	cliContext, err := client.GetClientTxContext(cmd)
```
ElliotFriedman, line 145: same positional CLI args comment.

cmd/ebrelayer/relayer/cosmosReplay.go
```diff
@@ -0,0 +1,201 @@
...
+	"github.com/cosmos/cosmos-sdk/crypto/keyring"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/ethclient"
+	tmClient "github.com/tendermint/tendermint/rpc/client/http"
+	"go.uber.org/zap"
+)
+
+// ProphecyLifeTime signature info life time on chain
+const ProphecyLifeTime = 520000
```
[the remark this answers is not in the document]
smartyalgo: small win https://github.com/Sifchain/sifnode/pull/1918

app/tokenMigration.go
```diff
@@ -0,0 +1,139 @@
...
+
+	bankKeeper.IterateAllBalances(ctx, getAll(&addresses, &coins))
+
+	for index, address := range addresses {
+
+		coin := coins[index]
+		amount := coin.Amount
+		// clear the balance for old denom
+		coin.Amount = sdk.NewInt(0)
+		err := bankKeeper.SetBalance(ctx, address, coin)
```
smartyalgo, line 66: Looks like we are zeroing the user's balance here, and updating his balance at L74. Should we move tokenMap check (L72) up so he wont end up in a situation where he's zeroed and the token isnt in the tokenMap?

smartyalgo: Also, I think we should check if the coins is already a migrated denom, because:
we panic whenever there's an issue, so we might have to rerun migration halfway through. This means there'll be coins already using the new denom. 
In current flow, we'll zero out his balance. Then, in L72, we'll use the migrated denom to lookup tokenMap, which will fail. His balance never gets refilled.

pandaring2you:
hey @smartyalgo - this task still valid? please close out if not

ElliotFriedman:
Closing so I stop receiving notifications about pushes to this PR.

PR #1930, opened 2021-09-27
Blocklist

daniportugalgit:
What is in this PR:
- [x] Create a blocklist contract
- [x] Create unit tests for the blocklist contract
- [x] Integrate the blocklist with BridgeBank
- [x] Add unit tests for the interaction between BridgeBank and the blocklist
- [x] Create a script that fetches and parses the OFAC blocklist; it should also update the blocklist accordingly.
- [x] Deploy the blocklist to Ropsten and to the Ethereum Mainnet
- [x] Support testing with a mainnet fork
- [x] Create a runbook for the script
- [x] Create the manifest variable injector
- [x] Create a script/test to test the integration between BridgeBank and the blocklist with a mainnet fork.
- [ ] Adjust integration tests so that BridgeBank knows the blocklist before trying to move tokens

What is NOT in this PR:
Anything related to Relayer / Sifnode

What will NOT be done:
Smart contracts will NOT emit an event when a blocklisted user tries to use BridgeBank and gets blocked.
Q: Why not?
A: Because we must `revert()` whenever a blocklisted user tries to move funds in any way. Since the EVM is transactional, we cannot do anything that would change the state of the blockchain if we want to revert. Emitting an event would change the state of the blockchain. Hence, it cannot be done.


daniportugalgit:
Almost done on the smart contracts side.
I still need to test the integration with BridgeBank with a mainnet fork (that's been done for the blocklist only).

IMPORTANT: these changes will break integration tests because the BridgeBank must have the blocklist address set before it can lock, burn or do anything like that.

daniportugalgit:
Something broke along the way while I was merging develop back into this branch.
Now, I took the fastest path: copied over my files to a new branch (blocklist-v2), created a new PR (https://github.com/Sifchain/sifnode/pull/2030) and it all worked.

PR #1949, opened 2021-09-30
devenv writes to different log files for ebrelayer, witness, sifnoded

banshee:


cmd/ebrelayer/main.go
```diff
@@ -211,6 +210,7 @@ func RunInitRelayerCmd(cmd *cobra.Command, args []string) error {
 
 	logConfig := zap.NewDevelopmentConfig()
 	logConfig.Sampling = nil
+	logConfig.Encoding = "json"
```
Brando753, line 213: Are you hardcoding log output as Json and not using a flag, is that something we always want to be JSON?

cmd/ebrelayer/main.go
```diff
@@ -337,6 +336,7 @@ func RunInitWitnessCmd(cmd *cobra.Command, args []string) error {
 	validatorMoniker := args[4]
 
 	logConfig := zap.NewDevelopmentConfig()
+	logConfig.Encoding = "json"
```
Brando753, line 339: Same question as above

smart-contracts/src/devenv/ebrelayer.ts
```diff
@@ -114,7 +120,8 @@ export class RelayerRunner extends ShellCommand<EbrelayerResults> {
     readonly chainNet = "localnet",
     readonly ebrelayerDB = `levelDB.db`,
     readonly relayerdbPath = "./relayerdb",
-    readonly symbolTranslatorFile = "../test/integration/config/symbol_translator.json"
+    readonly symbolTranslatorFile = "../test/integration/config/symbol_translator.json",
+    readonly logFile = "/tmp/sifnode/ebrelayer.log"
```
Brando753, line 124: Lets change this to relayer.log as ebrelayer is the same codebase for the relayers and witnesses. Also while this is fine for now we will have to think about how we want to sanely handle default file names for multiple instances of the relayers/witnesses so relayer-1.log relayer-2.log from the readonly constructor.

Brando753 (changes requested):

cmd/ebrelayer/main.go
banshee: Yes, right now it's hardcoded.  We should probably add a feature to let users select this, but I thought we should get it working with json first.

smart-contracts/src/devenv/ebrelayer.ts
banshee: Let's talk; I think we need a new word

Brando753:
Tested on docker, it works.

Brando753 (approved):

Issue #1865, opened 2021-09-14
Remove Whitelist from Peggy2.0 Sifnode Charge Users Fees for First Time Unpegs


ElliotFriedman:
When a user tries to double peg an asset to a network for the first time, they need to be charged an amount of native tokens to compensate our relayers for the cost of creating a new IbcToken on that corresponding EVM chain.

Need to add a check when a user submits a lock tx to see if that coin has been pegged onto that destination chain.

Algorithm:
if !(netdesc => denom => bool)
     charge user fee for first time double peg, if they didn't send enough fee to cover this, tx fails

Assets that have already been double pegged onto this network do not need to pay a fee as the IbcToken has already been created on the destination chain.

juniuszhou:
https://github.com/Sifchain/sifnode/pull/1953 is ready for code review.

pandaring2you:
Merge "Feature: For new exports to Ethereum - charge the user for deployment of a new erc20 token contract for all EVM chains" https://github.com/Sifchain/sifnode/issues/1936 here

PR #1953, opened 2021-10-01
charge for first user do double peggy

juniuszhou:
according to the issue, https://github.com/sifchain/sifnode/issues/1865. we need charge the account who is the first one to do double peggy for a token.


x/ethbridge/keeper/firstLockDoublePeg.go
```diff
@@ -0,0 +1,48 @@
...
+
+import (
+	"strconv"
+
+	"github.com/Sifchain/sifnode/x/ethbridge/types"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+)
+
+// GetFirstLockDoublePeg get first lock double peg
+func (k Keeper) GetFirstLockDoublePeg(ctx sdk.Context, denom string) bool {
```
Brando753, line 11: Why are we creating a new keeper for this? Shouldn't the token registry be keeping track of deployed tokens on cross chains? If so can we just query the token registry to see if this is our first double peg?

x/ethbridge/keeper/keeper_test.go
```diff
@@ -411,14 +411,14 @@ func TestProcessLockWithReceiver(t *testing.T) {
...
-	_, err = keeper.ProcessLock(ctx, cosmosReceivers[0], account.GetSequence(), &msg, testMetadataRowan)
+	_, err = keeper.ProcessLock(ctx, cosmosReceivers[0], account.GetSequence(), &msg, testMetadataRowan, false)
 	require.ErrorIs(t, err, sdkerrors.ErrInsufficientFunds)
 
 	coins = sdk.NewCoins(sdk.NewCoin("rowan", amount), sdk.NewCoin(crossChainFee, amount))
 	_ = bankKeeper.MintCoins(ctx, types.ModuleName, coins)
 	_ = bankKeeper.SendCoinsFromModuleToAccount(ctx, types.ModuleName, cosmosReceivers[0], coins)
 
-	_, err = keeper.ProcessLock(ctx, cosmosReceivers[0], account.GetSequence(), &msg, testMetadataRowan)
+	_, err = keeper.ProcessLock(ctx, cosmosReceivers[0], account.GetSequence(), &msg, testMetadataRowan, false)
```
Brando753, line 421: Shouldn't there be a test that tests the true condition of the first double peg?

x/ethbridge/keeper/firstLockDoublePeg.go
juniuszhou: I just realize I forget other parameter as NetworkDescriptor. and I prefer separate storage for it.

x/ethbridge/keeper/keeper_test.go
juniuszhou: will add a test case.

x/ethbridge/keeper/firstLockDoublePeg.go
Brando753: So I guess the reason I would think this should be part of the token registry is two-part. 

1. I would imagine the token registry as a snapshot of the current state of deployed tokens through sifchain so from a single query I can determine token metadata and deployed contracts across chains along with if tokens are IBC native or not. Rather then having to piece information together from many sources.
2. We have to determine not only if a token has been deployed to a chain before but we also need to know if a token is native to a chain or if its a non-native asset (IBC or double pegged) to ensure that double pegged assets are burned and not locked, since we already need a mechanism to track double pegged assets from native assets and this information would need to be quired from multiple EVM chains it would make sense then if there exists no record of the asset in the token registry when you go to peg it out to an EVM chain that its A) a nonnative asset and B) that its a first-time peg of a non-native asset.

Does this make sense? If anyone else wants to chime in please do, I would just think having a central store for the state of token metadata would make sense. And I would think this information could be gleaned from redundant information that needs to exist to determine if a token is EVM native or not.

proto/sifnode/tokenregistry/v1/types.proto
```diff
@@ -58,6 +58,7 @@ message RegistryEntry {
   // the packet level. i.e rowan -> microrowan i.e microrowan -> microrowan
   string ibc_counterparty_denom = 17;
   string ibc_counterparty_chain_id = 18;
+  repeated sifnode.oracle.v1.NetworkDescriptor double_pegged_networks = 19;
```
Brando753, line 61: Should this be a mapping? So that we can lookup by network id if its been double pegged or not?

x/tokenregistry/keeper/keeper.go
```diff
@@ -156,3 +157,24 @@ func (k keeper) Exists(ctx sdk.Context, key []byte) bool {
...
+func (k keeper) GetFirstLockDoublePeg(ctx sdk.Context, denom string, networkDescriptor oracletypes.NetworkDescriptor) bool {
+	registry := k.GetDenom(ctx, denom)
+	result := true
+	for _, pegged_network := range registry.DoublePeggedNetworks {
+		if networkDescriptor == pegged_network {
+			result = false
+			break
+		}
+	}
+	return result
```
Brando753, line 170: Again feel like this should be a mapping

Brando753 (commented):
Junius, lets talk today when you are available, I think we can get this wrapped up and merged today. I just want to ask a few follow up questions.

proto/sifnode/tokenregistry/v1/types.proto
juniuszhou: the map also works, the array is simpler than a map. we put the network descriptor in the array after it is double pegged.

Brando753 (approved):
Looks good lets merge it

Issue #1716, opened 2021-08-11
`sifnoded add-genesis-clp-admin` command has incorrect description


wojciech-sif:
**Current behavior**
`sifnode add-genesis-clp-admin --help` gives incorrect output:
```
Add a genesis account to genesis.json. The provided account must specify
the account address or key name and a list of initial coins. If a key name is given,
the address will be looked up in the local Keybase. The list of initial tokens must
contain valid denominations. Accounts may optionally be supplied with vesting parameters.
```

**Expected behavior**
Printing correct description of `add-genesis-clp-admin` command

**Steps to reproduce**
- Run `sifnode add-genesis-clp-admin --help`
- Look at printed description

iverc:
addressed in #1967

PR #1967, opened 2021-10-01
Fix add-genesis-clp-admin command description

iverc:
- addresses #1716


timlind (commented):
Would like to keep PR's to a single purpose so don't think it's a good idea to up the version, also we can keep the extra upgrade handler code for future use when / if needed.

PR #2009, opened 2021-10-08
Future/peggy2 deploy ibc tokens

zikarium:
- Added forking support without tSyringe.
- Removed tSyringe from the IbcToken deployment and register scripts.
- Refactored scripts to add a more visual logging style.
- Refactored the runbooks considering that each script will be run by a different person.


daniportugalgit:
Note: we're using this branch to deploy IBC-tokens.
It's a 'frozen' branch and the idea is that we won't have to worry about code updates. Whenever we need to deploy new ibc tokens, we just use the script on this branch and it always works from the get-go.
This PR will NOT be merged in in the forseeable future.

PR #2036, opened 2021-10-14
Set registry tx 

### LINKED ISSUE #1972: [Token Registry SetRegistry Tx
The token registery should have a SetRegistry transaction that overrides the entire registry in one transaction.

This would solve the problem of adding a new IBC token which get's added as a different denom, and then needing to remove the previous IBC hash denom.

It would also simplify the process of managing the registry.

brianosaurus:
https://app.zenhub.com/workspaces/current-sprint---engineering-615a2e9fe2abd5001befc7f9/issues/sifchain/sifnode/1972

Implements a SetRegistry command

Closes #1972


x/tokenregistry/types/msgs.go
```diff
@@ -55,6 +56,56 @@ func (m *MsgRegister) GetSigners() []sdk.AccAddress {
...
 }
 
+// MsgSetRegistry
+
+func (m *MsgSetRegistry) Route() string {
+	return RouterKey
+}
+
+func (m *MsgSetRegistry) Type() string {
+	return "register"
```
timlind, line 66: copy pasta?

x/tokenregistry/client/cli/tx.go
```diff
@@ -98,6 +99,39 @@ func GetCmdRegister() *cobra.Command {
...
+			if err := msg.ValidateBasic(); err != nil {
+				return err
+			}
+			return tx.GenerateOrBroadcastTxCLI(clientCtx, cmd.Flags(), &msg)
+		},
+	}
+	flags.AddTxFlagsToCmd(cmd)
+	return cmd
+}
+
```
timlind, line 134: 👍🏻

x/tokenregistry/keeper/keeper.go
```diff
@@ -95,24 +95,24 @@ func (k keeper) RemoveToken(ctx sdk.Context, denom string) {
...
 	store := ctx.KVStore(k.storeKey)
-	bz := k.cdc.MustMarshalBinaryBare(&wl)
-	store.Set(types.WhitelistStorePrefix, bz)
+	bz := k.cdc.MustMarshalBinaryBare(wl)
+	store.Set(types.RegistryPrefix, bz)
 }
 
-func (k keeper) GetRegistry(ctx sdk.Context) types.Registry {
-	var whitelist types.Registry
+func (k keeper) GetRegistry(ctx sdk.Context) *types.Registry {
```
timlind, line 109: Return a struct here so that there's an empty value instead of nil. If a nil value can be returned we should return an error to check as well. Returning an error is not usually necessary in most keeper functions.

brianosaurus: Thanks. Done!

timlind (approved):

PR #2069, opened 2021-10-20
Fix generate token registry entry command

iverc:


x/tokenregistry/client/cli/query.go
```diff
@@ -79,7 +79,7 @@ func GetCmdGenerateEntry() *cobra.Command {
 				return err
 			}
 			flags := cmd.Flags()
-			decimals, err := flags.GetInt64(flagDecimals)
+			decimals, err := flags.GetInt(flagDecimals)
```
timlind, line 82: Seems like it should just be left as GetInt64 instead of casting below?

PR #2070, opened 2021-10-20
Add new registry entries

iverc:


x/tokenregistry/client/cli/query.go
```diff
@@ -177,7 +177,7 @@ func GetCmdGenerateEntry() *cobra.Command {
 				denom = baseDenom
 			}
 			entry := types.RegistryEntry{
-				Decimals:                 decimals,
+				Decimals:                 int64(decimals),
```
timlind, line 180: This change is in another PR can it be left as is so it's only decided on in one pr?

timlind (approved):

PR #2117, opened 2021-11-01
Min tx fee limit

iverc:


app/ante/ante.go
```diff
@@ -31,46 +30,31 @@ func NewAnteHandler(ak authkeeper.AccountKeeper, bk bankkeeper.Keeper, gasConsum
...
-	loweredGasPrice := sdk.DecCoin{
+	minGasPrice := sdk.DecCoin{
 		Denom:  "rowan",
-		Amount: sdk.MustNewDecFromStr("0.00000005"),
+		Amount: sdk.MustNewDecFromStr("100000000000000000"),
 	}
-
-	if !loweredGasPrice.IsValid() {
-		return ctx, sdkerrors.Wrap(sdkerrors.ErrLogic, "unable to lower gas price")
+	if len(msgs) == 1 && (msgs[0].Type() == disptypes.MsgTypeCreateDistribution || msgs[0].Type() == disptypes.MsgTypeRunDistribution) {
```
brianosaurus, line 49: Any reason not to loop msgs looking for one that has a min gas price? Maybe the last one has a min and [0..n-1] are zero?

iverc: @brianosaurus we were checking if len(msgs) != 1 to return error cause create/run distribution should have exactly 1 msg

brianosaurus (approved):
:+1:

brianosaurus (approved):
:+1:

intl-man (dismissed):

intl-man (approved):

PR #2120, opened 2021-11-02
Hotfix | Improve min fee logic

iverc:


app/ante/ante.go
```diff
@@ -53,13 +54,30 @@ func (r AdjustGasPriceDecorator) AnteHandle(ctx sdk.Context, tx sdk.Tx, simulate
...
+			feeTx, ok := tx.(sdk.FeeTx)
+			if !ok {
+				return ctx, sdkerrors.Wrap(sdkerrors.ErrTxDecode, "Tx must be a FeeTx")
+			}
+			fees := feeTx.GetFee()
+			if len(fees) == 1 && fees[0].Amount.LT(sdk.NewInt(100000000000000000)) { // 0.1
+				return ctx, sdkerrors.Wrap(sdkerrors.ErrLogic, "fee is too low")
+			}
+			return next(ctx, tx, simulate)
+		} else if msgs[i].Type() == "transfer" {
```
intl-man, line 70: @iverc Do we want to have a catch-all `else` just in case we add a new message type in the future?

iverc: @intl-man we do not enforce min fee for all other messages, if we add a new message type it should be explicitly included in the if check to enforce min fee, otherwise we can skip it

intl-man: @iverc Gotcha. Just checking. Looks good!

intl-man (approved):

intl-man (dismissed):

brianosaurus (approved):
👍  - I see the version number is something we missed before. Nice catch. Makes me wonder if that should auto-increment somewhere?

brianosaurus (dismissed):
Perfect! 👍

brianosaurus (approved):
Shouldn't the version be 0.9.14 or 0.9.13-rc2?

The test cases don't cover the new conditions in ante. IMHO we should push this PR ASAP regardless as long as those tests are added after the fact. 👍

PR #2124, opened 2021-11-04
improve code cover for keeper/executors

wilsonc2020:
add unit tests for 
`CreatePool`   
`CreateLiquidityProvider` 
`AddLiquidity` 
`RemoveLiquidityProvider` 
`DecommissionPool` 
`RemoveLiquidity` 
`InitiateSwap` 
`FinalizeSwap` 
`ParseToInt`


khdegraaf (changes requested):
This looks like a good start.  You should fix lint problems at a minimum before we merge, but it would be good to continue with getting the coverage up to maybe 70 or 80% in here if possible.

wilsonc2020:
> This looks like a good start. You should fix lint problems at a minimum before we merge, but it would be good to continue with getting the coverage up to maybe 70 or 80% in here if possible.

fixed lint problems. Thanks~

PR #2130, opened 2021-11-04
Update remove unneeded function param

smartyalgo:
Removing unused param. Is this ok?


banshee:
Yes, it's fine - we should probably rename that function too.

PR #2132, opened 2021-11-05
Cosmos 0.44/ibc go v2.0.0

Jedi2002:
Changes 

1. Switched to ibc-go-v2.0.0-rc0 branch .
2. Removed token converstions on transfer from sifchain -> any network 
3. Modified logic for incoming tokens as follows 
    - If incoming token is Xtoken , do convertions and check whitelist 
    - If incoming token is Non - X token , just check for whitelist  
    
Next Steps

1. Modify tokenregistry as necessary .


x/ibctransfer/keeper/msg_server.go
```diff
@@ -46,26 +45,6 @@ func (srv msgServer) Transfer(goCtx context.Context, msg *sdktransfertypes.MsgTr
...
-				return nil, types.ErrAmountTooLargeToSend
-			}
-			err := helpers.PrepareToSendConvertedCoins(goCtx, msg, token, tokenConversion, srv.bankKeeper)
-			if err != nil {
-				return nil, sdkerrors.Wrap(types.ErrConvertingToCounterpartyDenom, err.Error())
-			}
-			msg.Token = tokenConversion
-		}
-	}
 	if !msg.Token.Amount.IsInt64() {
```
AustinoBombino, line 48: Do we still need these checks?

x/ibctransfer/onrecv_ack.go
```diff
@@ -25,21 +25,21 @@ func OnAcknowledgementMaybeConvert(
 	packet channeltypes.Packet,
```
AustinoBombino, line 25: Do we still need this function? Looks like it's no longer used.

AustinoBombino (commented):
logic changes look good, just a couple questions above about whether we need to keep some things

timlind (approved):

PR #2152, opened 2021-11-11
store the denom in map for performance

juniuszhou:


x/tokenregistry/keeper/keeper.go
```diff
@@ -85,75 +85,67 @@ func (k keeper) CheckDenomPermissions(ctx sdk.Context, denom string, requiredPer
 	return true
 }
 
+func (k keeper) GetDenomPrefix(ctx sdk.Context, denom string) []byte {
```
smartyalgo, line 88: Does this need to be visible outside of the package?

smartyalgo: If not, I think we can make it lowercase

x/tokenregistry/keeper/keeper.go
```diff
@@ -85,75 +85,67 @@ func (k keeper) CheckDenomPermissions(ctx sdk.Context, denom string, requiredPer
...
-	}
 
-	return types.RegistryEntry{
-		IsWhitelisted: false,
-		Denom:         denom,
-	}
+	var entry types.RegistryEntry
+	key := k.GetDenomPrefix(ctx, denom)
+	store := ctx.KVStore(k.storeKey)
+	bz := store.Get(key)
```
smartyalgo, line 97: bz is hard to understand, can you update this var name to be more descriptive?

x/tokenregistry/keeper/keeper.go
```diff
@@ -85,75 +85,67 @@ func (k keeper) CheckDenomPermissions(ctx sdk.Context, denom string, requiredPer
 	return true
 }
 
+func (k keeper) GetDenomPrefix(ctx sdk.Context, denom string) []byte {
+	return append(types.TokenDenomPrefix, []byte(denom)...)
+}
+
 func (k keeper) GetDenom(ctx sdk.Context, denom string) types.RegistryEntry {
```
smartyalgo, line 92: This func, if denom is not found, returns a RegistryEntry with Denom = denom.

Are we preserving it in this revision? I suggest adding a unit test to encapsulate this behaviour

smartyalgo (changes requested):

x/tokenregistry/keeper/keeper.go
```diff
@@ -85,75 +85,67 @@ func (k keeper) CheckDenomPermissions(ctx sdk.Context, denom string, requiredPer
...
-		if wl.Entries[i] != nil && strings.EqualFold(wl.Entries[i].Denom, denom) {
-			return *wl.Entries[i]
-		}
-	}
 
-	return types.RegistryEntry{
-		IsWhitelisted: false,
-		Denom:         denom,
-	}
+	var entry types.RegistryEntry
```
banshee, line 94: GetDenom needs to return a valid struct if there's no key stored for the denom

banshee:
It turns out that develop has changed this so much we're going to need to rework it anyway.  Let's take this PR for peggy2 and do the work necessary when we merge in develop to future/peggy2.

banshee (approved):

x/tokenregistry/keeper/keeper.go
juniuszhou: should be lowercase, not used outside.

x/tokenregistry/keeper/keeper.go
juniuszhou: accepted.

PR #2171, opened 2021-11-12
Reward Inflation Upgrade Handler

AustinoBombino:


app/setup_handlers.go
```diff
@@ -4,13 +4,29 @@ import (
...
+		if !ok {
+			panic("failed to convert mint amount")
+		}
+		mintCoins := sdk.NewCoins(sdk.NewCoin(app.StakingKeeper.BondDenom(ctx), mintAmount))
+		err := app.BankKeeper.MintCoins(ctx, dispensationtypes.ModuleName, mintCoins)
+		if err != nil {
+			panic(err)
+		}
+		// err = app.BankKeeper.SendCoinsFromModuleToAccount(ctx, dispensationtypes.ModuleName, address, mintCoins) // TODO: get destination address
+		// if err != nil {
```
marshsif: sif1ct2s3t8u2kffjpaekhtngzv6yc4vm97xajqyl3
the destination address, which is the same one we've used for distributions in the past

app/setup_handlers.go
```diff
@@ -4,13 +4,33 @@ import (
 	storetypes "github.com/cosmos/cosmos-sdk/store/types"
 	sdk "github.com/cosmos/cosmos-sdk/types"
 	"github.com/cosmos/cosmos-sdk/x/upgrade/types"
+
+	dispensationtypes "github.com/Sifchain/sifnode/x/dispensation/types"
 )
 
-const upgradeName = "0.9.12"
+const upgradeName = "Reward-Mint"
```
intl-man, line 11: We need to set this as the version (e.g.: 0.9.13).

banshee:
This shouldn't involve develop.  Needs to be a branch from master.

banshee (changes requested):
This can't be branched from develop.

app/setup_handlers.go
```diff
@@ -4,13 +4,33 @@ import (
...
 
 func SetupHandlers(app *SifchainApp) {
 	app.UpgradeKeeper.SetUpgradeHandler(upgradeName, func(ctx sdk.Context, plan types.Plan) {
 		app.Logger().Info("Running upgrade handler for " + upgradeName)
+
+		mintAmount, ok := sdk.NewIntFromString("200000000000000000000000000")
+		if !ok {
+			panic("failed to convert mint amount")
+		}
+		mintCoins := sdk.NewCoins(sdk.NewCoin(app.StakingKeeper.BondDenom(ctx), mintAmount))
```
banshee, line 21: We need a check to make sure the mint only happens on a single release.  If someone did a hotfix without removing this code, I think we'd mint another batch.

banshee (changes requested):

AustinoBombino:
closed to branch off of master:
https://github.com/Sifchain/sifnode/pull/2172

iverc (commented):
@AustinoBombino we should add a version check to only trigger the logic on a specific version (0.9.14) and discard it otherwise.

PR #2173, opened 2021-11-12
Feature | Inflation Rewards

iverc:


version
```diff
@@ -1 +1 @@
-0.9.12
+0.9.14
```
banshee, line 1: The previous version was 0.9.12, should this be 13?

app/setup_handlers.go
```diff
@@ -4,29 +4,51 @@ import (
...
 	sdk "github.com/cosmos/cosmos-sdk/types"
 	"github.com/cosmos/cosmos-sdk/x/upgrade/types"
+
+	dispensationtypes "github.com/Sifchain/sifnode/x/dispensation/types"
 )
 
-const upgradeName = "0.9.12"
+const upgradeName = "0.9.14"
+
+const distributionVersion = "0.9.14"
```
Jedi2002, line 13: Not a problem , but any reason why we have two constants with the same value ?

Jedi2002 (approved):

app/setup_handlers.go
iverc: @Jedi2002 cause the `upgradeName` constant will be changing for next versions, while the dispensation is a one time event

version
iverc: @banshee I think we have not made the previous deployment with fix update (was deployed to testnet, but not mainnet) and in that fix the version was changed from 0.9.12 to 0.9.13
https://github.com/Sifchain/sifnode/pull/2120
I think we should merge that PR first (also fixes integration tests), and then do this one. Does it sound good?
cc @intl-man @AustinoBombino @Jedi2002

intl-man: @iverc Agreed. #2120 has been merged.

intl-man (approved):

PR #2175, opened 2021-11-14
[ChainOps] Build updates

intl-man:
Includes:

* No longer prefix builds by the network (not necessary).
* Updated the release action to correctly set `ldflags`.
* Cleanup of the `sifnode` Makefile.


iverc (commented):
👍

PR #2176, opened 2021-11-15
Ethereum blacklisting

timlind:


brianosaurus:
I just looked at this @timlind and I didn't notice a cli command. Sure you know this, just throwing that out there.

timlind:
> I just looked at this @timlind and I didn't notice a cli command. Sure you know this, just throwing that out there.

Thanks for the reminder. The peggy side is actually using a JS script to do the updates so I haven't decided whether we'll do the same or need a CLI command, but I guess we should add it in either ways.

brianosaurus (approved):
👍

PR #2186, opened 2021-11-17
get the network id from cli

### LINKED ISSUE #2082: peggy2: cmd/ebrelayer/relayer/ethereum.go calls ethClient.NetworkID to get the network id, it should use value from command line
peggy2: cmd/ebrelayer/relayer/ethereum.go calls ethClient.NetworkID to get the network id, it should use value from command line.  In general we don't want to rely on what the ethereum chain reports for its network id.

juniuszhou:
The pr resolve #2082, get the network id from command line, not depend on the id what chain report.

### DIFF
--- a/cmd/ebrelayer/main.go
+++ b/cmd/ebrelayer/main.go
@@ -2,13 +2,14 @@ package main
 
 import (
 	"fmt"
-	"github.com/Sifchain/sifnode/x/instrumentation"
 	"log"
 	"net/url"
 	"os"
 	"strconv"
 	"strings"
 	"sync"
+
+	"github.com/Sifchain/sifnode/x/instrumentation"
 
 	"github.com/Sifchain/sifnode/cmd/ebrelayer/internal/symbol_translator"
 	"github.com/Sifchain/sifnode/cmd/ebrelayer/txs"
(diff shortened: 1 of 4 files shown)


cmd/ebrelayer/main.go
```diff
@@ -360,18 +370,28 @@ func RunInitWitnessCmd(cmd *cobra.Command, args []string) error {
 func replayEthereumCmd() *cobra.Command {
 	//nolint:lll
 	replayEthereumCmd := &cobra.Command{
-		Use:     "replayEthereum [tendermintNode] [web3Provider] [bridgeRegistryContractAddress] [validatorMoniker]",
+		Use:     "replayEthereum  [networkDescriptor][tendermintNode] [web3Provider] [bridgeRegistryContractAddress] [validatorMoniker]",
```
banshee, line 373: Instead of adding networkDescriptor this way, can we add it using the new AddRelayerFlagsToCmd function?  We should be using the Cobra features to add options, generally.

banshee (changes requested):

cmd/ebrelayer/main.go
smartyalgo: Do we need to keep replayEthereum()?

smartyalgo (approved):
LGTM. 
2 issues i see (that im ok not being addressed in this PR), is 
1. Replacing arguments with flags
2. Deprecating the replay functions because theyre no longer needed for peggy 2

cmd/ebrelayer/main.go
juniuszhou: I agree we can remove replay functions in peggy2, will use a new commit to do it. For flags instead of arguments, I will create a separate pr to refactor all those arguments before.

PR #2188, opened 2021-11-17
+frontend file for november 6th whitelist

daniportugalgit:


pandaring2you (approved):
null image urls?

daniportugalgit:
> null image urls?

It can happen. This script tries its best to fetch images automagically - it does that by querying Alchemy directly. Most of the time, Alchemy will have an image, but the frontend team is supposed to review it and fill in null fields and change URLs when appropriate (if we want to use icons from somewhere else, for example).

banshee (approved):

PR #2191, opened 2021-11-17
feat: 🎸 parallelize token distribution process with

canercandan:
as many threads as assets to pre-mint


khdegraaf:
@jzvikart rewrote the code, so this isn't relevant anymore.

PR #2197, opened 2021-11-18
Cosmos 0.44/upgrade handler

Jedi2002:


app/setup_handlers.go
```diff
@@ -1,17 +1,30 @@
...
+			registry := app.TokenRegistryKeeper.GetRegistry(ctx)
+			for _, entry := range registry.Entries {
+				if entry.Decimals > 9 && app.TokenRegistryKeeper.CheckEntryPermissions(entry, []tokenRegistryTypes.Permission{tokenRegistryTypes.Permission_CLP, tokenRegistryTypes.Permission_IBCEXPORT}) {
+					entry.Permissions = append(entry.Permissions, tokenRegistryTypes.Permission_IBCIMPORT)
+					entry.IbcCounterpartyDenom = ""
+				}
+			}
+			app.TokenRegistryKeeper.SetRegistry(ctx, registry)
+		}
 		return vm, nil
```
timlind, line 28: "All upgrade handler functions must end with the following line of code:"

```return app.mm.RunMigrations(ctx, cfg, fromVM)```

https://docs.cosmos.network/master/core/upgrade.html#running-migrations

File: app/setup_handlers.go [hunk 2]
```diff
@@ -1,17 +1,30 @@
...
 	storetypes "github.com/cosmos/cosmos-sdk/store/types"
 	sdk "github.com/cosmos/cosmos-sdk/types"
 	m "github.com/cosmos/cosmos-sdk/types/module"
+	"github.com/cosmos/cosmos-sdk/x/authz"
+	"github.com/cosmos/cosmos-sdk/x/feegrant"
 	"github.com/cosmos/cosmos-sdk/x/upgrade/types"
 )
 
-const upgradeName = "0.9.10"
+const upgradeName = "0.9.14"
```
timlind, line 13: should be 0.10.0-rc.1

app/ante/ante.go
```diff
@@ -7,90 +7,88 @@ import (
...
-			Amount: sdk.MustNewDecFromStr("0.00000005"),
-		}
-		if !minGasPrice.IsValid() {
-			return ctx, sdkerrors.Wrap(sdkerrors.ErrLogic, "invalid gas price")
-		}
-		ctx = ctx.WithMinGasPrices(sdk.NewDecCoins(minGasPrice))
-		return next(ctx, tx, simulate)
-	}
-	minFee := sdk.ZeroInt()
+	var found bool
```
timlind, line 63: This is the old ante handler, develop has the handler changes and rewritten for 0.44

app/setup_handlers.go
```diff
@@ -4,23 +4,40 @@ import (
...
+		// Old Modules can execute Migrations if needed .
+		// State migration logic should be added to a migrator function
+		// Migrating modules should increment the ConsensusVersion
+		for moduleName := range app.mm.Modules {
+			fromVM[moduleName] = 1
+		}
+		// New Modules must execute Init Genesis
+		fromVM[authz.ModuleName] = 0
+		fromVM[feegrant.ModuleName] = 0
+		fromVM["vesting"] = 0
```
timlind, line 29: I don't see vesting or crisis installed as a new module?

x/tokenregistry/keeper/migrations.go
```diff
@@ -0,0 +1,29 @@
...
+	registry := m.keeper.GetRegistry(ctx)
+	for _, entry := range registry.Entries {
+		if entry.Decimals > 9 && m.keeper.CheckEntryPermissions(entry, []tkrtypes.Permission{tkrtypes.Permission_CLP, tkrtypes.Permission_IBCEXPORT}) {
+			entry.Permissions = append(entry.Permissions, tkrtypes.Permission_IBCIMPORT)
+			entry.IbcCounterpartyDenom = ""
+		}
+	}
+	m.keeper.SetRegistry(ctx, registry)
+	return nil
+}
```
timlind, line 29: We need to keep the registry tx scripts up to date with this: scripts/ibc/tokenregistry/

app/setup_handlers.go
Jedi2002: The current master does not have vesting or crisis module . I assumed that master is what is deployed on presently on betanet 
https://github.com/Sifchain/sifnode/blob/master/app/app.go

Or do you mean that 
We don't need to add these modules during our 0.44 upgrade ?  
I am actually not clear on why these modules were added , these were already present when I started working on this PR , I assumed we need these modules . Mayber @iverc can comment .

x/tokenregistry/keeper/migrations.go
Jedi2002: Please review this script 
https://github.com/Sifchain/sifnode/blob/a5f74f0f6809cb4d3580ba959234dcf4db9990ab/scripts/ibc/tokenregistration/main.go

If found to be okay , the older files can be replaced 
https://github.com/Sifchain/sifnode/blob/a5f74f0f6809cb4d3580ba959234dcf4db9990ab/scripts/ibc/tokenregistration/main.go

app/setup_handlers.go
intl-man: @Jedi2002 I don't believe vesting is required. Any further thoughts?

iverc: @intl-man @Jedi2002 @timlind Vesting module is not required, but could be useful in the near future. I'd include it to avoid having a separate upgrade for it specifically as long as it does not introduce issues

iverc (commented):
@Jedi2002 looks good! @intl-man should we merge in changes from master to develop first?

iverc:
@Jedi2002 integration tests are failing (unable to connect to localhost). Should we use testnet?

app/setup_handlers.go
intl-man: That makes sense @iverc.

intl-man:
> @Jedi2002 integration tests are failing (unable to connect to localhost). Should we use testnet?

We should fix these tests. If localhost is not working, then it means that the service has not started for some reason. I'll take a look, but @Jedi2002 would appreciate your insight here too.

intl-man:
> @Jedi2002 looks good! @intl-man should we merge in changes from master to develop first?

Changes have already been merged down @iverc.

timlind (approved):

iverc:
> > @Jedi2002 looks good! @intl-man should we merge in changes from master to develop first?
> 
> Changes have already been merged down @iverc.

thanks!

PR #2206, opened 2021-11-19
Prettier: from tabs to spaces

daniportugalgit:
Updating whitespace in .js, .ts and .sol files.
We were using tabs and now we're using spaces.

There are no functionality changes here, only whitespace.


Brando753 (changes requested):
Please update the one comment and I will approve

Brando753 (approved):

Issue #2129, opened 2021-11-04
Output useful tools from Integration ENV


Brando753:
IntegrationENV should output a file with useful commands already filled out for operating sifnode correctly to increase development velocity.

smartyalgo:
Love this. Maybe even source it and it gives us alias so we dont hv to type out the whole thing.

pandaring2you:
No longer doing on devenv. Moving to integration env.

banshee:
You can look at test/integration/environment_setup.sh for an example of how we used to do this for the old environment.

pandaring2you:
Finished!

PR #2207, opened 2021-11-19
Update integration environment in Peggy2.0

jzvikart:
This is the same as #2203, but applied to Peggy2.0 branch.
Prerequisite for #2129.


jzvikart:
Merge request withdrawn since review/approval is not required for merging into Peggy2.0.

PR #2215, opened 2021-11-23
use command flag

### LINKED ISSUE #2216: Use the command flag to replace arguments for ebrelayer
Command flag is better way to input parameters for ebrelayer, it is more flexible.

juniuszhou:
The pr remove the arguments in the command line for ebrelayer, instead use the flag. will resolves #2216


cmd/ebrelayer/main.go
```diff
@@ -165,35 +167,40 @@ func RunInitRelayerCmd(cmd *cobra.Command, args []string) error {
 	}
 
 	// Validate and parse arguments
-	networkDescriptor, err := strconv.Atoi(args[0])
+	networkDescriptor, err := cmd.Flags().GetInt32(networkDescriptorFlag)
 	if err != nil {
-		return errors.Errorf("%s is invalid network id", args[0])
+		return err
```
Brando753, line 172: Is the new error message descriptive or generic? Does it specify that the network id flag was missing/invalid?

cmd/ebrelayer/main.go
```diff
@@ -370,6 +381,26 @@ func AddRelayerFlagsToCmd(cmd *cobra.Command) {
 		int32(oracleTypes.NetworkDescriptor_NETWORK_DESCRIPTOR_ETHEREUM),
 		"The network descriptor for the chain",
 	)
+	cmd.Flags().String(
```
Brando753, line 384: Are these flags being marked required or does the fetch and check error handle that automatically?

smart-contracts/scripts/devenv.ts
```diff
@@ -118,7 +118,9 @@ async function main() {
       sifnode.results.relayerAddresses[0],
       sifnode.results.witnessAddresses[0],
       golang.results,
-      hardhat.results.chainId
```
Brando753, line 121: Is there a reason this value is not able to come from the hardhat builder?

Brando753 (commented):
Once the questions are answered I will approve.

smart-contracts/scripts/devenv.ts
juniuszhou: The chainId from hardhat is 1, we need to configure it as 31337.

cmd/ebrelayer/main.go
juniuszhou: I didn't find any parameter to set the flag as mandatory, and cobra will check the validatity.

cmd/ebrelayer/main.go
juniuszhou: it is a generic error description.

Brando753: Can we make this specific? The problem when using this CLI is when a flag is incorrect/missing generic error messages make it a real pain to debug. What happens if the flag is missing here does it return an error?

juniuszhou: OK, will add specific error info in the command line.

Brando753 (approved):

PR #2242, opened 2021-12-01
test: 💍 add unit tests for margin/keeper module

canercandan:


x/margin/keeper/keeper_test.go
```diff
@@ -0,0 +1,197 @@
...
+	mtp.CustodyAmount = sdk.NewUint(0)
+	// FIX
+
+	ctx, app := test.CreateTestAppMargin(false)
+	marginKeeper := app.MarginKeeper
+	assert.NotNil(t, marginKeeper)
+
+	err := marginKeeper.TakeInCustody(ctx, mtp, pool)
+	assert.Nil(t, err)
+}
```
Jedi2002, line 197: General comments for all of these tests in the selection box 
- There is quite a bit of value in writing the tests in the form of suits , so that new values can be added in easily .
- The assertions should also assert the value set in the keeper and not just the error message .
- The assert values should be dependant on inputs. It might be a good idea to generalize the tests to also include tables ( json files with inputs and outputs calculated using a different program ) and asserting the output fields .

canercandan: @Jedi2002 thanks for your review

> There is quite a bit of value in writing the tests in the form of suits , so that new values can be added in easily .

Indeed my plan is to convert those aggregated tests to table-driven tests, to keep adding new test parameters going forward, this brings good foundation for fuzz testing with randomized set of values along with SimApp.

> The assertions should also assert the value set in the keeper and not just the error message .

Indeed my tests are missing assertion against the inputs I will definitely add that too. Thanks.

> The assert values should be dependant on inputs. It might be a good idea to generalize the tests to also include tables ( json files with inputs and outputs calculated using a different program ) and asserting the output fields .

True that is linked to the first bulletpoint, I do have any idea of how to approach the JSON inputs/output mapping used within the test scopes but feel free to share any code samples that do that, that could save me some time thanks ;-)

canercandan:
@timconner @Jedi2002 @AustinoBombino fyi 85% code coverage on the keeper methods

x/margin/keeper/genesis_test.go
```diff
@@ -0,0 +1,26 @@
+package keeper_test
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
```
timlind, line 6: Generally using require instead of assert is preferred.

x/margin/keeper/genesis_test.go
```diff
@@ -0,0 +1,26 @@
...
+
+	"github.com/Sifchain/sifnode/x/margin/test"
+)
+
+func TestKeeperExportGenesis(t *testing.T) {
+	ctx, app := test.CreateTestAppMargin(false)
+	marginKeeper := app.MarginKeeper
+	assert.NotNil(t, marginKeeper)
+	state := marginKeeper.ExportGenesis(ctx)
+	assert.NotNil(t, state)
```
timlind, line 16: We should test for what we want to export here, params and mtps for example.

x/margin/keeper/keeper_test.go
```diff
@@ -0,0 +1,585 @@
...
+		err := marginKeeper.SetMTP(ctx, &mtp)
+		assert.NoError(t, err)
+	})
+}
+
+func TestKeeper_GetMTP(t *testing.T) {
+	t.Run("get MTP from a store key that exists", func(t *testing.T) {
+		ctx, app, marginKeeper := initKeeper(t)
+		want := addMTPKey(t, ctx, app, marginKeeper, "ceth", "xxx")
+		got, err := marginKeeper.GetMTP(ctx, "ceth", "xxx")
```
timlind, line 49: GetMTP signature has been updated to take in collateralAsset, custodyAsset, mtpAddress.

x/margin/keeper/keeper_test.go
```diff
@@ -0,0 +1,585 @@
...
+			t.Error("got an error but didn't want one")
+		}
+		if !reflect.DeepEqual(got, want) {
+			t.Errorf("got %v want %v", got, want)
+		}
+	})
+
+	t.Run("fails when store keys does not exist", func(t *testing.T) {
+		ctx, _, marginKeeper := initKeeper(t)
+		marginKeeper.GetMTP(ctx, "ceth", "xxx")
```
timlind, line 61: Assertion missing here.

x/margin/keeper/keeper_test.go
```diff
@@ -0,0 +1,585 @@
...
+
+func TestKeeper_GetMTPIterator(t *testing.T) {
+	ctx, app, marginKeeper := initKeeper(t)
+	want := addMTPKey(t, ctx, app, marginKeeper, "ceth", "xxx")
+	iterator := marginKeeper.GetMTPIterator(ctx)
+	bytesValue := iterator.Value()
+	var got types.MTP
+	types.ModuleCdc.MustUnmarshal(bytesValue, &got)
+
+	if !reflect.DeepEqual(got, want) {
```
timlind, line 73: Don't see a need for deep equal here, I would suggest sticking with the assertion library require.Equal()

x/margin/keeper/keeper_test.go
```diff
@@ -0,0 +1,585 @@
...
+		t.Errorf("got %v want %v", got, want)
+	}
+}
+
+func TestKeeper_GetMTPsForAsset(t *testing.T) {
+	ctx, app, marginKeeper := initKeeper(t)
+	key1 := addMTPKey(t, ctx, app, marginKeeper, "ceth", "key1")
+	key2 := addMTPKey(t, ctx, app, marginKeeper, "ceth", "key2")
```
timlind, line 93: Could add a non ceth entry here and ensure it isn't returned.

x/margin/keeper/keeper_test.go
```diff
@@ -0,0 +1,585 @@
...
+		t.Fatal("didn't get an error but wanted one")
+	}
+	if got != want {
+		t.Errorf("got %q, want %q", got, want)
+	}
+}
+func assertErrorString(t testing.TB, got error, want error) {
+	t.Helper()
+	if got == nil {
+		t.Fatal("didn't get an error but wanted one")
```
timlind, line 574: using require instead of assert will cause the test case to stop.

x/margin/keeper/keeper_test.go
```diff
@@ -0,0 +1,585 @@
...
+func assertError(t testing.TB, got error, want error) {
+	t.Helper()
+	if got == nil {
+		t.Fatal("didn't get an error but wanted one")
+	}
+	if got != want {
+		t.Errorf("got %q, want %q", got, want)
+	}
+}
+func assertErrorString(t testing.TB, got error, want error) {
```
timlind, line 571: These 3 helpers shouldn't be necessary when using testify methods directly.

x/margin/keeper/keeper_test.go
```diff
@@ -0,0 +1,585 @@
...
+		t.Errorf("got %v want %v", got, want)
+	}
+}
+
+func TestKeeper_DestroyMTP(t *testing.T) {
+	t.Run("key does not exist", func(t *testing.T) {
+		ctx, _, marginKeeper := initKeeper(t)
+		got := marginKeeper.DestroyMTP(ctx, "ceth", "xxx")
+
+		assertError(t, got, types.ErrMTPDoesNotExist)
```
timlind, line 121: require.ErrorIs(t, err, types.ErrMTPDoesNotExist)

x/margin/keeper/keeper_test.go
```diff
@@ -0,0 +1,585 @@
...
+	})
+	t.Run("key exists", func(t *testing.T) {
+		ctx, app, marginKeeper := initKeeper(t)
+		addMTPKey(t, ctx, app, marginKeeper, "ceth", "xxx")
+		got := marginKeeper.DestroyMTP(ctx, "ceth", "xxx")
+
+		assertNoError(t, got)
```
timlind, line 128: require.NoError(t, err)

x/margin/keeper/keeper_test.go
```diff
@@ -0,0 +1,585 @@
...
+			denom:      "rowan",
+			decimals:   9,
+			to:         "rowan",
+			sentAmount: sdk.NewUint(1000000000000),
+			err:        clptypes.ErrNotEnoughAssetTokens,
+		},
+	}
+
+	for _, tt := range custodySwapTests {
+		t.Run(tt.name, func(t *testing.T) {
```
timlind, line 220: For the linter errors below, insert the line tt := tt before passing tt to t.Run()

x/margin/keeper/keeper_test.go
```diff
@@ -0,0 +1,585 @@
...
+				Decimals:    tt.decimals,
+				Permissions: []tokenregistrytypes.Permission{tokenregistrytypes.Permission_CLP},
+			})
+
+			mtp := addMTPKey(t, ctx, app, marginKeeper, tt.to, tt.address)
+
+			got := marginKeeper.Borrow(ctx, tt.to, tt.collateralAmount, tt.borrowAmount, mtp, pool, tt.leverage)
+
+			if tt.errString != nil {
+				assertErrorString(t, got, tt.errString)
```
timlind, line 344: require.Equal(t, tt.errString, err.Error())

x/margin/keeper/keeper_test.go
Jedi2002: Hi for test tables here is some sample code , this can be a good start 
http://github.com/Sifchain/sifnode/blob/4e5002434b1be3adc12c3591d36f1c91529b8add/x/clp/keeper/table_test.go#L22-L23

x/margin/keeper/genesis_test.go
canercandan: updated the test from init/export genesis functions so it asserts the existence of the params field along with its fields but I am not familiar with MTP-related fields with the GenesisState struct, can you clarify thanks @timlind 

fyi: https://github.com/Sifchain/sifnode/pull/2242/commits/e1ec8f1c00602bdb09806d687c20e6d047416387#diff-6f2afbfb379f05d7031c6f9a0d9754f691a5a890818e5a367beaa7c255c640c9R64-R87

PR #2243, opened 2021-12-02
feat: margin open long

timlind:
[x] msg server
[x] cli


x/margin/keeper/keeper.go
```diff
@@ -102,3 +105,112 @@ func (k Keeper) DestroyMTP(ctx sdk.Context, symbol string, mtpAddress string) er
...
+	   Notice this is NOT a simple hybrid of uniswap model and Thorchain slippaged based model
+	   But a upgraded version that include swap, updating bouding curve (to be inside the old one)
+	   One can think about this as a state jump:
+	*/
+
+	// todo review
+	//receivedAmount := sentAmount.Mul(receivedBalance.Add(receivedLiabilities)).
+	//	Quo(sentAmount.Add(sentBalance).Add(sentLiabilities)).
+	//	Add(sentAmount.Mul(receivedBalance.Add(receivedLiabilities)).
+	//	Mul(sentBalance.Add(sentLiabilities)).Quo(sentAmount.Add(sentBalance).Add(sentLiabilities))).MulUint64(2)
```
timlind, line 134: @AustinoBombino can we get this on the spec as a multi step function using the sdk math functions?

AustinoBombino: @timlind Nice catch, Just got into this and realized that it also probably needs to be more sophisticated to account for our decimal normalization logic.

AustinoBombino: @timlind CustodySwap with sdk math functions and decimal normalization is in this PR: https://github.com/Sifchain/Documentation-Private/pull/2

x/margin/types/genesis.go
```diff
@@ -0,0 +1,12 @@
+package types
+
+import sdk "github.com/cosmos/cosmos-sdk/types"
+
+func DefaultGenesis() *GenesisState {
+	return &GenesisState{
+		// TODO review default param values
+		Params: &Params{
+			LeverageMax: sdk.NewUint(0),
+		},
```
timlind, line 10: @AustinoBombino are there default values we should be using here for the params?

AustinoBombino: Yes, this will be something we'll have to tune. But we can start it at 1.

PR #2263, opened 2021-12-09
Cli set consensus needed

### LINKED ISSUE #2236: flexible witness and relayer node number for Devnet test env
Currently, we have one witness node and one relayer node in Devnet. To verify the different threshold, need more flexible setting for these nodes.

### LINKED ISSUE #2264: Add cli to set consensus for each network

juniuszhou:
PR resolves #2264


Brando753 (approved):
We may want to revisit this PR as there are a few concerns I have, however we are doing a massive merge with develop today and I want this code in before that merge.

PR #2274, opened 2021-12-13
0.10.0-rc.4

timlind:
Upgrade for devnet and testnet from rc.3

### DIFF
--- a/app/setup_handlers.go
+++ b/app/setup_handlers.go
@@ -11,7 +11,7 @@ import (
 	ibcconnectiontypes "github.com/cosmos/ibc-go/v2/modules/core/03-connection/types"
 )
 
-const upgradeName = "0.10.0-rc.4"
+const upgradeName = "0.10.0"
 
 func SetupHandlers(app *SifchainApp) {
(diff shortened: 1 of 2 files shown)


app/setup_handlers.go
```diff
@@ -38,6 +38,11 @@ func SetupHandlers(app *SifchainApp) {
 		// This is to make sure auth module migrates after staking
 		return app.mm.RunMigrations(ctx, app.configurator, newVM)
 	})
+	app.UpgradeKeeper.SetUpgradeHandler("0.10.0-rc.4", func(ctx sdk.Context, plan types.Plan, vm m.VersionMap) (m.VersionMap, error) {
```
Jedi2002, line 41: this might skip some of the migratios because we are not explicitly settng the version to 1 .
Is that the plan for this handler ?

timlind: Well haven't those migrations already been run with rc.3? Setting to 1 is only for the first time a the new migration system is used right?

Jedi2002 (approved):

PR #2295, opened 2021-12-20
Integ test enable multiple coins

wilsonc2020:
Integ test enable multiple coins


x/dispensation/handler_test.go
```diff
@@ -74,6 +75,196 @@ func TestNewHandler_CreateDistribution_MultipleTypes(t *testing.T) {
...
+	assert.NoError(t, err)
+	totalCoins = totalCoins.Add(totalCoins...)
+	err = sifapp.AddCoinsToAccount(types.ModuleName, app.BankKeeper, ctx, distributor, totalCoins)
+	assert.NoError(t, err)
+	// Airdrop distribution type
+	msgAirdrop := types.NewMsgCreateDistribution(distributor, types.DistributionType_DISTRIBUTION_TYPE_AIRDROP, outputList, runner.String())
+	res, err := handler(ctx, &msgAirdrop)
+	require.NoError(t, err)
+	require.NotNil(t, res)
+	res, err = handler(ctx, &msgAirdrop)
```
timlind, line 96: why is this being created twice?

x/dispensation/handler_test.go
```diff
@@ -74,6 +75,196 @@ func TestNewHandler_CreateDistribution_MultipleTypes(t *testing.T) {
...
+	distributionName := fmt.Sprintf("%d_%s", ctx.BlockHeight(), msgAirdrop.Distributor)
+	msgRun := types.NewMsgRunDistribution(runner.String(), distributionName, types.DistributionType_DISTRIBUTION_TYPE_AIRDROP)
+	res, err = handler(ctx, &msgRun)
+	require.NoError(t, err)
+	require.NotNil(t, res)
+	records := keeper.GetRecordsForNameAndStatus(ctx, distributionName, types.DistributionStatus_DISTRIBUTION_STATUS_COMPLETED)
+	assert.Len(t, records.DistributionRecords, 3)
+	records = keeper.GetRecordsForNameAndStatus(ctx, distributionName, types.DistributionStatus_DISTRIBUTION_STATUS_PENDING)
+	assert.Len(t, records.DistributionRecords, (recipients*2)-3)
+
```
timlind, line 119: Can the additional coin balance be asserted here on the receiving addresses?

x/dispensation/handler_test.go
```diff
@@ -74,6 +75,196 @@ func TestNewHandler_CreateDistribution_MultipleTypes(t *testing.T) {
...
+	assert.Equal(t, recipients, len(msgValidatorSubsidyOutput))
+	for i := 0; i < len(msgValidatorSubsidyOutput); i++ {
+		assert.True(t, msgValidatorSubsidyOutput[i].Coins.AmountOf("catk").Equal(sdk.NewInt(10)) ||
+			msgValidatorSubsidyOutput[i].Coins.AmountOf("ceth").Equal(sdk.NewInt(10)))
+	}
+
+	msgValidatorSubsidy = types.NewMsgCreateDistribution(distributor, types.DistributionType_DISTRIBUTION_TYPE_VALIDATOR_SUBSIDY, outputList, "")
+	_, err = handler(ctx, &msgValidatorSubsidy)
+	assert.Error(t, err, "Failed in collecting funds")
+}
```
timlind, line 266: Can this be rewritten as a test table with several cases?

wilsonc2020:
> Going by the name of the PR , it might be a good idea to add integration tests which interact through the CLI / REST , and run a blockchain network .

we are in the middle way of migrating python integration tests to golang. We don't want step back.

x/dispensation/handler_test.go
wilsonc2020: Second handler will fail and we catch that at line 98 and line 99.

x/dispensation/handler_test.go
wilsonc2020: sure.

x/dispensation/handler_test.go
wilsonc2020: So I tried to use test table here. But it is pretty hard to do that for this case, because we use randomize logic when try to distribute coins to users, so the output is not fixed.

wilsonc2020:
Hi @khdegraaf and @timlind please review the latest version of this pr. I have no idea why the lint is broken. Looks like all recent pr got the same error:   Error: `precendence` is a misspelling of `precedence` (misspell). Thanks.

PR #2297, opened 2021-12-22
Fix/tokenregistry output

timlind:
The generate command is now outputting yaml so needs to take an output flag to specify json


brianosaurus (approved):
:+1: I used it and works geat.

PR #2302, opened 2021-12-23
Just an update for iris.json

brianosaurus:
Minor update to the devnet token


timlind (approved):

timlind (changes requested):
@brianosaurus can you include the changes to the values in the scripts that are used to generate this JSON?

timlind (approved):

PR #2309, opened 2021-12-30
add new case peggy erc20 token

### LINKED ISSUE #2286: Extend Typescript Test
- Have the typescript test to lock/burn with ETH do the same with erc-20 tokens

juniuszhou:
The pr resolves #2286


banshee:
I think we should merge this since it has no other impacts; it's one new file and one trivial change to another file.  If we decide to just refactor test_lockburn.js it's easy to pull this out.

banshee (approved):

PR #2330, opened 2022-01-07
Margin Force Close

AustinoBombino:
Force close margin position by any address


marshsif:
@canercandan  to review

canercandan (approved):
all tests looks great thanks for adding them and will do a post test coverage analysis once everything gets merged with the parent branch

x/margin/client/cli/tx.go
```diff
@@ -110,3 +110,51 @@ func GetCloseLongCmd() *cobra.Command {
 	return cmd
 
 }
+
+func GetForceCloseLongCmd() *cobra.Command {
```
timlind, line 114: Will there be a need for a different force close api / cmd for different types of positions or can this just be force-close? How will an agent know what type of position they are closing?

timlind: Also how do we enforce on sifnode that the correct closure is used?

AustinoBombino: @timlind 
Added a position field to the MTP type. We'll use "long" and "short" as the types.

My plan is to have two sets of messages and handlers eg: OpenLong, CloseLong, ForceCloseLong   /   OpenShort, CloseShort, ForceCloseShort

Within each handler we can then determine if the position is a double based on the CollateralAsset and CustodyAsset, just like we do in CLP when determining if we need to perform a double swap.

AustinoBombino: So we can either have two sets of messages: OpenLong, CloseLong, ForceCloseLong  /  OpenShort, CloseShort, ForceCloseShort

Or we can add the position type as an argument to one set of messages: Open, Close, ForceClose
and then handle routing to the right functions within the msgserver

Do you have any thoughts/preferences?

AustinoBombino: @timlind @canercandan 

this PR currently has the first option implemented. I also put together the second option in this PR: https://github.com/Sifchain/sifnode/pull/2418

I'm not totally sold on either solution so would be keen to hear your thoughts.

canercandan: @AustinoBombino here are two examples of CEXs that use the field `side` to define what type of position between "LONG" and "SHORT" or "BUY" and "SELL" as to be created with the position/order:

* https://docs.ftx.com/#place-order
* https://bybit-exchange.github.io/docs/inverse/#t-placeactive

```
{
  "market": "XRP-PERP",
  "side": "sell",
  "price": 0.306525,
  "type": "limit",
  "size": 31431.0,
  "reduceOnly": false,
  "ioc": false,
  "postOnly": false,
  "clientId": null
}
```

I guess going for either option is fine as long the API/CLI experience allow to easily choose which type of position to open/close with a simple field/flag with all the others parameters shaped the same.

cc @timlind

timlind (approved):

canercandan (approved):

PR #2345, opened 2022-01-11
convert Integ test liquidity in python to golang.

wilsonc2020:
convert Integ test liquidity in python to golang. Basically cover all testcases we have in `test_liquidity_pools`.


Jedi2002 (commented):
Hi @wilsonc2020 Do you think it makes sense to use a test suite / table to write out the test cases . I feel functions like `TestPoolMultiplyCases` would benifit a lot from this approach . resulting in a cleaner and more readable code

Jedi2002:
> Hi @wilsonc2020 Do you think it makes sense to use a test suite / table to write out the test cases . I feel functions like `TestPoolMultiplyCases` would benifit a lot from this approach . resulting in a cleaner and more readable code

You can use these as an examples 

http://github.com/Sifchain/sifnode/blob/021dd1c1a7ca9f95aeac14208a6c9f68cc37a932/x/tokenregistry/types/msgs_test.go#L11-L11

http://github.com/Sifchain/sifnode/blob/021dd1c1a7ca9f95aeac14208a6c9f68cc37a932/x/tokenregistry/handler/handler_test.go#L121-L121

wilsonc2020:
> > Hi @wilsonc2020 Do you think it makes sense to use a test suite / table to write out the test cases . I feel functions like `TestPoolMultiplyCases` would benifit a lot from this approach . resulting in a cleaner and more readable code
> 
> You can use these as an examples
> 
> [http://github.com/Sifchain/sifnode/blob/021dd1c1a7ca9f95aeac14208a6c9f68cc37a932/x/tokenregistry/types/msgs_test.go#L11-L11](https://github.com/Sifchain/sifnode/blob/021dd1c1a7ca9f95aeac14208a6c9f68cc37a932/x/tokenregistry/types/msgs_test.go#L11-L11)
> 
> [http://github.com/Sifchain/sifnode/blob/021dd1c1a7ca9f95aeac14208a6c9f68cc37a932/x/tokenregistry/handler/handler_test.go#L121-L121](https://github.com/Sifchain/sifnode/blob/021dd1c1a7ca9f95aeac14208a6c9f68cc37a932/x/tokenregistry/handler/handler_test.go#L121-L121)

Sure. I will do that. Thanks for catching that.

PR #2358, opened 2022-01-14
Feature/ethbridge admin

Jedi2002:


x/ethbridge/client/cli/tx.go
```diff
@@ -392,19 +392,19 @@ func GetCmdSetBlacklist() *cobra.Command {
...
 
-			err = json.Unmarshal(contents, &msg)
+			err = json.Unmarshal(contents, &addresses)
 			if err != nil {
 				return err
 			}
 
 			msg.From = clientCtx.FromAddress.String()
-
+			msg.Addresses = addresses
```
timlind, line 407: There is already a script that prepares the JSON in this format so this should stay as is.

juniuszhou (approved):

timlind:
Can include an RC handler for TestNet

intl-man (approved):

PR #2365, opened 2022-01-17
Marsifi/onboarding week1

marsifi:
This PR contains a few minor changes to the clp-tutorial which had some outdated commands..


x/clp/keeper/msg_server.go
```diff
@@ -349,9 +350,10 @@ func (k msgServer) RemoveLiquidity(goCtx context.Context, msg *types.MsgRemoveLi
 
 func (k msgServer) CreatePool(goCtx context.Context, msg *types.MsgCreatePool) (*types.MsgCreatePoolResponse, error) {
 	ctx := sdk.UnwrapSDKContext(goCtx)
+
 	// Verify min threshold
-	MinThreshold := sdk.NewUintFromString(types.PoolThrehold)
-	if msg.NativeAssetAmount.LT(MinThreshold) { // Need to verify
```
marsifi, line 354: This is where the hardcoded PoolThreshold was being used

timlind:
Thanks @marsifi. Can you put the param change into a it's own PR so that we can handle them separately and get this one merged in?

marsifi:
@timlind I removed the param changes

x/clp/keeper/msg_server.go
Jedi2002: hi @marsifi  , This number was decided by the product team as the min amount of rowan required to create a pool . At that time we decided not to apply an limits on the external asset

Jedi2002 (approved):

PR #2372, opened 2022-01-18
Fixes script that will check ibc client states

brianosaurus:
Uses consensus state to check clients rather than the revision height which is the wrong way (I found out).


gzukel (approved):
approved

PR #2373, opened 2022-01-19
First pass with SDK v45

marsifi:
Investigating upgrade to Cosmos SDK v45


brianosaurus:
:+1: I ran IBC tests and it looks good. Needs `go mod tidy` to fixup modules however.

faddat:
Unless in a giant hurry, it would make more sense to use this as the v10.0.0

marsifi:
moved here: https://github.com/Sifchain/sifnode/pull/2568

PR #2376, opened 2022-01-19
Closes #2370 by looking at pagination

### LINKED ISSUE #2370: Fix cli q ibc client states
Not all the time but periodically 
```
sifnoded q ibc client states --node $testnet
```

Won't return the full list of connected clients' states. This causes confusion when trying to quickly inspect a client to find client_id's, connection id's, etc... If someone looks at all of the connections they can see the list of connected clients is larger than the clients returned from `client states`.

Fix and submit a PR to ibc-go to resolve the issue.

brianosaurus:
The issue is that `sifnoded q client states --node $testnet` doesn't return all connected clients. It was breaking scripts that look for expired clients etc...

The fix was to mind the next_page attribute of pagination in the output.

Closes #2370


gzukel (approved):
approved

PR #2395, opened 2022-01-25
rocksdb and state sync

faddat:
non-state breaking siffie upgrade that supports rocksdb and state sync


faddat:
merge this into v0.9.14

PR #2418, opened 2022-02-01
One set of messages for all margin position types

AustinoBombino:


x/margin/types/msgs_test.go
```diff
@@ -203,34 +203,34 @@ func TestTypes_MsgCloseLongValidateBasic(t *testing.T) {
...
-			errString:    "empty address string is not allowed",
+			name:      "no signer",
+			msgClose:  types.MsgClose{},
+			errString: "empty address string is not allowed",
 		},
 		{
 			name: "wrong address",
-			msgCloseLong: types.MsgCloseLong{
+			msgClose: types.MsgClose{
 				Signer: "xxx",
```
canercandan, line 220: @AustinoBombino here we need to set the new field `position` to `long`

x/margin/keeper/msg_server.go
```diff
@@ -23,13 +23,109 @@ func NewMsgServerImpl(k types.Keeper) types.MsgServer {
...
+}
+
+func (k msgServer) Close(goCtx context.Context, msg *types.MsgClose) (*types.MsgCloseResponse, error) {
+	ctx := sdk.UnwrapSDKContext(goCtx)
+
+	var mtp *types.MTP
+	var err error
+
+	switch msg.Position {
+	case "long":
```
timlind, line 65: would suggest a proto enum here

timlind (approved):

PR #2423, opened 2022-02-01
Feature/margin 1 addup amount existing mtp

canercandan:
@AustinoBombino @timlind you can find here the use case where the user adds up a certain amount to an existing asset long MTP

We had some discussions over having several MTPs created for the same asset, some I am not quite sure if I should rather except to a list of MTPs for a given asset rather than getting a single one.

So far the test obviously overrides the existing MTP for the used asset.


canercandan:
changes have been moved to @timlind PR https://github.com/Sifchain/sifnode/pull/2426 cc @AustinoBombino

PR #2426, opened 2022-02-02
Feature/margin 1 multiple positions

timlind:


canercandan:
@timlind I fixed the broken test cases issues in msg server test file, I have also added unit tests for force close although the later seems to return an `mtp not found` error when trying to force close an existing mtp position cc @AustinoBombino

x/margin/keeper/msg_server_test.go
```diff
@@ -387,6 +430,239 @@ func TestKeeper_Close(t *testing.T) {
...
+			var signer string = msg.Signer
+			if tt.overrideSigner != "" {
+				signer = tt.overrideSigner
+			}
+
+			if !tt.mtpCreateDisabled {
+				addMTPKey(t, ctx, app, marginKeeper, tt.msgOpen.CollateralAsset, tt.msgOpen.BorrowAsset, signer, tt.msgOpen.Position, 1)
+			}
+
+			_, got := msgServer.ForceClose(sdk.WrapSDKContext(ctx), &msg)
```
timlind, line 653: @canercandan the MTP that is closed uses msg.MtpAddress, not msg.Signer, so the above overrides need to apply to MtpAddress as well.

canercandan: @timlind thanks for your review, I fixed this issue in the last commit I pushed and added additional test cases with different values for force close threshold, things are working as expected.

I did notice though that the force close threshold param is not initiated here: https://github.com/Sifchain/sifnode/blob/feature/margin-1-multiple-positions/x/margin/types/genesis.go#L17

cc @AustinoBombino

marshsif:
Changes pushed, waiting for review from @timlind

PR #2427, opened 2022-02-02
Future/peggy2 fix smartcontract tests

Brando753:


smartyalgo:
For context, whats the reason to include those fields?

juniuszhou (approved):

x/ethbridge/types/msgs.go
```diff
@@ -224,12 +224,12 @@ func ComputeProphecyID(cosmosSender string, sequence uint64, ethereumReceiver st
 	bytes, _ := arguments.Pack(
```
smartyalgo, line 224: I think this'll fail with "argument count mismatch", since argument at L197 wasnt updated

Brando753: Great catch Michael. I have updated that section accordingly as well as fixed the unit test for msgs.go

Brando753:
Going to merge this for now, pending review if the prophecy id is needed at all.

PR #2441, opened 2022-02-07
Future/peggy2 commission erc20

Brando753:
This adds support for commission charging tokens along with new test cases. Hardhat will require a call to compile before running tests because of a typechain dependency import. I am unsure how to resolve this.


smart-contracts/contracts/BridgeBank/BridgeBank.sol
```diff
@@ -434,13 +434,57 @@ contract BridgeBank is BankStorage, CosmosBank, EthereumWhiteList, CosmosWhiteLi
       contractDecimals[token] = _decimals;
     } catch {
       // if we can't access the decimals function of this token,
-      // assume that it has 18 decimals
-      decimals = 18;
+      // assume that it has 0 decimals
+      decimals = 0;
```
banshee, line 438: I think we also need to modify the section above (where we check decimals > 0)

smart-contracts/contracts/BridgeBank/BridgeBank.sol
```diff
@@ -419,9 +419,9 @@ contract BridgeBank is BankStorage, CosmosBank, EthereumWhiteList, CosmosWhiteLi
   }
 
   /**
-   * @dev Fetches de decimals of a token by address
+   * @dev Fetches the decimals of a token by address
    * @param token The bridgeTokens's address
-   * @return The bridgeTokens's decimals or 18
+   * @return The bridgeTokens's decimals or 0
```
banshee, line 424: Do we need getDecimals at all?

smart-contracts/contracts/BridgeBank/BridgeBank.sol
```diff
@@ -434,13 +434,57 @@ contract BridgeBank is BankStorage, CosmosBank, EthereumWhiteList, CosmosWhiteLi
...
+    tokenToTransfer.safeTransferFrom(msg.sender, address(this), amount);
+
+    //Fetch the updated balance reported after the transfer
+    uint256 newBalance = getBalance(token);
+
+    //Calculate the total amount transfered from the newbalance vs the old balance
+    //Since this contract uses solidity 0.8+ overflows from bad acting tokens should 
+    //revert.
+    uint256 transferedAmount = newBalance - oldBalance;
+
```
banshee, line 484: Add the require > 0 here instead of in the callers?

smart-contracts/test/test_bridgeBankLock.js
```diff
@@ -99,6 +99,39 @@ describe("Test Bridge Bank", function () {
...
+      const devAccount = userOne.address;
+      const user = userTwo;
+      const devFee = 500; // 5% dev fee
+      const initialBalance = 100_000
+      const token = await deployCommissionToken(devAccount, devFee, user.address, initialBalance);
+      const transferAmount = 10_000
+      const remaining = 90_000;
+      const afterTransfer = 9_500;
+      // Get balances before locking
+      const beforeBridgeBankBalance = Number(
```
banshee, line 112: BigNumber instead of Number?

banshee (changes requested):

smart-contracts/contracts/BridgeBank/BridgeBank.sol
Brando753: contractDecimals[token] is a mapping of token addresses to stored decimals from previous lookups. Since the default uint8 type will be 0, we must always check if it's something different than 0. If it's 0, it has two possible values, either 0 or not fetched, so we reach out to the contract and query the correct decimal number. If reaching out fails for any reason, we default to 0; otherwise, we return the result. I believe this is the correct way to handle it; let me know if not.

smart-contracts/contracts/BridgeBank/BridgeBank.sol
Brando753: That structure of events expects them and a lot of the architecture already written expects them. Whether the current system needs them or not is something that could be evaluated, I assume not, but at this point, I would rather keep it in than rearchitect this area for MVP. I am trying to make the contract changes be as minimally invasive as possible.

smart-contracts/contracts/BridgeBank/BridgeBank.sol
Brando753: I was thinking about that as well, I wanted to have a more generic transfer function that does not make assumptions about what has to happen after the transfer occurs and so I coded it this way. That being said if there is never a time in our code we would not want that check then it makes more sense to include it in the transfer function. Do you think that's the best?

smart-contracts/contracts/BridgeBank/BridgeBank.sol
banshee: I meant the function itself - just call decimals on the contract instead of the function, as a way to reduce contract size.  Looks like we don't need it though.

banshee (approved):

juniuszhou (approved):

smart-contracts/contracts/BridgeBank/BridgeBank.sol
jzvikart: Could there be any issues with concurrency here?

smart-contracts/contracts/Mocks/CommissionToken.sol
```diff
@@ -0,0 +1,43 @@
...
+
+import "@openzeppelin/contracts/token/ERC20/ERC20.sol";
+
+/**
+ * @title Commission Token
+ * @dev This token will charge a programmable fee on every transfer that is credited to the the devolpoers address
+ *      this test token will help test various tokens that may transfer a different amount then the transfer requests
+ *      to the recipient.
+ **/
+contract CommissionToken is ERC20 {
```
jzvikart, line 12: We should probably have an integration test for this?

jzvikart (approved):

smartyalgo (commented):
Added comments

smart-contracts/contracts/BridgeBank/BridgeBank.sol
Brando753: @jzvikart As I understand it, there is no concurrency on standard EVM implementations such that it should not be a concern. If I am wrong please let me know, this is why I wanted as many eyes on the code as possible.

Brando753: @smartyalgo Unlock operations should be prevented on the ethbridge layer for values of 0 since those are costs incurred by the relayer. We should never revert on a valid call to unlock as that could cause problems with the ordering. As for the unlock event, an unlock request was received for and transferred out of n units, I don't know if we would want to show an event for less than what the ethbridge module created a claim for.

smart-contracts/contracts/Mocks/CommissionToken.sol
Brando753: We can, a commission token will have the same flows like any other ERC20 token in our codebase so long as the events emitted are the same, which I can extend the commission test to check for. But more testing never hurts.

PR #2448, opened 2022-02-08
Develop upgrade bridgebank blocklist

banshee:
This is already in production and should be merged immediately.


banshee:
Need to close this because we merged develop into the branch.

PR #2461, opened 2022-02-14
add register all

### LINKED ISSUE #2460: Re-implement register all as single transaction
The previous implementation of the register all includes a loop and generates one transaction for each item in JSON file. we need to change it, use one transaction to add all items.

juniuszhou:
The PR resolves #2460.


x/tokenregistry/keeper/keeper.go
```diff
@@ -165,3 +165,46 @@ func (k keeper) SetFirstLockDoublePeg(ctx sdk.Context, denom string, networkDesc
...
 }
+
+func (k keeper) AddMultipleTokens(ctx sdk.Context, entries []*types.RegistryEntry) {
+	wl := k.GetRegistry(ctx)
+
+	for _, entry := range entries {
+		found := false
+		for i := range wl.Entries {
+			// update if nil or the same denom
+			if wl.Entries[i] != nil && strings.EqualFold(wl.Entries[i].Denom, entry.Denom) {
```
jp-sif, line 176: Is it compulsory to have to write all this logic manually in Go? It is a common task to compare a parameter of an element in a list of things against a key in order to avoid inserting a duplicate. Also, if duplicates are unwanted, why not make `wl.Entries` into a map keyed by `Denom`? Does Go have a collections library for this sort of thing?

These lists better not be too large either, because this is `O(m * n)`.

juniuszhou: a map was implemented before, and solution was reverted after the last merge with develop branch. will recover the solution later.

Brando753 (changes requested):
Still waiting on maps to be added back to this PR

proto/sifnode/tokenregistry/v1/tx.proto
```diff
@@ -21,6 +23,13 @@ message MsgRegister {
 
 message MsgRegisterResponse {}
 
+message MsgRegisterAll {
+  string from = 1;
```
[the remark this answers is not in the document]
juniuszhou: will add the comments.

x/tokenregistry/types/msgs.go
```diff
@@ -60,6 +60,59 @@ func (m *MsgRegister) GetSigners() []sdk.AccAddress {
...
+		}
+		if !coin.IsValid() {
+			return errors.New("Denom is not valid")
+		}
+		_, err := sdk.AccAddressFromBech32(m.From)
+		if err != nil {
+			return sdkerrors.Wrap(err, "invalid from address")
+		}
+		if entry.Decimals <= 0 {
+			return errors.New("Decimals cannot be zero")
```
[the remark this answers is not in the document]
juniuszhou: will update it.

x/tokenregistry/keeper/keeper.go
```diff
@@ -84,63 +84,70 @@ func (k keeper) GetEntry(wl types.Registry, denom string) (*types.RegistryEntry,
...
 	store := ctx.KVStore(k.storeKey)
-	bz := k.cdc.MustMarshal(&wl)
-	store.Set(types.WhitelistStorePrefix, bz)
+	store.Delete(key)
 }
 
+// GetRegistry get all token's metadata
 func (k keeper) GetRegistry(ctx sdk.Context) types.Registry {
-	var whitelist types.Registry
+	var entries []*types.RegistryEntry
```
Brando753, line 123: This won't work for Peggy 2.0. It loops over every token correct? If so you have to assume there are millions of tokens imported so it must support pagination.

Brando753 (changes requested):
New ticket #2550 created in regards to GetRegistry

x/tokenregistry/keeper/keeper.go
```diff
@@ -84,63 +84,70 @@ func (k keeper) GetEntry(wl types.Registry, denom string) (*types.RegistryEntry,
...
-		if t != nil && !strings.EqualFold(t.Denom, denom) {
-			updated = append(updated, t)
-		}
-	}
-	k.SetRegistry(ctx, types.Registry{
-		Entries: updated,
-	})
+// SetToken add a new denom
+func (k keeper) SetToken(ctx sdk.Context, entry *types.RegistryEntry) {
+	entry.Sanitize()
```
banshee, line 105: Why are we calling Sanitize now?

banshee, line 105: Why does Sanitize() modify the struct that's passed in?  Seems like it should be modifying a copy

Brando753: Should be fine right, its just trimming the ends of the string for '/'
```go
func (r *RegistryEntry) Sanitize() {
	r.Path = strings.Trim(r.Path, "/")
}
```

Brando753: That may be a bug, would make sense to modify a copy. That being said AddTokenMetadata calls this and it does not appear to cause any issues if its trimmed. msg_server Register calls this where it also should not be a problem.

juniuszhou: Ok, will use the copy instead of pointer.

x/tokenregistry/keeper/keeper.go
juniuszhou: so we use this pr to fix new ticket, or create new one?

Brando753: I created a new ticket and am going to take a look at it.

Brando753: Ticket is #2550 we can merge once everything else is done, will make this a separate PR.

Brando753 (approved):
Once merge conflicts are resolved this may be merged in.

PR #2473, opened 2022-02-18
test: 💍 pmtp integration test env

canercandan:
- [ ] use a constant mnemonic to prepare the genesis state when used by testutil/simapp
- [ ] create the bank genesis state with a rowan balance attached to same validator address

next:
- [ ] currently using cosmos sdk v0.45.1, make it work with v.0.43.0
- [ ] add additional tokens
- [ ] add CLPs


canercandan:
Updates:

* the PR includes also a documentation about the native integration test environment used and is a good walk through to understand and create new integration test suites with customized and pre-defined genesis states.

AustinoBombino (approved):
@canercandan looks great, nice job!

I would maybe add some functions to the test util for getting the pmtp parameter set so we can have tests verifying that they are updating correctly

PR #2484, opened 2022-02-24
Update burn denom

### LINKED ISSUE #2483: Identify denom from burn event in EVM
before token burned in Ethereum, we already have the denom generated in sifnode.
for instance, the native rowan or ibc token. the denom hash build from network and
contract address can't be taken as denom. need to get denom from token registry keeper

juniuszhou:
The pr resolves #2483

### DIFF
--- a/x/tokenregistry/keeper/keeper.go
+++ b/x/tokenregistry/keeper/keeper.go
@@ -165,3 +165,16 @@ func (k keeper) SetFirstLockDoublePeg(ctx sdk.Context, denom string, networkDesc
 		instrumentation.PeggyCheckpoint(ctx.Logger(), instrumentation.SetFirstLockDoublePeg, "networkDescriptor", networkDescriptor, "registry", registry)
 	}
 }
+
+// TODO get the denom temporarily, will add a map to keep the data from network+address to denom
+// after confirmed we really need this to identify the denom after receive the burn event from Ethereum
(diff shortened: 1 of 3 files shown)


jp-sif:
I read this, it looks OK. I will let someone with more experience with the codebase approve it though.

x/tokenregistry/keeper/keeper.go
```diff
@@ -165,3 +165,16 @@ func (k keeper) SetFirstLockDoublePeg(ctx sdk.Context, denom string, networkDesc
 		instrumentation.PeggyCheckpoint(ctx.Logger(), instrumentation.SetFirstLockDoublePeg, "networkDescriptor", networkDescriptor, "registry", registry)
 	}
 }
+
+// TODO get the denom temporarily, will add a map to keep the data from network+address to denom
+// after confirmed we really need this to identify the denom after receive the burn event from Ethereum
+func (k keeper) GetDenomFromContract(ctx sdk.Context, networkDescriptor oracletypes.NetworkDescriptor, contract string) (string, error) {
+	entries := k.GetRegistry(ctx)
+	for _, entry := range entries.Entries {
```
jp-sif, line 173: Is it possible to have an attack vector where this list is too long to traverse efficiently?

juniuszhou: It is good question. currently, I am not 100% sure we should get the denom this way. After confirmed, I will create a map in the keeper to store (network, address) to denom.

juniuszhou:
Denom could be got from ethereum event, the pr is unnecessary, close it.

PR #2503, opened 2022-03-03
feat: print channel and chain id data

timlind:
This tool is useful when whitelisting new ibc tokens. It shows all existing channels with their chain ids.


marsifi (approved):
Looks good to me.
What do you think of adding this to the `ibc-diag` command in sifnoded?

timlind:
> Looks good to me.
> What do you think of adding this to the `ibc-diag` command in sifnoded?

That's worth considering but this was built using the JSON api so I think it's best kept as a script.

PR #2507, opened 2022-03-07
Pmtp Begin Block Basic functionality

Jedi2002:

PR #2508, opened 2022-03-07
feat: on-chain rewards

timlind:


proto/sifnode/clp/v1/params.proto
```diff
@@ -6,4 +6,16 @@ import "gogoproto/gogo.proto";
...
+  repeated RewardPeriod reward_periods = 3;
+}
+
+message RewardPeriod {
+  uint64 start_block = 1;
+  uint64 end_block = 2;
+  string allocation = 3 [
+    (gogoproto.customtype) = "github.com/cosmos/cosmos-sdk/types.Uint"
+  ];
+}
```
AustinoBombino, line 21: @timlind @Jedi2002 Having a type for periods seems clean, we could do this for PMTP also: https://github.com/Sifchain/sifnode/pull/2507/files

with startblock, endblock, and epochlength

AustinoBombino, line 21: @timlind @Jedi2002 Do you guys think we should use a type for PMTP period params also? Looks like this is here to allow for multiple periods at once but it may be good to do on PMTP also to keep things uniform?
https://github.com/Sifchain/sifnode/pull/2507/files

AustinoBombino (commented):
Looking good! No issues that I can see

x/clp/keeper/rewards.go
```diff
@@ -0,0 +1,258 @@
...
+		value := it.Value()
+		var lp types.LiquidityProvider
+		if len(value) <= 0 {
+			continue
+		}
+		err := k.cdc.Unmarshal(value, &lp)
+		if err != nil {
+			continue
+		}
+		k.PruneUnlockRecords(ctx, lp, params.LiquidityRemovalLockPeriod, params.LiquidityRemovalCancelPeriod)
```
Jedi2002, line 24: Can we move this function call to `Remove Liquidity` and `Swap`.The logic being 
- User requests remove liquidity/swap 
- Calculate lpuints
- Expire Unlocks 
- Check if user has enough lpuints 

The double for loop at begin block might be quite heavy , considering the number of lp's, Since an LP is unique to address + asset

timlind: That's a good point. The advantage of doing it here is that we can trigger an event with the right cancellation block height. I think we can probably relax that requirement esp because UI is not showing a history from data services anymore, but we should chat to the team to let them know.

timlind: In that case UseUnlockedLiquidity should also ensure the record is not passed a cancel period.

x/clp/keeper/rewards.go
```diff
@@ -0,0 +1,188 @@
...
+		if height >= period.StartBlock && height <= period.EndBlock {
+			return period
+		}
+	}
+	return nil
+}
+
+func (k Keeper) DistributeDepthRewards(ctx sdk.Context, period *types.RewardPeriod, pools []*types.Pool) error {
+	periodLength := period.EndBlock - period.StartBlock + 1
+	blockDistribution := period.Allocation.QuoUint64(periodLength)
```
Jedi2002, line 55: Can we move this calculation out of this function , and do it only once at period.StartBlock

x/clp/types/params.go
```diff
@@ -15,7 +15,10 @@ const (
 
 // Parameter store keys
 var (
-	KeyMinCreatePoolThreshold = []byte("MinCreatePoolThreshold")
+	KeyMinCreatePoolThreshold       = []byte("MinCreatePoolThreshold")
+	KeyLiquidityRemovalLockPeriod   = []byte("LiquidityRemovalLockPeriod")
+	KeyLiquidityRemovalCancelPeriod = []byte("LiquidityRemovalCancelPeriod")
+	KeyRewardPeriods                = []byte("RewardPeriods")
```
Jedi2002, line 21: Does the whole list need to be passed, every time we want to append a rewards period?

timlind: Yes. Are you thinking of another way it could be done?

x/clp/keeper/rewards.go
AustinoBombino: Yeah I can see how iterating through every LP could be heavy and possibly an attack vector to slow down the network even.

Triggering it on other address actions seems like a good workaround to me.

PR #2531, opened 2022-03-11
Feature/dispensation record param

Jedi2002:


x/dispensation/keeper/distributionRecords_test.go
```diff
@@ -105,7 +105,7 @@ func TestKeeper_GetPendingRecordsLimited(t *testing.T) {
 		_, err = keeper.GetDistributionRecord(ctx, name, rec.Address, types.DistributionStatus_DISTRIBUTION_STATUS_COMPLETED, record.DistributionType)
 		assert.NoError(t, err)
 	}
-	assert.Len(t, keeper.GetLimitedRecordsForStatus(ctx, types.DistributionStatus_DISTRIBUTION_STATUS_PENDING).DistributionRecords, 10)
+	assert.Len(t, keeper.GetLimitedRecordsForStatus(ctx, types.DistributionStatus_DISTRIBUTION_STATUS_PENDING).DistributionRecords, 20)
```
marshsif, line 108: Should 20 be a constant variable rather than just hard coded here?

x/dispensation/keeper/executors_test.go
```diff
@@ -63,7 +63,7 @@ func TestKeeper_DistributeDrops_For_Address_Fail(t *testing.T) {
 	runner := ""
 	err = keeper.CreateDrops(ctx, outputList, distributionName, types.DistributionType_DISTRIBUTION_TYPE_AIRDROP, runner)
 	assert.NoError(t, err)
-	_, err1 := keeper.DistributeDrops(ctx, 4657424885079777562, distributionName, runner, types.DistributionType_DISTRIBUTION_TYPE_AIRDROP)
+	_, err1 := keeper.DistributeDrops(ctx, 4657424885079777562, distributionName, runner, types.DistributionType_DISTRIBUTION_TYPE_AIRDROP, 10)
```
marshsif, line 66: should these lines with 10 instead refer to a constant variable, say DEFAULT_DISTRIBUTION_COUNT?

x/dispensation/keeper/executors.go
```diff
@@ -31,8 +31,13 @@ func (k Keeper) CreateDrops(ctx sdk.Context, output []banktypes.Output, name str
 
 // DistributeDrops is called at the beginning of every block .
 // It checks if any pending records are present , if there are it completes the top 10
```
arrivets, line 33:
```suggestion
// It checks if any pending records are present , if there are it completes the top 'distributionCount' items
```

arrivets:
We should test the a valid `DistributionCount` value here

https://github.com/Sifchain/sifnode/blob/3cbe086dcac3f4e2b11132e48954db4ec35326cb/x/dispensation/types/msgs_test.go#L385-L397

and create a new test with an invalid value (> Max)

x/dispensation/keeper/executors.go
```diff
@@ -31,8 +31,13 @@ func (k Keeper) CreateDrops(ctx sdk.Context, output []banktypes.Output, name str
 
 // DistributeDrops is called at the beginning of every block .
```
arrivets, line 32: Is it actually run at the beginning of every block? I couldn't find that logic.

arrivets (changes requested):

marsifi (dismissed):
The integration tests are not passing. Things specific to dispensation

x/dispensation/keeper/executors.go
Jedi2002: No its not , that was the older design , ill remove the comment

x/dispensation/keeper/executors_test.go
Jedi2002: We don't have a DEFAULT_DISTRIBUTION_COUNT under the design , the number 10 is just the value we had earlier, to help the tests pass.
There is going to be a separate PR with unit tests, ill try to include the changes related to tests in that

Jedi2002:
> The integration tests are not passing. Things specific to dispensation

@sheokapr Is working on fixing those

AustinoBombino (approved):

PR #2542, opened 2022-03-16
add admin for registry

juniuszhou:


cmd/sifnoded/cmd/tokenregistry.go
```diff
@@ -0,0 +1,75 @@
...
+`,
+		Args: cobra.ExactArgs(1),
+		RunE: func(cmd *cobra.Command, args []string) error {
+			clientCtx := client.GetClientContextFromCmd(cmd)
+			cdc := clientCtx.Codec
+			serverCtx := server.GetServerContextFromCmd(cmd)
+			config := serverCtx.Config
+			config.SetRoot(clientCtx.HomeDir)
+			addr, err := sdk.AccAddressFromBech32(args[0])
+			if err != nil {
```
banshee, line 35: Why do we need this?  Can't we just require an address?

banshee (approved):

cmd/sifnoded/cmd/tokenregistry.go
juniuszhou: We need the context, then can write the address into json config file.

PR #2543, opened 2022-03-16
update denom

juniuszhou:


x/ethbridge/keeper/msg_server.go
```diff
@@ -53,9 +54,13 @@ func (srv msgServer) Lock(goCtx context.Context, msg *types.MsgLock) (*types.Msg
 		return &types.MsgLockResponse{}, fmt.Errorf("token metadata not available for %s", msg.DenomHash)
 	}
 
-	doublePeg := tokenMetadata.NetworkDescriptor != msg.NetworkDescriptor
-	firstDoublePeg := doublePeg && srv.tokenRegistryKeeper.GetFirstLockDoublePeg(ctx, msg.DenomHash, msg.NetworkDescriptor)
+	doublePeg := true
+	// all ibc tokens lock is double peg, burns via ibc token transfer message
+	if !strings.HasPrefix(msg.DenomHash, "ibc/") {
+		doublePeg = tokenMetadata.NetworkDescriptor != oracletypes.NetworkDescriptor_NETWORK_DESCRIPTOR_UNSPECIFIED && tokenMetadata.NetworkDescriptor != msg.NetworkDescriptor
```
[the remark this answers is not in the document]
Brando753: Shouldn't all this logic for double pegging belong in the burn functions? For example lets lock ETH from Ethereum to Sifchain, then we burn it on sifchain to say Polygon so the only IOU exists on polygon. They can burn that IOU for it to be minted on Sifchain which can be burned to get it back on ethereum.

Brando753 (changes requested):
Pretty sure the logic for first time double pegging needs to be part of the burn logic not the lock logic.

cmd/ebrelayer/relayer/ethereum.go
```diff
@@ -393,8 +393,7 @@ func (sub EthereumSub) logToEvent(networkDescriptor oracletypes.NetworkDescripto
 		event.ClaimType = ethbridgetypes.ClaimType_CLAIM_TYPE_BURN
 		event.To = append(event.To, decodedEvent.To...)
 		// burn doesn't have symbol or name
-		//event.Symbol = decodedEvent.Symbol
-		//event.Name = decodedEvent.Name
+		event.Symbol = decodedEvent.Denom
```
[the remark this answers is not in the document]
juniuszhou: Ok, will add a new field in proto.

x/ethbridge/keeper/msg_server.go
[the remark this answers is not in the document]
juniuszhou: If the network descriptor is not unspecified, it means the token issued from sifnode, like rowan.

proto/sifnode/ethbridge/v1/tx.proto
```diff
@@ -134,10 +134,10 @@ message MsgSetFeeInfo {
     (gogoproto.nullable) = false,
     (gogoproto.moretags) = "yaml:\"minimum_burn_cost\""
   ];
-  string first_lock_double_peggy_cost = 7 [
+  string first_burn_double_peggy_cost = 7 [
```
banshee, line 137: This is a bit strange, since we want to charge the fee on any export to an ethereum chain.  That could be either a lock or a burn.

x/ethbridge/keeper/keeper.go
```diff
@@ -82,8 +82,8 @@ func (k Keeper) ProcessSuccessfulClaim(ctx sdk.Context, claim *types.EthBridgeCl
 		coins = sdk.NewCoins(sdk.NewCoin(claim.Denom, claim.Amount))
 		err = k.bankKeeper.MintCoins(ctx, types.ModuleName, coins)
 	case types.ClaimType_CLAIM_TYPE_BURN:
-		coins = sdk.NewCoins(sdk.NewCoin(claim.Denom, claim.Amount))
-		err = nil
+		coins = sdk.NewCoins(sdk.NewCoin(claim.CosmosDenom, claim.Amount))
```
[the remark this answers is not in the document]
juniuszhou: will add comments

proto/sifnode/ethbridge/v1/tx.proto
juniuszhou: This fee just charged for creating a new contract on evm based blockchain. According to discussion before, we will use burn for "burning" imported token regradless the destination is back to original network or new network. So the cose just for burn case. Sifchain native token and ibc tokens, we need deploy contracts manually.

cmd/ebrelayer/txs/parser.go
```diff
@@ -73,7 +73,7 @@ func EthereumEventToEthBridgeClaim(valAddr sdk.ValAddress, event types.EthereumE
 	witnessClaim.TokenName = event.Name
 	// the nonce from ethereum event is lock burn nonce, not transaction nonce
 	witnessClaim.Denom = ethbridge.GetDenom(networkDescriptor, tokenContractAddress)
```
smartyalgo, line 75: Is there a reason why we want to create a new field CosmosDenom, instead of using Denom field?

juniuszhou: It is from Brando's comment, a new field will be much clear, Denom for lock and CosmosDemon for burn. 

[Brando753](https://github.com/Brando753) [6 days ago](https://github.com/Sifchain/sifnode/pull/2543#discussion_r829558342)
Should we add a field labeled CosmosDenom in the protofiles rather then mapping the cosmos denom to Symbol?

x/ethbridge/keeper/msg_server.go
```diff
@@ -131,18 +127,25 @@ func (srv msgServer) Burn(goCtx context.Context, msg *types.MsgBurn) (*types.Msg
 		return nil, fmt.Errorf("token metadata not available for %s", msg.DenomHash)
 	}
 
-	prophecyID, err := srv.Keeper.ProcessBurn(ctx, cosmosSender, account.GetSequence(), msg, tokenMetadata)
+	doublePeg := tokenMetadata.NetworkDescriptor != msg.NetworkDescriptor
```
Brando753, line 130: Isn't this always true for burns?

x/ethbridge/keeper/msg_server.go
```diff
@@ -131,18 +127,25 @@ func (srv msgServer) Burn(goCtx context.Context, msg *types.MsgBurn) (*types.Msg
...
 		return nil, err
 	}
 
-	logger.Info("sifnode emit burn event.", "message", msg)
-	globalSequence := srv.Keeper.GetGlobalSequence(ctx, msg.NetworkDescriptor)
 	srv.Keeper.UpdateGlobalSequence(ctx, msg.NetworkDescriptor, uint64(ctx.BlockHeight()))
 
-	doublePeg := tokenMetadata.NetworkDescriptor != msg.NetworkDescriptor
+	if firstDoublePeg {
+		srv.tokenRegistryKeeper.SetFirstLockDoublePeg(ctx, msg.DenomHash, msg.NetworkDescriptor)
```
Brando753, line 144: Should this function be renamed to FirstBurnDoublePeg or just FirstDoublePeg?

Brando753 (changes requested):
Minor questions, I am going to merge in future/peggy2 to resolve some merge conflicts shortly. Once you answer those questions or make changes we can merge.

x/ethbridge/keeper/msg_server.go
juniuszhou: If the token burned and go back to its orignal network, it is not double peg.

x/ethbridge/keeper/msg_server.go
juniuszhou: FirstDoublePeg is fine since double peg only happened when token burned. will update it.

x/ethbridge/keeper/msg_server.go
Brando753: right, don't know what I was thinking.

Brando753 (approved):
Looks ready to go

PR #2551, opened 2022-03-23
Feature/pmtp unit test coverage

canercandan:
~81% code coverage


AustinoBombino (approved):
@canercandan These tests look great!

One thing to add would be cases for swaps and liquidity removes that have `pmtpCurrentRunningRate`'s other than 1.0 so we can make sure the calcs are behaving as expected.

We'll need to get/come up with sample values for those cases.

canercandan:
> @canercandan These tests look great!
> 
> One thing to add would be cases for swaps and liquidity removes that have `pmtpCurrentRunningRate`'s other than 1.0 so we can make sure the calcs are behaving as expected.
> 
> We'll need to get/come up with sample values for those cases.

@AustinoBombino just added more use cases to cover behavior and expected outputs for all the possible values of `pmtpCurrentRunningRate` between `0.0` to `1.0` with `0.1` increments, let me know if that what you had in mind or if there are any other assessments that need to be done. Thanks for your feedback

marshsif:
pmtpCurrentRunningRate range should be 0 thru 10, decimal format.

canercandan:
@AustinoBombino just updated the branch cover all the values for pmtp current running rate from 0 to 10 with 1.0 increments.

AustinoBombino:
@canercandan nice! looks good to me

PR #2568, opened 2022-03-29
Feature/v45

marsifi:
Updating to Cosmos SDK `v0.45.1`


app/app.go
```diff
@@ -493,17 +530,21 @@ func NewSifApp(
...
+		ibchost.ModuleName,
+		disptypes.ModuleName,
+		transferModule.Name(),
 		clptypes.ModuleName,
+		ethbridgetypes.ModuleName,
+		tokenregistrytypes.ModuleName,
 		oracletypes.ModuleName,
-		ethbridge.ModuleName,
-		dispensation.ModuleName,
-		tokenregistry.ModuleName,
```
AustinoBombino, line 505: What's the reason for removing these modules from the InitGenesis order?

app/app.go
```diff
@@ -472,17 +475,51 @@ func NewSifApp(
...
+		authtypes.ModuleName,
+		banktypes.ModuleName,
+		govtypes.ModuleName,
+		crisistypes.ModuleName,
+		genutiltypes.ModuleName,
+		feegrant.ModuleName,
+		paramstypes.ModuleName,
+		vestingtypes.ModuleName,
 		ibchost.ModuleName,
-		dispensation.ModuleName,
```
AustinoBombino, line 476: Why is this being removed?

x/ethbridge/keeper/blacklist.go
```diff
@@ -59,9 +59,11 @@ func (k Keeper) SetBlacklist(ctx sdk.Context, msg *types.MsgSetBlacklist) error
 
 func (k Keeper) GetBlacklist(ctx sdk.Context) []string {
 	var addresses []string
+
 	store := ctx.KVStore(k.storeKey)
-	iter := store.Iterator(types.BlacklistPrefix, nil)
+	iter := sdk.KVStorePrefixIterator(store, types.BlacklistPrefix)
```
AustinoBombino, line 64: Does store.Iterator() need to be replaced throughout? I believe it also appears in the oracle keepr

app/app.go
marsifi: It's still there, under `disptypes.ModuleName`

app/app.go
marsifi: They are still there: under `ethbridgetypes.ModuleName`, `disptypes.ModuleName`, and `tokenregistrytypes.ModuleName`.

`ibctransfertypes.ModuleName` is still there too, as `transferModule.Name()`

x/ethbridge/keeper/blacklist.go
marsifi: Good point, I updated it in the oracle keeper as well, and extended the relevant unit test.
It doesn't appear anywhere else.

Jedi2002:
Closing in favour of newer pull request 
[https://github.com/Sifchain/sifnode/pull/2819](url)

PR #2569, opened 2022-03-29
Feature/pmtp begin blocker tests

canercandan:


x/clp/abci.go
```diff
@@ -24,7 +25,8 @@ func BeginBlocker(ctx sdk.Context, k keeper.Keeper) {
 		k.PolicyStart(ctx)
 		k.Logger(ctx).Info(fmt.Sprintf("Starting new policy | Start Height : %d | End Height : %d", pmtpPeriodStartBlock, pmtpPeriodEndBlock))
 	}
-	var pmtpCurrentRunningRate sdk.Dec
+	// default to current pmtp current running rate value
+	pmtpCurrentRunningRate := k.GetPmtpRateParams(ctx).PmtpCurrentRunningRate
```
canercandan, line 29: @Jedi2002 @AustinoBombino I had to default this variable to the default `PmtpCurrentRunningRate` otherwise it was causing a seg fault when calling `k.PolicyRun` as it is past in parameter without getting initiated

AustinoBombino (approved):

AustinoBombino:
May want to cover cases where the current block is before / after a policy window

PR #2572, opened 2022-03-29
In multiLockBurn, update lockBurnNonce before we do any external calls.

banshee:
Also avoid recalculating recipient.length on every pass through the loop

### DIFF
--- a/smart-contracts/contracts/BridgeBank/BridgeBank.sol
+++ b/smart-contracts/contracts/BridgeBank/BridgeBank.sol
@@ -566,6 +566,10 @@ contract BridgeBank is BankStorage, CosmosBank, EthereumWhiteList, CosmosWhiteLi
         _lockTokens(recipient[i], token[i], amount[i], startingLockBurnNonce + i);
       }
     }
+
+    // If we get any reentrant calls from the _{burn,lock}Tokens functions, 
+    // make sure that lockBurnNonce is what we expect it to be.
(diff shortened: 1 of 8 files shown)


smart-contracts/contracts/BridgeBank/BridgeBank.sol
```diff
@@ -548,18 +548,24 @@ contract BridgeBank is BankStorage, CosmosBank, EthereumWhiteList, CosmosWhiteLi
     require(token.length == amount.length, "M_P");
     require(token.length == isBurn.length, "M_P");
 
-    uint256 intermediateLockBurnNonce = lockBurnNonce;
+    uint256 recipientLength = recipient.length;
+
+    // lockBurnNonce contains the previous nonce that was
+    // sent in the LogLock/LogBurn, so the first one we send
+    // should be lockBurnNonce + 1
+    uint256 startingLockBurnNonce = lockBurnNonce + 1;
```
smartyalgo, line 556: If i execute a reentrancy at _burn/_lockTokens, would I be able to emit an event with nonce that is ahead, such that nonce is no longer strictly increasing? If so it would halt the bridge I believe

Brando753 (approved):
looks good

banshee:
We need to add a test that does a multiLockBurn followed by a regular lock, and tests that the LogLock nonces are as expected.

PR #2580, opened 2022-03-31
Load testing framework

### LINKED ISSUE #2519: GRPC siftool support for sifchain transactions (for load testing)
Create python library to support communicating with sifnoded via grpc instead of command line.

This gives us more granular control over sequence number in order to invoke multiple tx at the same time

jzvikart:
Implemented load test for `tx ethbridge burn` between M sif acccounts and N ethereum accounts via gRPC.

Closes #2519.

### DIFF
--- a/test/integration/framework/src/siftool/cosmos.py
+++ b/test/integration/framework/src/siftool/cosmos.py
@@ -2,6 +2,42 @@ from siftool.common import *
 
 akash_binary = "akash"
 
+
+def balance_normalize(bal=None):
+    if type(bal) == list:
+        bal = dict(((k, v) for v, k in bal))
+    elif type(bal) == dict:
+        pass
+    else:
+        assert False, "Balances should be either a dict or a list"
+    return {k: v for k, v in bal.items() if v != 0}
+
+
+def balance_add(bal1, bal2):
+    result = {}
+    for denom in set(bal1.keys()).union(set(bal2.keys())):
+        val = bal1.get(denom, 0) + bal2.get(denom, 0)
+        if val != 0:
+            result[denom] = val
+    return result
+
+
+def balance_neg(bal):
+    return {k: -v for k, v in bal.items()}
+
+
+def balance_sub(bal1, bal2):
+    return balance_add(bal1, balance_neg(bal2))
+
+
+def balance_zero(bal):
+    return len(bal) == 0
+
+
+def balance_format(bal):
+    return ",".join("{}{}".format(v, k) for k, v in bal.items())
+
+
 # <editor-fold>
 
 # This is for Akash, but might be useful for other cosmos-based chains as well. (If not, it should be moved to separate
--- a/test/integration/framework/src/siftool/run_env.py
+++ b/test/integration/framework/src/siftool/run_env.py
@@ -845,7 +845,7 @@ class Peggy2Environment(IntegrationTestsEnvironment):
         assert ceth_symbol == "sifBridge99990x0000000000000000000000000000000000000000"
         # This goes to validator0, i.e. sifnode_validators[0]["address"]
         validator_mint_amounts = [
-            [999999 * 10**21, "rowan"],
+            [999999 * 10**27, "rowan"],
             [137 * 10**16, "ibc/FEEDFACEFEEDFACEFEEDFACEFEEDFACEFEEDFACEFEEDFACEFEEDFACEFEEDFACE"],
             [999999 * 10**21, ceth_symbol],
         ]
@@ -855,7 +855,7 @@ class Peggy2Environment(IntegrationTestsEnvironment):
         denom_whitelist_file = project_dir("test", "integration", "whitelisted-denoms.json")
         # These go to admin account, relayers and witnesses
         admin_account_mint_amounts = [
-            [10**20, "rowan"],
+            [10**27, "rowan"],
             [2 * 10**19, ceth_symbol]
         ]
         registry_json = project_dir("smart-contracts", "src", "devenv", "registry.json")
--- a/test/integration/framework/src/siftool/run_env.py
+++ b/test/integration/framework/src/siftool/run_env.py
@@ -568,9 +568,7 @@ class IntegrationTestsEnvironment:
         validator_count = 1
         network_definition_file = os.path.join(networks_dir, "network-definition.yml")
         seed_ip_address = "192.168.1.2"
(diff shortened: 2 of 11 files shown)


Brando753 (changes requested):
All new Python code MUST use type hints for all new or modified functions.

jzvikart:
I wasn't aware that typing hints were mandatory. Since this was a blocker and since we're short on time and since this is basically a prototype and there is still a lot of major refactoring still happening and since we already have a ticket for adding type hints I didn't pay special attention to it.

IMO it would be more effective to take some time separately for this task, clean up the code piece by piece and when it stabilizes, add documentation for stable parts first and type hints second.

Brando753:
> I wasn't aware that typing hints were mandatory. Since this was a blocker and since we're short on time and since this is basically a prototype and there is still a lot of major refactoring still happening and since we already have a ticket for adding type hints I didn't pay special attention to it.
> 
> IMO it would be more effective to take some time separately for this task, clean up the code piece by piece and when it stabilizes, add documentation for stable parts first and type hints second.

I disagree; typehints are a quick, simple feature that reduces bugs and problems that waste developer time. You wrote a bunch of code for siftool without using any typehints, and it was accepted because it would be a lot of work to refactor the entire toolset in bulk. I then asked that all future code that is modified or added please have type hints so we can gradually add typing to this system. I don't want newer code added in with the assumption that we will go through and add them later. It's easiest to do this when you just wrote or modified a function since you should know exactly what the functions are consuming and what they are returning. That outstanding ticket is for adding type hints to all the code missing typehints, not so that we cant continue to make new code not have them. 

A perfect example of why we should do this is because simple code can be difficult for other developers to work with when there are no typehints. This results in wasted time building everything, running the integration tests (which can take a while to complete), then finding out that code either acted improperly or had runtime errors because of a type mismatch. Case in point where that has already [happened in this codebase ](https://github.com/Sifchain/sifnode/commit/6ff922eb2c55975ac31d61c0fa9ef18f565a0413#diff-98ebc3eee0af89e0fb65b5575527b8d0a6de7b3f7c57816947e7b34a079e8682) the function `sifchain_denom_hash(network_descriptor, token_contract_address)` seemed simple enough, it took two params one being `network_descriptor` when I was working on this code I assumed `network_descriptor` would be an integer because that's how it is defined in our proto files. Most of the time, it was, except for the one time where it was a string and as a result, I wasted time building and running this only to discover sometimes it's not an integer and had to add additional code. I assume that it was always meant to be an integer, and the string is a bug, but the linter could not detect this because we did not have typehints. 
Further typehints make code reviews take less time because I don't have to hunt down many references just to figure out what's going in and coming out of function calls. I am not approving any new PRs that don't add type hints to modified functions and have them in existing functions.

test/integration/framework/src/siftool/cosmos.py
```diff
@@ -1,7 +1,66 @@
...
 
 akash_binary = "akash"
 
+LegacyBalance = List[Union[int, str]]  # e.g. [[3, "rowan"], [2, "ibc/xxxxx"]]
+Balance = Mapping[str, int]
+CompatBalance = Union[LegacyBalance, Balance]
+Address = str
+
+
+def balance_normalize(bal: CompatBalance = None):
```
juniuszhou, line 12: missing return type

test/integration/framework/src/siftool/run_env.py
```diff
@@ -568,9 +568,7 @@ def run(self):
         validator_count = 1
         network_definition_file = os.path.join(networks_dir, "network-definition.yml")
         seed_ip_address = "192.168.1.2"
-        mint_amount = [[999999 * 10**21, "rowan"],
-                       [137 * 10**16, "ibc/FEEDFACEFEEDFACEFEEDFACEFEEDFACEFEEDFACEFEEDFACEFEEDFACEFEEDFACE"],
-                       [10**28, "sifBridge99990x0000000000000000000000000000000000000000"]]
+        mint_amount = [[999999 * 10**21, "rowan"], [137 * 10**16, "ibc/FEEDFACEFEEDFACEFEEDFACEFEEDFACEFEEDFACEFEEDFACEFEEDFACEFEEDFACE"]]
```
juniuszhou, line 571: please don't remove the sifBridge99990x0000000000000000000000000000000000000000, it is used to pay cross chain fee from sifnode to ethereum.

juniuszhou (changes requested):

jzvikart:
`network_descriptor` was simply a bug, I already fixed it. There will always be bugs, type hints or not. As for type hints - it is a matter of taste, so there is no point arguing it.

I added them now (in part), but there is a significant downside to it. Since we are importing dynamically generated gRPC stuff this now means that we have to generate it even if we're not using it (= every time we run siftool). Generating it means, among other things, cloning entire cosmos SDK repo which is 500M and takes ~40 mins on my machine.

test/integration/framework/src/siftool/run_env.py
jzvikart: I removed it from class `IntegrationTestsEnvironment` which is used for exclusively for Peggy1. As far as I understand this denom is Peggy2-specific.

jzvikart:
The basic problem is this:
- `cosmos` is not a library. It is dynamically generated from protobuf files for working with gRPC.
- Before adding typing hints, files for `cosmos` were generated with `siftool generate-python-protobuf-stubs` and `import cosmos` statements were lazy evaluated (in a function) only when you actually wanted to work with gRPC.
- Now, the `import cosmos` statements have to be evaluated on a module level because they are needed for typing hints.
- To generate Python files for `cosmos`, I need to combine protobuf files from different sources, one of them is cosmos-sdk.
- To get cosmos-sdk, we need to do a `git clone` first. This is a large repository (500M) and takes significant time to download (on my machine ~40 mins).

So because of typing hints we have essentially run into a catch-22 problem, and even if we work around it, we still have to download cosmos-sdk before siftool can be used. Doing it once on a developer machine might be acceptable, but it could be a problem in CI/CD.

I see few possible solutions:
- Avoid type hints for anything having to do with `cosmos` , and go back to lazy evaluation of `imports`
- Include generated files in git
- Include protobuf files in git

In any case, protobuf files are a mess (a legacy we inherited from cosmos) and we should find a better way to manage them.

Brando753 (approved):
We discussed refactors that should happen in a huddle, none of those should block this from merge.

PR #2581, opened 2022-03-31
Peggy2/2519/siftool env debugging

smartyalgo:
In this PR:

- Non-code change cleanup
- Correct how launch.json is constructed for witness


jzvikart (approved):
Looks good to me.

PR #2585, opened 2022-04-01
[ChainOps] Token registry

intl-man:
Includes:

* Added the missing type for instantiating an empty token registry.
* Allow a token registry admin address to be added at genesis.


canercandan (approved):
LGTM!

PR #2586, opened 2022-04-01
Automatically deploy bridge token for ibc export

juniuszhou:


smart-contracts/contracts/CosmosBridge.sol
```diff
@@ -347,10 +347,10 @@ contract CosmosBridge is CosmosBridgeStorage, Oracle {
         );
       } else {
         // if we are double pegging and already control the token, then we are going to need to get the address on this chain
-        tokenAddress = sourceAddressToDestinationAddress[claimData.tokenAddress];
+        tokenAddress = cosmosDenomToDestinationAddress[claimData.cosmosDenom];
       }
     } else {
-      tokenAddress = claimData.tokenAddress;
+        tokenAddress = claimData.tokenAddress;
```
Brando753, line 353: What if its a first time double pegged token, how is it getting created?

smart-contracts/contracts/CosmosBridge.sol
```diff
@@ -25,7 +25,7 @@ contract CosmosBridge is CosmosBridgeStorage, Oracle {
   /**
    * @notice Maps the original address of a token to its address in another network
```
smartyalgo, line 26: Update comments

smart-contracts/contracts/CosmosBridge.sol
```diff
@@ -450,4 +450,42 @@ contract CosmosBridge is CosmosBridgeStorage, Oracle {
...
     emit LogProphecyCompleted(prophecyID, success);
   }
+
+  /**
+   * @dev Verifies if `cosmosDenom` is a ibc token
+   * @param cosmosDenom The denom of the token in sifnode
+   * @return Boolean: is `cosmosDenom` a ibc token?
+   */
+  function isIbcToken(string calldata cosmosDenom) private view returns (bool) {
+    bytes calldata cosmosDenomBytes = bytes(cosmosDenom);
```
Brando753, line 460: What about rowan?

smart-contracts/test/helpers/denoms.js
```diff
@@ -54,6 +54,12 @@ const DENOM_4 = generateDenom({
   isERC20: true,
 });
 
+const IBC_DENOM = generateDenom({
```
Brando753, line 57: This appears to be ETH, why does it say IBC denom?

smart-contracts/test/test_ibcTokenExport.js
```diff
@@ -0,0 +1,144 @@
+const Web3Utils = require("web3-utils");
```
Brando753, line 1: Is this a brand new JS file? Why was it not made as a typescript file?

Brando753 (changes requested):

smart-contracts/test/helpers/denoms.js
juniuszhou: you can check generateDenom method, if isERC20 as false, it will generate a ibc denom

smart-contracts/contracts/CosmosBridge.sol
juniuszhou: rowan is not ibc token, the bridge token address is available from claim.tokenAddress

smart-contracts/contracts/CosmosBridge.sol
juniuszhou: this else branch for not double peg case.

smart-contracts/test/test_ibcTokenExport.js
juniuszhou: will change it to ts file

smart-contracts/contracts/CosmosBridge.sol
Brando753: Is that going to be true on other networks we deploy to?

smart-contracts/contracts/CosmosBridge.sol
```diff
@@ -332,10 +333,34 @@ contract CosmosBridge is CosmosBridgeStorage, Oracle {
         );
       } else {
         // if we are double pegging and already control the token, then we are going to need to get the address on this chain
-        tokenAddress = sourceAddressToDestinationAddress[claimData.tokenAddress];
+        tokenAddress = cosmosDenomToDestinationAddress[claimData.cosmosDenom];
       }
-    } else {
-      tokenAddress = claimData.tokenAddress;
+    } else if (isIbcToken(claimData.cosmosDenom)) {
```
Brando753, line 338: Do we need to actually make a distinction between a double pegged token and a ibc token on the smart contract side of things? Instead of having a field that says `doublePeg` why not instead just have a field that says `bridgeToken` and offload that logic out of the smart contracts? Both will be bridge tokens be it double pegged or IBC and both will be burned so why handle them differently?

Brando753 (changes requested):

smart-contracts/contracts/CosmosBridge.sol
juniuszhou: I think you are right, now we can use a flag like bridgeToken to replace both doublePeg and isIbcToken judgement. the reason that we handle it differently is the requirement, I think. The old procedure is we add the ibc denom into registry and deploy bridge token for ibc token, and in the cosmosBridge contract, we use the original contract address as the key, but ibc token doesn't have the field.

juniuszhou: In the sifnode side, we still need doublePeg field since we have different strategy whether to charge the bridge token deployment fee for user.

Brando753: sure but we can move the logic into sifnode or even the relayers to look at the cosmos denom and determine if its a double pegged token or a ibc token right? If its `rowan` its native rowan, if it starts with `ibc/` its an ibc asset else its a double bridged asset.

x/ethbridge/keeper/msg_server.go
```diff
@@ -64,6 +65,9 @@ func (srv msgServer) Lock(goCtx context.Context, msg *types.MsgLock) (*types.Msg
 	globalSequence := srv.Keeper.GetGlobalSequence(ctx, msg.NetworkDescriptor)
 	srv.Keeper.UpdateGlobalSequence(ctx, msg.NetworkDescriptor, uint64(ctx.BlockHeight()))
 
+	// we also take ibc token is bridge token, since it is not from sifnode
+	isBridgeToken := strings.HasPrefix(msg.DenomHash, "ibc/")
```
Brando753, line 69: What about if the token is rowan? This will work currently when we are on Ethereum, but what about when more chains are added?

Brando753 (changes requested):

Brando753 (approved):

PR #2594, opened 2022-04-02
[ChainOps] API/gRPC updates for sifgen

intl-man:
Includes:

* We'll now enable the API and gRPC by default. There was no need to not enable these services. We will still have control over making said services available to the outside world however (via helm).


canercandan (approved):
LGTM

PR #2595, opened 2022-04-02
Feature/pmtp incremental tests

canercandan:


x/clp/abci_test.go
```diff
@@ -223,3 +223,202 @@ func TestBeginBlocker(t *testing.T) {
...
+			app.ClpKeeper.SetParams(ctx, tc.params)
+			app.ClpKeeper.SetPmtpRateParams(ctx, types.PmtpRateParams{
+				PmtpPeriodBlockRate:    sdk.ZeroDec(),
+				PmtpCurrentRunningRate: sdk.ZeroDec(),
+			})
+			app.ClpKeeper.SetPmtpEpoch(ctx, tc.epoch)
+
+			for i := 0; i < len(tc.expectedStates); i++ {
+				clp.BeginBlocker(ctx, app.ClpKeeper)
+
```
Jedi2002, line 413: Add `ctx = ctx.WithBlockHeight(ctx.BlockHeight()+1)`

x/clp/abci_test.go
```diff
@@ -223,3 +223,202 @@ func TestBeginBlocker(t *testing.T) {
...
+		nativeBalance          sdk.Int
+		externalBalance        sdk.Int
+		nativeAssetAmount      sdk.Uint
+		externalAssetAmount    sdk.Uint
+		poolUnits              sdk.Uint
+		poolAssetPermissions   []tokenregistrytypes.Permission
+		nativeAssetPermissions []tokenregistrytypes.Permission
+		params                 types.Params
+		epoch                  types.PmtpEpoch
+		expectedStates         ExpectedStates
```
Jedi2002, line 251: Can we read all these values from a file , it might be easier to geenrate the test cases from a json file

Jedi2002: But this can be done later as a refactor as well .

x/clp/abci_test.go
```diff
@@ -223,3 +223,202 @@ func TestBeginBlocker(t *testing.T) {
...
+					pool: types.Pool{
+						ExternalAsset:        &types.Asset{Symbol: "eth"},
+						NativeAssetBalance:   sdk.NewUint(1000),
+						ExternalAssetBalance: sdk.NewUint(1000),
+						PoolUnits:            sdk.NewUint(1000),
+					},
+					SwapPriceNative:   sdk.MustNewDecFromStr("0.998002996005000000"),
+					SwapPriceExternal: sdk.MustNewDecFromStr("0.998002996005000000"),
+					pmtpRateParams: types.PmtpRateParams{
+						PmtpPeriodBlockRate:    sdk.NewDec(0),
```
Jedi2002, line 319: Should'nt the block and running rate start changing at height 1 ?

x/clp/abci_test.go
canercandan: yep agree json or a csv for those repetitive values would make more sense as we want to extend the tests to a greater number of blocks,

x/clp/abci_test.go
canercandan: @Jedi2002 I am just initiating them here at a nil value but I think I was doing that before we solve the bugs we did last days, will try to get rid of that

x/clp/abci_test.go
canercandan: nice thanks

x/clp/abci_test.go
canercandan: oh no sorry those are expected values, I just put the values that the current version was returning but now as I am incrementing the height it looks like those values are changing will update the branch with the new values soon

PR #2599, opened 2022-04-04
Inter-Policy Rate Parameter

AustinoBombino:
Added a parameter to track rate between policies and add on to future policies. This is to avoid rate resets on subsequent policy starts.


canercandan (approved):
LGTM!

PR #2602, opened 2022-04-05
PMTP - fix first block calc

AustinoBombino:
The existing calc isn't inclusive of the first block of a policy. Add one to the exponent to compensate.


canercandan (approved):
LGTM thanks

PR #2611, opened 2022-04-07
Load testing from ethereum to sifnode

### LINKED ISSUE #2528: Peggy 2.0 Load Testing: Large number of transactions of tokens to a single wallet

juniuszhou:
The pr will resolve #2528


test/load/test_load_tx_ethereum.py
```diff
@@ -0,0 +1,268 @@
...
+
+fund_amount_eth = 10 * eth.ETH
+fund_amount_sif = 10 * test_utils.sifnode_funds_for_transfer_peggy1
+rowan_contract_address = "0x5FbDB2315678afecb367f032d93F642f64180aa3"
+
+threads_num = 3
+amount_to_send = 10000
+
+def bridge_bank_lock_burn(eth_tx_wrapper: eth.EthereumTxWrapper, bridge_bank_sc: Contract, test_eth_account: str, 
+    recipient: str, amount_to_send: int, nonce: int, token_sc: Contract = None, isLock: bool = True):
```
jp-sif, line 22: There's a function-scoped variable here called `amount_to_send`, whose name is the same as a defined module variable in L19 above. This may lead to confusion.

jp-sif (approved):
I'm not sure about the web3 stuff and the contract dances, which I suppose is really what you would like me to review. When I get more familiar with that sort of thing, my reviews will become more useful.

test/load/test_load_tx_ethereum.py
juniuszhou: will rename the variable.

jzvikart:
At first glance there seems a lot of repetition in the tests. I strongly encourage you to factor out repeating parts of code into functions that are shared among tests and can simply be called from each test with different parameters.

test/load/test_load_tx_ethereum.py
```diff
@@ -0,0 +1,273 @@
...
+from siftool.eth import NULL_ADDRESS
+from siftool.common import *
+from web3.eth import Contract
+import web3
+
+from eth_typing import (
+    ChecksumAddress,
+)
+
+# In separate window: test/integration/framework/siftool run-env
```
jprupp, line 16: Instead of having the user launch this is a separate window, I think it would make sense for the test suite to launch these things directly, and then shut them down when it is done.

test/load/test_load_tx_ethereum.py
```diff
@@ -0,0 +1,273 @@
...
+# test/integration/framework/venv/bin/python3 test/load/test_load_tx_ethereum.py
```
jprupp, line 17: This too, it is a Python program, so it could conceivably be launched by the test suite and include code to clean up when tests finish.

jprupp (approved):
This looks good to me. I finally understand what it does.

test/load/test_load_tx_ethereum.py
juniuszhou: In most of case, we create the testing env and run lots of different test cases, integration test or load test. The startup needs quite a long time, so it is a waste of launching it to run a single test.

jzvikart:
@jprupp I was actually thinking about launching an environment if you run the test through `__main__`. It's actually pretty trivial and it would turn a test into standalone runnable file, which is nice. During development however I prefer having the environment running in another window to avoid startup delay (it's about 30 seconds, but still).

test/load/test_load_tx_ethereum.py
```diff
@@ -41,39 +41,44 @@ def bridge_bank_lock_burn(eth_tx_wrapper: eth.EthereumTxWrapper, bridge_bank_sc:
...
-        rowan_sc = w3_conn.eth.contract(abi=rowan_abi, address=rowan_contract_address)
-        test_eth_account = test_eth_accounts[i]
-        nonce = w3_conn.eth.get_transaction_count(test_eth_account)
-        recipient = test_utils.sif_addr_to_evm_arg(test_sif_account)
-        eth_tx_wrapper = eth.EthereumTxWrapper(w3_conn, ctx.eth.is_local_node)
-        eth_tx_wrapper.set_private_key(test_eth_account, ctx.eth._get_private_key(test_eth_account))
-        eth_tx_wrappers.append(eth_tx_wrapper)
-        threads.append(threading.Thread(target=bridge_bank_lock_burn, args=(
-            eth_tx_wrapper, bridge_bank_sc, test_eth_account, recipient, amount_to_send, nonce, rowan_sc, False)))
+            w3_conn = eth.web3_connect(w3_url, websocket_timeout=90)
```
jzvikart, line 63: I think this part is over-indented by one level.

jzvikart (approved):

PR #2612, opened 2022-04-07
Fix Github Actions go.yml test

khdegraaf:


khdegraaf:
Need to cherry pick just my stuff, as this breaks other stuff.

PR #2613, opened 2022-04-08
Fix GitHub action go.yml tests

khdegraaf:
We should at least have this stuff running and passing if possible in peggy2 active branch.  Running on push,pull_request just ends up running twice often and isn't needed.  I have fixed at least unit tests so they run correctly, but integration and localnet tests are still broken.

### DIFF
--- a/.github/workflows/gitleaks.yml
+++ b/.github/workflows/gitleaks.yml
@@ -1,6 +1,6 @@
 name: gitleaks
 
-on: [push,pull_request]
+on: push
   
 jobs:
   gitleaks:
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -24,10 +24,6 @@ jobs:
     - name: Check out code into the Go module directory
       uses: actions/checkout@v2
 
-    - name: Get dependencies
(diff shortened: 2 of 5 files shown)


khdegraaf:
The integration tests are still broken.  In trying to run it locally (failed so far), I see a missing requirement in requirements.txt (web3), and incorrect vagrantenv.sh (smart contracts directory info).

.github/workflows/gitleaks.yml
```diff
@@ -1,6 +1,6 @@
 name: gitleaks
 
-on: [push,pull_request]
```
smartyalgo, line 3: Not required for this PR:

A user is still able to commit sensitive information, because this is only ran on push, and on on commit. Would on pre-commit be a better place?

smartyalgo (approved):

.github/workflows/gitleaks.yml
khdegraaf: This check is actually completely broken and doesn't work correctly.  It also takes about 27 minutes to run.  So yeah, this needs to be improved...

PR #2620, opened 2022-04-08
rename reward params, empty defaults

AustinoBombino:
renames reward parameters to be more verbose. 
empty reward and pmtp defaults


canercandan (approved):
LGTM thanks for working on that one

PR #2625, opened 2022-04-10
[ChainOps] Cleanup.

intl-man:
Includes:

* Removed previous committed files to `x/` that should not have been committed in the first place.
* Removed the UI tests - these don't currently work. We will need to revisit these at a later date.
* Bumped the version to `0.13.0-rc1` in `version`.
* Linter fixes.


x/clp/types/msgs.go
```diff
@@ -82,10 +82,10 @@ func (m MsgUpdateRewardsParamsRequest) Type() string {
 }
 
 func (m MsgUpdateRewardsParamsRequest) ValidateBasic() error {
-	if m.LiquidityRemovalCancelPeriod < 0 {
+	if m.LiquidityRemovalCancelPeriod == 0 {
 		return errors.Wrap(ErrInvalid, "LiquidityRemovalCancelPeriod cannot be less than 0")
 	}
-	if m.LiquidityRemovalLockPeriod < 0 {
+	if m.LiquidityRemovalLockPeriod == 0 {
```
canercandan, line 88: @intl-man I have to check those error cases with @Jedi2002 but if we change the condition operator the error message now does not have the same meaning as what we are verifying for the 3 conditions you changed line 40, 85 and 88

canercandan (commented):
@intl-man please review my comment thanks

x/clp/types/msgs.go
intl-man: @canercandan - yep, totally. 

I guess the question is why on earth are we trying to check for less than zero on `m.LiquidityRemovalLockPeriod` ? It's `uint64` which has a min value of 0 (can't be negative). So this check is probably completely redundant and maybe it should be removed?

AustinoBombino: Yeah I don't see why we have these cases

AustinoBombino (approved):

intl-man:
Closing, as most issues have been dealt with in other PRs.

PR #2627, opened 2022-04-11
Feature/tokenregistry admin accounts

Jedi2002:


x/tokenregistry/types/genesis.go
```diff
@@ -30,7 +37,38 @@ func GetGenesisStateFromAppState(marshaler codec.JSONCodec, appState map[string]
...
+				AdminAddress: "sif1l7hypmqk2yc334vc6vmdwzp5sdefygj2ad93p5",
+			},
+			{
+				AdminType:    AdminType_ETHBRIDGE,
+				AdminAddress: "sif1syavy2npfyt9tcncdtsdzf7kny9lh777yqc2nd",
+			},
+			{
+				AdminType:    AdminType_TOKENREGISTRY,
+				AdminAddress: "sif1syavy2npfyt9tcncdtsdzf7kny9lh777yqc2nd",
+			},
```
AustinoBombino, line 66: @Jedi2002 is this sif1syavy2npfyt9tcncdtsdzf7kny9lh777yqc2nd the real admin account? or do we still need to update the migrator?

Jedi2002: WE need to add to the migrator as well , I was thinking of an if condition like so 
if ctx.ChainID() != "sifchain-1" { } 
if ctx.ChainID() != "sifchain-devnet-1" { } 
if ctx.ChainID() != "sifchain-testnet-1" { } 

Just need the correct addresses for that

x/tokenregistry/keeper/genesis.go
```diff
@@ -8,23 +8,19 @@ import (
...
-			panic(err)
+	if state.AdminAccounts != nil {
+
+		for _, adminAccount := range state.AdminAccounts.AdminAccounts {
+			k.SetAdminAccount(ctx, adminAccount)
 		}
-		k.SetAdminAccount(ctx, addr)
-	}
-	if state.Registry != nil {
-		k.SetRegistry(ctx, *state.Registry)
```
canercandan, line 19: @Jedi2002 I think those 2 lines have been mistakenly removed

canercandan (approved):
LGTM

Issue #2520, opened 2022-03-09
Sifnode Load Testing for Peggy 2.0


pandaring2you:
Hey @Jedi2002 - The peggy team was going through the load testing plan for Peggy 2.0 and realized that Sifnode will likely want to conduct load testing given that Peggy 2.0 will be removing the whitelist and potentially thousands of new tokens will be in our system. Have you considered what type of load testing we'll need to conduct?

One potential scenario:
- Create a wallet with 1K, 10K, 100K, different denoms

@smartyalgo @banshee - feel free to comment if you think of additional scenarios sifnode should be testing

cc: @marshsif

pandaring2you:
Testing needed prior to peggy 2.0. Depending on results, need to determine if it blocks peggy 2.0 launch

pandaring2you:
hey @Jedi2002 - can you give this a look and let us know timing?

pandaring2you:
@khdegraaf We will need a partner on the testing team to suggest scenarios to test for to work with @iverc @joneskm . Do you have a suggestion on who would be best to work on this?

Known issue: Adding thousands of tokens in the registry, increasing the cost of queries/operations. 

Potential solutions: 
- Optimize queries/operations
- Add cost to import tokens. 

@pandaring2you - Write up doc on potential token spam attacks, get feedback from testing team and peggy. Then organize meeting to draw up solutions.

pandaring2you:
@jzvikart - to assess scenarios - order them by severity. determine severity and discuss solutions with sifnode/peggy teams

jzvikart:
We can get a large number of tokens in a sif wallet in two ways:
- by importing them from ERC20 tokens
- by creating them in sifnoded genesis (requires parametrization of environment + starting environment from within the test itself)

Because those tokens are different (one are cosmos native and others are not), we need to test them with two different scenarios. For now, we will focus on EVM-native tokens.

Next steps:
- Expand load testing framework to deploy and mint an arbitrary number of ERC20 tokens and send them to a sif wallet
- Run it with 10, 100, 1k, 10k, 100k and see how many tokens can be realistically created in ~30 mins
- Perform some basic tests on this wallet such as checking balance, sending some tokens back to Ethereum, sending some tokens to another sif wallet
- Discuss with @iverc how to measure/detect potential performance issues with Sifnode involving a large number of 
denoms in a wallet (performance must be O(1) or O(n) at worst) - might require profiling instrumentation on sifnode side
[...]

PR #2639, opened 2022-04-13
Load test: many denoms sif wallet

jzvikart:
This test creates up to 5000 ERC20 tokens and transfers them to a single sif account.
It checks that the wallet is usable (it can transfer funds to/from another sif wallet and do ethbridge burn).
It asserts that the fees are constant.
It measures the relative performance of a wallet with many denoms vs. wallet with one denom.

This PR is part of #2520.

### DIFF
--- a/test/integration/framework/src/siftool/common.py
+++ b/test/integration/framework/src/siftool/common.py
@@ -82,7 +82,7 @@ def popen(args: Sequence[str], cwd: Optional[str] = None, env: Optional[Mapping[
 ) -> subprocess.Popen:
     if env:
         env = dict_merge(os.environ, env)
-    logging.debug(f"popen(): args={repr(args)}, cwd={repr(cwd)}")
+    log.debug(f"popen(): args={repr(args)}, cwd={repr(cwd)}")
     return subprocess.Popen(args, cwd=cwd, env=env, stdin=stdin, stdout=stdout, stderr=stderr, text=text)
 
 def dict_merge(*dicts, override=True):
@@ -108,6 +108,7 @@ def basic_logging_setup():
     logging.getLogger("eth").setLevel(logging.WARNING)
     logging.getLogger("websockets").setLevel(logging.WARNING)
     logging.getLogger("web3").setLevel(logging.WARNING)
+    logging.getLogger("asyncio").setLevel(logging.WARNING)
 
 # Recursively transforms template strings containing "${VALUE}". Example:
 # >>> template_transform("You are ${what}!", {"what": "${how} late", "how": "very"})
--- a/test/integration/framework/src/siftool/cosmos.py
+++ b/test/integration/framework/src/siftool/cosmos.py
@@ -29,6 +29,15 @@ def balance_add(*bal: Balance) -> Balance:
     return result
 
 
+def balance_mul(bal: Balance, multiplier: Union[int, float]) -> Balance:
+    result = {}
+    for denom, value in bal.items():
+        val = value * multiplier
+        if val != 0:
+            result[denom] = val
+    return result
+
+
 def balance_neg(bal: Balance) -> Balance:
     return {k: -v for k, v in bal.items()}
 
--- a/test/integration/framework/src/siftool/cosmos.py
+++ b/test/integration/framework/src/siftool/cosmos.py
@@ -19,10 +19,11 @@ def balance_normalize(bal: CompatBalance = None) -> Balance:
     return {k: v for k, v in bal.items() if v != 0}
 
 
-def balance_add(bal1: Balance, bal2: Balance) -> Balance:
+def balance_add(*bal: Balance) -> Balance:
     result = {}
-    for denom in set(bal1.keys()).union(set(bal2.keys())):
-        val = bal1.get(denom, 0) + bal2.get(denom, 0)
+    all_denoms = set(flatten_list([[*b.keys()] for b in bal]))
+    for denom in all_denoms:
+        val = sum(b.get(denom, 0) for b in bal)
         if val != 0:
             result[denom] = val
     return result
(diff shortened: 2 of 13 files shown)


test/integration/framework/src/siftool/cosmos.py
```diff
@@ -19,10 +19,20 @@ def balance_normalize(bal: CompatBalance = None) -> Balance:
     return {k: v for k, v in bal.items() if v != 0}
 
 
-def balance_add(bal1: Balance, bal2: Balance) -> Balance:
+def balance_add(*bal: Balance) -> Balance:
```
juniuszhou, line 22: This function name is confusing to me, I think for a while and then realize what is doing. I think balance_aggregate or balance_combine is better to describe its functionality.

juniuszhou, line 22: This function name is confusing to me, I think for a while and then realize what is doing. I think balance_aggregate or balance_combine is better to describe its functionality.

test/integration/framework/src/siftool/cosmos.py
```diff
@@ -32,8 +42,8 @@ def balance_neg(bal: Balance) -> Balance:
     return {k: -v for k, v in bal.items()}
 
 
-def balance_sub(bal1: Balance, bal2: Balance) -> Balance:
-    return balance_add(bal1, balance_neg(bal2))
+def balance_sub(bal1: Balance, *bal2: Balance) -> Balance:
+    return balance_add(bal1, *[balance_neg(b) for b in bal2])
```
juniuszhou, line 46: balance_add just take one parameter

juniuszhou, line 46: balance_add just take one parameter

test/integration/framework/src/siftool/cosmos.py
jzvikart: It's a n-ary add operation, essentially a + b + c + ...
It's just a generalization of a + b.

test/integration/framework/src/siftool/cosmos.py
jzvikart: No, the `*` in Python stands for argument splicing, both `balance_add` and `balance_sub` support a list of arguments. With `*` you can write instead of `balance_sub(a, [b, c, d])` just `balance_sub(a, b, c, d)`.
The semantics is `a - b - c - d`.

test/integration/framework/src/siftool/cosmos.py
juniuszhou: understand now, thanks!

juniuszhou (approved):

test/integration/framework/src/siftool/cosmos.py
```diff
@@ -19,10 +19,20 @@ def balance_normalize(bal: CompatBalance = None) -> Balance:
...
 
-def balance_add(bal1: Balance, bal2: Balance) -> Balance:
+def balance_add(*bal: Balance) -> Balance:
     result = {}
-    for denom in set(bal1.keys()).union(set(bal2.keys())):
-        val = bal1.get(denom, 0) + bal2.get(denom, 0)
+    all_denoms = set(flatten_list([[*b.keys()] for b in bal]))
+    for denom in all_denoms:
+        val = sum(b.get(denom, 0) for b in bal)
+        if val != 0:
```
smartyalgo, line 27: I dont think you'll run into val == 0, It should be safe to remove this if clause

test/integration/framework/src/siftool/eth.py
```diff
@@ -242,6 +247,15 @@ def _send_raw_transaction(self, smart_contract_call_obj, from_addr, tx_opts=None
         txhash = self.w3_conn.eth.send_raw_transaction(signed_tx.rawTransaction)
         return txhash
 
+    def wait_for_all_transaction_receipts(self, tx_hashes: Sequence[HexBytes], sleep_time: int = 5,
```
smartyalgo, line 250: `w3_conn.eth.wait_for_transaction_receipt` returns `TxReceipt` , should we return `Sequence[TxReceipt] ` to be more precise?

also, add return type hint for `wait_for_transaction_receipt`

test/localnet/utils/tests/__snapshots__/getChains.test.mjs.snap
```diff
@@ -124,7 +124,6 @@ Object {
   },
```
smartyalgo, line 124: What is this file for? How is this file generated?

smartyalgo (changes requested):
WIP

test/integration/framework/src/siftool/common.py
```diff
@@ -108,6 +108,7 @@ def basic_logging_setup():
     logging.getLogger("eth").setLevel(logging.WARNING)
```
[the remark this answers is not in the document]
jzvikart: This part sets the logging levels. We set the overall logging level to `DEBUG`, but since certain external libraries such as web3 emit a lot of noise with `DEBUG`, we silence those libraries to `WARNING` level.

test/integration/framework/src/siftool/cosmos.py
jzvikart: Why not? `1 + (-1) == 0`. This keeps the balance dict normalized, which means it should not contain any keys for which the value is zero. In certain cases, test "== 0" could be a potential problem with numerical stability for float types, but AFAIK we're not doing anything like that for now.

test/localnet/utils/tests/__snapshots__/getChains.test.mjs.snap
jzvikart: This is @canercandan's file, from what I saw its only reason for existence is that a unit test can compare it to another JSON file. The test broke when I fixed the other JSON file and from my point of view this test makes no sense at all, but it's out of scope for my work and @canercandan was unavailable at the time to confirm it.

test/integration/framework/src/siftool/eth.py
jzvikart: Agree, fixed.

PR #2643, opened 2022-04-14
Updated most of the smart contracts scripts to typescript. Updated th…

Brando753:
Integrate the blocklist deploy into the future/peggy2 branch


smart-contracts/tasks/blocklist/ofacParser.ts
```diff
@@ -0,0 +1,49 @@
...
+  const response = await axios.get(finalUrl).catch((e) => {
+    throw e;
+  });
+
+  const addresses = extractAddresses(response.data);
+
+  return addresses;
+}
+
+export function extractAddresses(rawFileContents: string) {
```
jzvikart, line 28: Wasn't there also a formatted version of OFAC blocklist file? I thought somebody mentioned it a while ago in a meeting...

Brando753: We had a script that would write these results to a file, but I deleted it as I didn't see it used for anything nor could I understand the use case.

jzvikart: No, I think I remember seeing the actual JSON/XML file or something... I'm not 100% sure, but try asking James

Brando753: You sure it wasn't a file generated by https://github.com/Sifchain/sifnode/blob/716e55c0f28ca3ae142706ee031d3bdf3aa1b8b8/smart-contracts/scripts/download_ofac_blocklist.js That would make a JSON file.

jzvikart:
How much do we really need Hardhat here, and for what? It would be nice to decouple from hardhat as much as possible, especially since we're thinking about using geth as the main implementation.

Brando753:
> How much do we really need Hardhat here, and for what? It would be nice to decouple from hardhat as much as possible, especially since we're thinking about using geth as the main implementation.

Hardhat lets us trivially test locally and compiles our smart contracts, and it provides typing through the typechain plugin, which is super useful when scripting. Just because the load tests will use geth rather than a hardhat node does not mean we need to decouple hardhat from our tests/scripts. Hardhat takes any network RPC in its config. Just as I can point hardhat to Ethereum mainnet I can point it to your local geth instance. I see no reason to decouple it unless hardhat itself is causing issues and not just the hardhat node instance.

smart-contracts/tasks/blocklist/ofacParser.ts
banshee: OFAC publishes a machine-readable list, we don't need to parse the plaintext.  https://home.treasury.gov/policy-issues/financial-sanctions/specially-designated-nationals-list-data-formats-data-schemas

banshee (approved):

PR #2655, opened 2022-04-20
limit query getProphciesCompleted query result

juniuszhou:
we need to avoid the getPropheciesCompleted return too much prophecies in single query.


x/oracle/keeper/prophecy.go
```diff
@@ -225,5 +229,5 @@ func (k Keeper) GetProphecyInfoWithScopeGlobalSequence(ctx sdk.Context,
 		globalSequence++
 		result = append(result, &prophecyInfo)
 	}
-
+	return result
```
jprupp, line 232: Interesting that you didn't need this explicit return in the previous version, and you need it now.

jprupp (approved):
Looks good to me.

x/oracle/keeper/prophecy.go
juniuszhou: the loop changed from without any condition to a fixed times.

smartyalgo: @jprupp Golang supports named return. that's why
https://stackoverflow.com/questions/45239409/empty-return-in-func-with-return-value-in-golang

PR #2657, opened 2022-04-20
Fix a typo prophcies

juniuszhou:
typo prophcies in proto buf definition file


smartyalgo (approved):

smartyalgo:
@jprupp those `.pb.` are generated via `make .proto-gen` at root using protobuf files. It creates artifacts for cosmos

PR #2665, opened 2022-04-22
Feature/bridge pauser

Jedi2002:
This PR adds a new transaction to pause `Lock` and `Burn` transactions.


x/ethbridge/keeper/msg_server_test.go
```diff
@@ -0,0 +1,92 @@
...
+	"github.com/Sifchain/sifnode/x/ethbridge/types"
+	types2 "github.com/Sifchain/sifnode/x/tokenregistry/types"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/stretchr/testify/require"
+	"testing"
+)
+
+func TestMsgServer_Lock(t *testing.T) {
+	ctx, app := test.CreateSimulatorApp(false)
+	receiverCoins := app.BankKeeper.GetAllBalances(ctx, cosmosReceivers[0])
```
smartyalgo, line 15: Why do we need this and the next line? Is it because we are reusing cosmosReceivers[0]? 
In this case, why do we have cosmosReceivers, instead of generating an address through test_helper::CreateTestAddrs inside the unit test?

x/ethbridge/keeper/msg_server_test.go
```diff
@@ -0,0 +1,92 @@
+package keeper_test
```
smartyalgo, line 1: Can you also add test to assert behaviour when we pass `SetPauser` with a non-admin account?

smartyalgo (changes requested):

x/ethbridge/client/cli/tx.go
```diff
@@ -417,3 +418,35 @@ func GetCmdSetBlacklist() *cobra.Command {
...
+	cmd := &cobra.Command{
+		Use:   "set-pauser",
+		Short: "pause or unpause Lock and Burn transactions",
+		Args:  cobra.ExactArgs(1),
+		RunE: func(cmd *cobra.Command, args []string) error {
+			clientCtx, err := client.GetClientTxContext(cmd)
+			if err != nil {
+				return err
+			}
+			isPaused, err := utils.ParseStringToBool(args[0])
```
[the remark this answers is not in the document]
Jedi2002: The ParseStringToBool() function does the exact same thing but ignores single letter inputs and digits. Is there a reason you prefer ParseBool() over this custom function?

x/ethbridge/keeper/msg_server.go
```diff
@@ -25,31 +25,46 @@ func NewMsgServerImpl(keeper Keeper) types.MsgServer {
...
 var _ types.MsgServer = msgServer{}
 
+func (srv msgServer) SetPauser(goCtx context.Context, msg *types.MsgPauser) (*types.MsgPauserResponse, error) {
+	response := &types.MsgPauserResponse{}
+	ctx := sdk.UnwrapSDKContext(goCtx)
+	signer, err := sdk.AccAddressFromBech32(msg.Signer)
+	if err != nil {
+		return response, err
+	}
+	if !srv.tokenRegistryKeeper.IsAdminAccount(ctx, tokenregistrytypes.AdminType_ETHBRIDGE, signer) {
```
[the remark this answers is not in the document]
Jedi2002: The admin functionality is concentrated in the token registry module as of now until we extract it out to it own separate module

x/ethbridge/keeper/pauser.go
```diff
@@ -0,0 +1,23 @@
+package keeper
+
+import (
+	"github.com/Sifchain/sifnode/x/ethbridge/types"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+)
+
+func (k Keeper) SetPauser(ctx sdk.Context, pauser *types.Pauser) {
+	store := ctx.KVStore(k.storeKey)
+	store.Set(types.PauserPrefix, k.cdc.MustMarshal(pauser))
```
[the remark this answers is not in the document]
Jedi2002: Just following the general pattern followed by most setter functions, can you think of any reason why it would be a problem?

x/ethbridge/keeper/msg_server_test.go
Jedi2002: We don't , there were just remenants from the original testcase I used as a template , which does a balance check at the end .

x/ethbridge/client/cli/tx.go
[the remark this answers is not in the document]
smartyalgo: If we explicitly don't want the single letter/digit behaviour, then this is fine. Otherwise, this is one less func we have to maintain.

x/ethbridge/keeper/pauser.go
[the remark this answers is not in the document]
smartyalgo: Ok. Just curious

x/ethbridge/keeper/msg_server.go
```diff
@@ -25,31 +25,46 @@ func NewMsgServerImpl(keeper Keeper) types.MsgServer {
 
 var _ types.MsgServer = msgServer{}
 
+func (srv msgServer) SetPauser(goCtx context.Context, msg *types.MsgPauser) (*types.MsgPauserResponse, error) {
+	response := &types.MsgPauserResponse{}
+	ctx := sdk.UnwrapSDKContext(goCtx)
+	signer, err := sdk.AccAddressFromBech32(msg.Signer)
```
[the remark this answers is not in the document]
Jedi2002: The signer pays for the fee, and we verify whether the signer is an admin.
If the person has the priv key for admin , and enough rowan for fees , yes they can send this tx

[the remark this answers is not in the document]
smartyalgo: Msg Signer is the signer used to sign the message.

smartyalgo (approved):

x/ethbridge/keeper/msg_server.go
```diff
@@ -24,26 +25,45 @@ func NewMsgServerImpl(keeper Keeper) types.MsgServer {
...
+	response := &types.MsgLockResponse{}
 	ctx := sdk.UnwrapSDKContext(goCtx)
+	if srv.Keeper.IsPaused(ctx) {
+		return response, types.ErrPaused
+	}
 	logger := srv.Keeper.Logger(ctx)
 	if srv.Keeper.ExistsPeggyToken(ctx, msg.Symbol) {
 		logger.Error("pegged token can't be lock.", "tokenSymbol", msg.Symbol)
-		return nil, errors.Errorf("pegged token %s can't be locked", msg.Symbol)
+		return response, errors.Errorf("pegged token %s can't be lock", msg.Symbol)
```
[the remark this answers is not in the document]
Jedi2002: The response only matters if err == nil ? 
do you see any problems with using an empty struct ?

x/ethbridge/types/msgs.go
```diff
@@ -17,6 +17,33 @@ const (
...
+
+// Route should return the name of the module
+func (msg MsgPauser) Route() string { return RouterKey }
+
+// Type should return the action
+func (msg MsgPauser) Type() string { return "pauser" }
+
+// ValidateBasic runs stateless checks on the message
+func (msg MsgPauser) ValidateBasic() error {
+	return nil
```
smartyalgo, line 30: What validations would we want in place?

x/ethbridge/keeper/msg_server_test.go
```diff
@@ -0,0 +1,99 @@
...
+import (
+	types2 "github.com/Sifchain/sifnode/x/admin/types"
+	keeper2 "github.com/Sifchain/sifnode/x/ethbridge/keeper"
+	"github.com/Sifchain/sifnode/x/ethbridge/test"
+	"github.com/Sifchain/sifnode/x/ethbridge/types"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/stretchr/testify/require"
+	"testing"
+)
+
```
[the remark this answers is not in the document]
smartyalgo: Test added

smartyalgo (approved):

juniuszhou:
looks good to me, just add an unit test, and removed unused 'fmt'.

juniuszhou (dismissed):

smartyalgo (approved):

x/ethbridge/keeper/msg_server.go
Brando753, line 52: Why does this report `lock` instead of `locked`

test/integration/framework/src/siftool/test_utils.py
```diff
@@ -358,12 +359,18 @@ def get_blocklist_sc(self):
         result = self.w3_conn.eth.contract(address=address, abi=abi)
         return result
 
-    def get_bridge_bank_sc(self):
+    def get_bridge_bank_sc(self) -> Contract:
```
Brando753, line 362: Can our typings be more specific then Contract? Like BridgeBankContract? I assume this Contract object gives access to methods unique to BridgeBank?

test/integration/framework/src/siftool/test_utils.py
```diff
@@ -358,12 +359,18 @@ def get_blocklist_sc(self):
...
         abi, _, address = self.abi_provider.get_descriptor("BridgeBank")
         assert address, "No address for BridgeBank"
         result = self.w3_conn.eth.contract(address=address, abi=abi)
         return result
 
+    def get_bridge_token_sc(self):
```
Brando753, line 368: This is missing a return type

test/integration/src/py/test_sifnode_peggy1_pause_unpause.py
```diff
@@ -0,0 +1,114 @@
+from typing import Tuple
```
Brando753, line 1: Why is this test labeled as a peggy1 test? Why would it not work on Peggy2

test/integration/src/py/test_sifnode_peggy1_pause_unpause.py
```diff
@@ -0,0 +1,114 @@
...
+import pytest
+
+import siftool_path
+from siftool import eth, test_utils, sifchain, cosmos
+from siftool.inflate_tokens import InflateTokens
+from siftool.common import *
+from siftool.test_utils import EnvCtx
+
+def test_pause_unpause_no_error(ctx: EnvCtx):
+    res = ctx.sifnode.pause_peggy_bridge(ctx.sifchain_ethbridge_admin_account)
```
Brando753, line 11: These return types should be better typed, I have no way of knowing what to expect the return type to look like without running it in a debugger.

Brando753 (changes requested):

test/integration/src/py/test_sifnode_peggy1_pause_unpause.py
```diff
@@ -0,0 +1,114 @@
...
+def test_pause_burn_valid(ctx: EnvCtx):
+    fund_amount_sif = 10 * test_utils.sifnode_funds_for_transfer_peggy1
+    fund_amount_eth = 1 * eth.ETH
+
+    test_sif_account = ctx.create_sifchain_addr(fund_amounts=[[fund_amount_sif, "rowan"]])
+    ctx.tx_bridge_bank_lock_eth(ctx.eth_faucet, test_sif_account, fund_amount_eth)
+    ctx.eth.advance_blocks(100)
+    # Setup is complete, test account has rowan AND eth
+    # Sleep for relayer to complete transaction
+    time.sleep(5)
```
Brando753, line 65: Do we not have a way to wait on balance change with timeout rather then just sleeping?

x/ethbridge/client/cli/tx.go
```diff
@@ -417,3 +418,35 @@ func GetCmdSetBlacklist() *cobra.Command {
 
 	return cmd
 }
+
+func GetCmdPauser() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "set-pauser [pause]",
```
Brando753, line 424: Why is this called set-pauser? This appears to pause or unpause not add a pauser roll.

x/ethbridge/keeper/msg_server.go
```diff
@@ -72,33 +92,36 @@ func (srv msgServer) Lock(goCtx context.Context, msg *types.MsgLock) (*types.Msg
...
 	logger := srv.Keeper.Logger(ctx)
-
+	if srv.Keeper.IsPaused(ctx) {
+		return response, types.ErrPaused
+	}
 	if !srv.Keeper.ExistsPeggyToken(ctx, msg.Symbol) {
 		logger.Error("Native token can't be burn.",
 			"tokenSymbol", msg.Symbol)
-		return nil, errors.Errorf("native token %s can't be burned", msg.Symbol)
+		return response, errors.Errorf("native token %s can't be burn", msg.Symbol)
```
Brando753, line 108: Again why did we make this grammatically incorrect?

x/ethbridge/keeper/msg_server.go
```diff
@@ -72,33 +92,36 @@ func (srv msgServer) Lock(goCtx context.Context, msg *types.MsgLock) (*types.Msg
...
 	if err != nil {
-		return nil, err
+		return response, err
 	}
 
 	account := srv.Keeper.accountKeeper.GetAccount(ctx, cosmosSender)
 	if account == nil {
 		logger.Error("account is nil.", "CosmosSender", msg.CosmosSender)
-		return nil, sdkerrors.Wrap(sdkerrors.ErrInvalidAddress, msg.CosmosSender)
+		return response, sdkerrors.Wrap(sdkerrors.ErrInvalidAddress, msg.CosmosSender)
```
Brando753, line 119: what was the reason to change all of these from nil to response objects?

x/ethbridge/keeper/msg_server_test.go
```diff
@@ -0,0 +1,119 @@
...
+		IsPaused: true,
+	}
+	msgPause := types.MsgPauser{
+		Signer:   admin.String(),
+		IsPaused: true,
+	}
+	msgUnPause := types.MsgPauser{
+		Signer:   admin.String(),
+		IsPaused: false,
+	}
```
Brando753, line 57: Should we test setting pause to false when its currently not paused to verify that the corner case works? So not paused, set pause to false, lock and burn still work.

x/ethbridge/keeper/pauser.go
```diff
@@ -0,0 +1,23 @@
+package keeper
+
+import (
+	"github.com/Sifchain/sifnode/x/ethbridge/types"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+)
+
+func (k Keeper) SetPauser(ctx sdk.Context, pauser *types.Pauser) {
```
Brando753, line 8: This needs to be renamed, it is not storing who is a pauser but the status of the chain as being paused or not.

x/ethbridge/keeper/pauser.go
Brando753, line 10: What is the pauser key? Is this key value an empty struct?

Brando753 (changes requested):

x/ethbridge/keeper/pauser.go
```diff
@@ -0,0 +1,23 @@
...
+	store := ctx.KVStore(k.storeKey)
+	store.Set(types.PauserPrefix, k.cdc.MustMarshal(pauser))
+}
+
+func (k Keeper) getPauser(ctx sdk.Context) *types.Pauser {
```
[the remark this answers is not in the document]
Brando753: How was this resolved?

x/ethbridge/types/msgs.go
Brando753, line 30: should we verify signer != nil?

test/integration/framework/src/siftool/sifchain.py
```diff
@@ -535,6 +539,32 @@ def validate_genesis(self):
         res = exactly_one(stdout(res).splitlines())
         assert res.endswith(" is a valid genesis file")
 
+    # Pause the ethbridge module's Lock/Burn on an evm_network_descriptor
+    def pause_peggy_bridge(self, admin_account_address) -> List[Mapping[str, Any]]:
```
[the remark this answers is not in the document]
jzvikart: There is already a `JsonDict` type for this purpose.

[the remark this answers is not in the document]
Brando753: which creates a dictionary in runtime right? The whole point in having a type is that we can explicitly know what members will belong in the result for use. It could only have the fields we need but it should not be magic like `res[0]['code']`

x/ethbridge/types/msgs.go
juniuszhou: could add signer is empty check

x/ethbridge/keeper/pauser.go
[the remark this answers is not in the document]
juniuszhou: no panic, just get the default value. you already get it in unit test.

test/integration/src/py/test_sifnode_peggy1_pause_unpause.py
smartyalgo: Agreed, updated

juniuszhou (approved):

test/integration/framework/src/siftool/test_utils.py
smartyalgo: The web3.eth method returns a Base contract object, which has a `Caller` object you can use for something like this 
```    
contract = ctx.get_bridge_bank_sc()
output = contract.caller.hasBlocklist()
```
This code runs but
Type hint is not available because the functions are read from abi and added as part of `__init__`

test/integration/src/py/test_sifnode_peggy_pause_unpause.py
```diff
@@ -0,0 +1,114 @@
...
+    else:
+        eth_balance_before = ctx.eth.get_eth_balance(test_eth_destination_account)
+
+    ctx.sifnode_client.send_from_sifchain_to_ethereum(test_sif_account, test_eth_destination_account, send_amount, denom)
+
+    sif_balance_after = ctx.sifnode.wait_for_balance_change(test_sif_account, sif_balance_before)
+    try:
+        eth_balance_after = ctx.wait_for_eth_balance_change(test_eth_destination_account, eth_balance_before, token_addr=erc20_token_addr, timeout=30)
+    except Exception as e:
+        eth_balance_after = eth_balance_before
```
Brando753, line 111: Should we not fetch the user balance rather than assuming it has not changed if an exception is raised?

x/ethbridge/types/msgs.go
```diff
@@ -17,6 +17,36 @@ const (
...
+}
+
+// GetSignBytes encodes the message for signing
+func (msg MsgPause) GetSignBytes() []byte {
+	return sdk.MustSortJSON(ModuleCdc.MustMarshalJSON(&msg))
+}
+
+// GetSigners defines whose signature is required
+func (msg MsgPause) GetSigners() []sdk.AccAddress {
+	addr, err := sdk.AccAddressFromBech32(msg.Signer)
```
Brando753, line 43: We should probably use MustAccAddressFromBech32 here as it panics on failure rather than rewriting it ourselves

Brando753 (approved):
There are still things that should change in the tests, and minor code refactors that I have commented about. However, the code looks fine for merge, and I have approved the PR so that we can cut the next release once you think everything is good @smartyalgo.

x/ethbridge/types/msgs.go
Brando753: @smartyalgo I Just realized the integration tests using siftool may not be included in the automated tests for the PR. Can you verify the .github file has been updated to use the new siftool tests?

juniuszhou (dismissed):

test/integration/src/py/test_sifnode_peggy_pause_unpause.py
smartyalgo: `wait_for_eth_balance_change` will always return value != old_balance, or it will raise Exception. When we receive an exception we know we've timed out, and new_balance == old_balance

One improvement is to assert it is a Timedout exception. Jure and I are working on that 

This function will be moved into siftool as well as part of send. It wont return until balance is changed

x/ethbridge/types/msgs.go
smartyalgo: Updated to use MustAccAddressFromBech32.
Added to github action

Brando753 (approved):

Brando753 (approved):

Brando753 (approved):
All tests passing locally; we should still make those requested changes but it should be good for now.

PR #2678, opened 2022-04-26
Peggy2 Upgrade Scripts

Brando753:
Scripts to upgrade the Bridgebank and Cosmos Bridge to Peggy2 versions on mainnet. WIP.


Brando753:
This code should support a Peggy 2 migration, however, it's reporting that the contracts are not upgrade safe. It could be that more work is needed on this but I want to get the code changes in as I continue on this as it touches a lot of files.

PR #2683, opened 2022-04-27
Load test for token lock/burn in sifnode

juniuszhou:


test/load/test_load_tx_ethbridge_burn.py
```diff
@@ -71,25 +145,57 @@ def _test_load_tx_ethbridge_burn(ctx: test_utils.EnvCtx, amount_per_tx: int, tra
...
+
 
     # Create a dispensation sif account that will receive all locked ETH and dispense it to each sif account
     # (we do this in one transaction because lock transactions take a lot of time).
     # Dispensation account needs rowan for distributing ceth to sif_accts.
-    dispensation_sif_acct: cosmos.Address = ctx.create_sifchain_addr(fund_amounts={rowan: n_sif * sif_tx_fee_in_rowan})
+    if token_denom == rowan:
+        amount = sum_all * amount_per_tx + n_sif * 2 * sif_tx_fee_in_rowan
+        dispensation_sif_acct: cosmos.Address = ctx.create_sifchain_addr(fund_amounts={rowan: amount})
+    elif token_denom == ctx.ceth_symbol:
```
jprupp, line 160: That ceth symbol there, what token does it represent?

Is it cETH from Compound, does it represent the native token of whatever chain we are interacting with, or something else?

test/load/test_load_tx_ethbridge_burn.py
```diff
@@ -22,30 +23,103 @@
...
+            [100, 100, 100],
+            [100, 100, 100],
+            [100, 100, 100],
+            [10, 20, 30],
+        ]
+    amount_per_tx = 1000100101
+    eth_sc = ctx.get_generic_erc20_sc(NULL_ADDRESS)
+    _test_load_tx_ethbridge_lock_burn(ctx, amount_per_tx, transfer_table, eth_sc)
+
+def _test_load_tx_ethbridge_lock_burn(ctx: test_utils.EnvCtx, amount_per_tx: int,
```
jprupp, line 115: It would be very helpful for me if I could use the code in this test to understand better the code that is being tested. I tried to read this procedure, taking notes in a piece of paper, but two hours later I gave up. It's too complicated.

jprupp (changes requested):
If you could refactor the test code to make it possible to understand the test in isolation, without being familiar with the code being tested, it would be possible for me to give you a better review. I did spent a couple of hours trying to understand it, but didn't get the picture.

test/load/test_load_tx_ethbridge_burn.py
```diff
@@ -22,30 +23,103 @@
 # E0326 11:41:27.125006742  636480 fork_posix.cc:70]           Fork support is only compatible with the epoll1 and poll polling strategies
 # Maybe: https://github.com/grpc/grpc/issues/29044
 
+eth_account_number = 2
+sif_account_number = 1
+transaction_number = 2
+
+# build the transfer matrix
```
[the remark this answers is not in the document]
juniuszhou: To run load test, you may set these number as different value. For example, 1000 ethereum account. It is hard to write such matrix manually.

test/load/test_load_tx_ethbridge_burn.py
```diff
@@ -22,30 +23,103 @@
...
+    test_sif_account = ctx.create_sifchain_addr(fund_amounts={rowan:fund_amount_sif, ctx.ceth_symbol:fund_amount_eth})
+    test_eth_account = ctx.create_and_fund_eth_account(fund_amount=fund_amount_eth)
+    ctx.sifnode_client.send_from_sifchain_to_ethereum(test_sif_account, test_eth_account, rowan_transfer_amount, rowan)
+    ctx.advance_blocks()
+
+    # wait for erc20 contract created
+    token_address = ctx.wait_for_new_bridge_token_created(rowan)
+    return token_address
+
+# test single sif account burn erc20 to multiple ethereum accounts
```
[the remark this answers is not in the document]
juniuszhou: rowan is native token in sifnode side. ERC20 means the bridge token created for rowan.

test/load/test_load_tx_ethbridge_burn.py
juniuszhou: The denom is from peggy1.0. We just add 'c' as prefix for any token imported from Ethereum. Ceth represents the eth in sifnode.

test/load/test_load_tx_ethbridge_burn.py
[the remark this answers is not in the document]
jzvikart: You can use list comprehensions: `[[2 for y in range(1000)] for x in range(1000)]`.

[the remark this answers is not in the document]
jprupp: I did this in the python REPL:

```python
[[2 for _ in range(3)] for _ in range(10)]
```

That's cool because if you don't need the value, you can just replace it with an underscore. It makes code easier to understand.

test/load/test_load_tx_ethbridge_burn.py
```diff
@@ -22,30 +23,97 @@
...
+def test_single_sif_to_multiple_eth_account_lock_rowan(ctx: test_utils.EnvCtx):
+    # get rowan contract address
+    rowan_token_address = burn_rowan_get_erc20_address(ctx)
+
+    print("+++++++++++++++++++++++++ ", rowan_token_address)
```
jzvikart, line 58: This probably needs to be cleaned up? If you want to keep it, it's better to use logging.

test/load/test_load_tx_ethbridge_burn.py
```diff
@@ -71,25 +139,58 @@ def _test_load_tx_ethbridge_burn(ctx: test_utils.EnvCtx, amount_per_tx: int, tra
...
+    lock_eth_amount = sif_tx_burn_fee_in_ceth
+    if token_denom == ctx.ceth_symbol:
+        lock_eth_amount += amount_per_tx
+    ctx.bridge_bank_lock_eth(ctx.operator, dispensation_sif_acct, sum_all * lock_eth_amount)
     ctx.advance_blocks()
-    new_balances = ctx.wait_for_sif_balance_change(dispensation_sif_acct, old_balances)
+    old_balances = ctx.wait_for_sif_balance_change(dispensation_sif_acct, old_balances)
+
+    # just for erc20 token
+    if token_address != None and not isRowan:
```
jzvikart, line 170: While `token_address != None` will work, the correct (idiomatic) way is to use `token_address is not None`.

test/load/test_load_tx_ethbridge_burn.py
```diff
@@ -71,25 +139,58 @@ def _test_load_tx_ethbridge_burn(ctx: test_utils.EnvCtx, amount_per_tx: int, tra
...
     # Dispense from sif_dispensation_acct to sif_accts
     for i, sif_acct in enumerate(sif_accts):
         b_sif_acct_before = ctx.get_sifchain_balance(sif_acct)
         b_disp_acct_before = ctx.get_sifchain_balance(dispensation_sif_acct)
-        amount_ceth = sum_sif[i] * (amount_per_tx + sif_tx_burn_fee_in_ceth)
-        ctx.send_from_sifchain_to_sifchain(dispensation_sif_acct, sif_acct, {ctx.ceth_symbol: amount_ceth})
-        b_sif_acct_after = ctx.wait_for_sif_balance_change(sif_acct, b_sif_acct_before)
+
+        # if token is ceth, combine and fund
+        if token_address == None:
```
jzvikart, line 181: `token_address is None`

test/load/test_load_tx_ethbridge_burn.py
```diff
@@ -22,30 +23,97 @@
...
+            [100, 100, 100],
+            [100, 100, 100],
+            [100, 100, 100],
+            [10, 20, 30],
+        ]
+    amount_per_tx = 1000100101
+    _test_load_tx_ethbridge_lock_burn(ctx, amount_per_tx, transfer_table, None)
+
+def _test_load_tx_ethbridge_lock_burn(ctx: test_utils.EnvCtx, amount_per_tx: int, 
+    transfer_table: List[List[int]], token_address: Address, isRowan: bool = False, randomize: bool = None):
```
jzvikart, line 108: `token_address: Optional[Address]` is better since it can be null.
Or (depending on style preferences) also `Union[Address, None]`

test/load/test_load_tx_ethbridge_burn.py
```diff
@@ -71,25 +139,58 @@ def _test_load_tx_ethbridge_burn(ctx: test_utils.EnvCtx, amount_per_tx: int, tra
     sif_balances_initial: List[cosmos.Balance] = [ctx.get_sifchain_balance(sif_acct) for sif_acct in sif_accts]
     eth_balances_initial: List[str] = [ctx.eth.get_eth_balance(eth_acct) for eth_acct in eth_accts]
     assert all([b == 0 for b in eth_balances_initial])  # Might be non-zero if we're recycling accounts
+    
+    if token_address != None:
```
jzvikart, line 143: Also here

test/load/test_load_tx_ethbridge_burn.py
```diff
@@ -171,12 +272,18 @@ def sif_acct_sender_fn(sif_acct: cosmos.Address, tx_stub: cosmos_tx_grpc.Service
     last_change_time = None
     last_change = None
     last_change_timeout = 90
-    cumulative_timeout = sum_all * 10  # Equivalent to min rate of 0.1 tps
+    cumulative_timeout = 30 + sum_all * 10  # Equivalent to min rate of 0.1 tps
     while True:
-        eth_balances = [ctx.eth.get_eth_balance(eth_acct) for eth_acct in eth_accts]
-        balance_delta = sum([eth_balances[i] - eth_balances_initial[i] for i in range(n_eth)])
+        if token_address == None:
```
jzvikart, line 277: Also here

test/load/test_load_tx_ethbridge_burn.py
```diff
@@ -71,25 +139,58 @@ def _test_load_tx_ethbridge_burn(ctx: test_utils.EnvCtx, amount_per_tx: int, tra
...
+    ctx.bridge_bank_lock_eth(ctx.operator, dispensation_sif_acct, sum_all * lock_eth_amount)
     ctx.advance_blocks()
-    new_balances = ctx.wait_for_sif_balance_change(dispensation_sif_acct, old_balances)
+    old_balances = ctx.wait_for_sif_balance_change(dispensation_sif_acct, old_balances)
+
+    # just for erc20 token
+    if token_address != None and not isRowan:
+        token_sc = ctx.get_generic_erc20_sc(token_address)
+        ctx.send_from_ethereum_to_sifchain(ctx.operator, dispensation_sif_acct, sum_all * amount_per_tx, token_sc=token_sc, isLock=True)
+        _ = ctx.wait_for_sif_balance_change(dispensation_sif_acct, old_balances)
```
jzvikart, line 173: If you don't need the return value, it's better to do just

```
ctx.wait_for_sif_balance_change(dispensation_sif_acct, old_balances)
```

jzvikart (approved):
There are still some things that can be improved (see my comments below), but they are mostly stylistic and do not affect the test itself so it's OK with me.

test/load/test_load_tx_ethbridge_burn.py
jprupp: This test seems to have multiple stages, which should be separated in their own named procedures:

- Validation of inputs.
- Setup of test, which itself can be split into multiple stages:
  - Creation and funding of accounts on the Sifchain side.
  - Creation and funding of accounts on the Ethereum side.
- One or more of these actions:
  - Execution of test transactions.
  - Validation of results of test transactions.

It is much easier to understand that way, instead of having one long procedure where all these stages are mushed together.

A comment like this where the `transfer_table` parameter of the function is declared would have helped:

> Rows are Sifchain accounts, Columns are Ethereum accounts, numbers are the count of transactions to perform for each pair.

I thought for a long while that those numbers represented amounts of a token, and nothing made much sense.

test/load/test_load_tx_ethbridge_burn.py
```diff
@@ -22,30 +23,97 @@
...
+        ]
+    amount_per_tx = 1000100101
+    _test_load_tx_ethbridge_lock_burn(ctx, amount_per_tx, transfer_table, None)
+
+def _test_load_tx_ethbridge_lock_burn(ctx: test_utils.EnvCtx, amount_per_tx: int, 
+    transfer_table: List[List[int]], token_address: Address, isRowan: bool = False, randomize: bool = None):
+    # rowan is natvie token, denom not from contract in Ethereum
+    if isRowan:
+        token_denom = rowan
+    elif token_address == None:
```
jprupp, line 112: This test is used in multiple places in the code, and it is not necessarily obvious that `token_address == None` means that the token being transacted is CETH, except rigt here.

The code below that uses this test could use instead `token_denom == ctx.ceth_symbol` if such things are comparable like that. Else it would be convenient to set a `ceth_flag = true` upon discovering that `token_address == None`, and then refer to that flag where there's code that depends on the token being CETH.

Another possibility is to simply pass `token_denom` as an argument, and then simplify the function's signature such that instead of receiving both a `token_address` and a `isRowan` argument, that it may only have a `token_denom` argument, as long as `token_address` or `isRowan` can be extracted as internal flags from `token_denom`, when necessary.

jprupp (approved):
I will not request further changes at this time. The comments that I wrote about this code readability apply to the contents of this file even before any of the changes in this PR were done, and that would require a refactor to address that isn't in scope for this PR.

PR #2704, opened 2022-05-02
Missing cosmosDenom when building the prophecy id

juniuszhou:
End to end test passed.


Brando753:
Heads up @juniuszhou This looks good but I want to add some tests as well and remove unnecessary fields from the claimdata such as symbol, name, and decimals before we merge this in.

x/ethbridge/types/msgs.go
```diff
@@ -207,29 +242,45 @@ func ComputeProphecyID(cosmosSender string, sequence uint64, ethereumReceiver st
...
+			Type: uint8Ty,
+		},
+		{
+			Type: int32Ty,
+		},
+		{
+			Type: boolTy,
+		},
+		{
+			Type: uint256Ty,
```
Brando753, line 264: the claim data struct in solidity labels nonce as a uint128 why did you change this to uint256?

Brando753 (changes requested):
Once I understand this one comment change I will merge.

x/ethbridge/types/msgs.go
juniuszhou: there is no uint128 type in go, so I use the big.Int as big.NewInt(int64(globalNonce)). In abi argument, we need a uint256 to pack it.

x/ethbridge/types/msgs.go
```diff
@@ -79,16 +79,28 @@ func (msg MsgLock) GetSigners() []sdk.AccAddress {
 }
 
 // GetProphecyID get prophecy ID for lock message
-func (msg MsgLock) GetProphecyID(doublePeggy bool, sequence, globalNonce uint64, tokenAddress string) []byte {
+func (msg MsgLock) GetProphecyID(
```
smartyalgo, line 82: Not required for this PR.

Have we decided against merging these two functions?

juniuszhou: you means GetProphecyID for MsgLock and MsgBurn?

smart-contracts/contracts/CosmosBridge.sol
```diff
@@ -132,12 +132,20 @@ contract CosmosBridge is CosmosBridgeStorage, Oracle {
    * @param _networkDescriptor Unique identifier of the network
```
smartyalgo, line 132: Comments outdated

x/ethbridge/types/msgs.go
Brando753: without the generics support, let's just leave it as is.

Brando753 (approved):

kleimanb81 (commented):
Nice job, quick response!

PR #2706, opened 2022-05-02
Feature/cancel unbond

timlind:


x/clp/keeper/msg_server.go
```diff
@@ -37,6 +37,45 @@ func (k msgServer) SetSymmetryThreshold(goCtx context.Context, threshold *types.
...
+	ctx := sdk.UnwrapSDKContext(goCtx)
+	lp, err := k.Keeper.GetLiquidityProvider(ctx, request.ExternalAsset.Symbol, request.Signer)
+	if err != nil {
+		return nil, types.ErrLiquidityProviderDoesNotExist
+	}
+	// Prune unlocks
+	params := k.GetRewardsParams(ctx)
+	k.PruneUnlockRecords(ctx, &lp, params.LiquidityRemovalLockPeriod, params.LiquidityRemovalCancelPeriod)
+
+	err = k.UseUnlockedLiquidity(ctx, lp, request.Units)
```
timlind, line 58: This needs to use unbond-records in any state.

AustinoBombino (commented):
logic looks good to me.

do we need a rest tx for this?

canercandan (approved):
LGTM

timlind:
Looks like we skipped the rest endpoint for unbond as well so unless FE specifically needs it we should be good without.

PR #2726, opened 2022-05-05
store both BridgeBank and CosmosBridge addresses

juniuszhou:
Now, relayer queries the BridgeBank and CosmosBridge addresses each time. we can store them in memory.


Brando753 (approved):

cmd/ebrelayer/txs/registry.go
```diff
@@ -13,44 +13,51 @@ import (
...
+	case BridgeBank:
+		if bridgeBankAddress.Hex() != nullAddress {
+			return bridgeBankAddress, nil
+		}
+	case CosmosBridge:
+		if cosmosBridgeAddress.Hex() != nullAddress {
+			return cosmosBridgeAddress, nil
+		}
+	default:
+		panic("invalid target contract address")
```
smartyalgo, line 48: Why do we panic instead of throwing error here?

cmd/ebrelayer/txs/registry.go
```diff
@@ -13,44 +13,51 @@ import (
...
+	return [...]string{"bridgebank", "cosmosbridge"}[d-1]
 }
 
 // GetAddressFromBridgeRegistry queries the requested contract address from the BridgeRegistry contract
 func GetAddressFromBridgeRegistry(client *ethclient.Client, registry common.Address, target ContractRegistry,
 	sugaredLogger *zap.SugaredLogger) (common.Address, error) {
+	// Return address if already got and stored
+	switch target {
+	case BridgeBank:
+		if bridgeBankAddress.Hex() != nullAddress {
```
smartyalgo, line 40: I think the current code works fine, so the following idea isnt required for this PR

I would consider packing L52 - 84 into it's own function, call it `func lazyloadBridgeBankAddress()`, which loads value for bridgebankAddress and returns void, and also returns immediately if bridgebankAddress is not null. 
```
func lazyloadBridgeBankAddress(registry common.Address, client *ethclient.Client) {
    if bridgeBankAddress.Hex() != nullAddress {
        // We have already loaded bridgebank's address, no need to load again
        return nil
    }
    // Logic from L52 - 84, interact with bridgeregistry.sol to get address and memoize
}
```


Then in this switch statement, it would look like
```
switch target {
case BridgeBank:
    if err := lazyloadBridgeBank(registry); err != nil {
        return bridgeBankAddress, nil
    } else {
        return common.Address{}, err
    }
case CosmosBridge:
// Same thing
}
```

The benefit is 
- you can test the load functions separately, 
- All of the errors are inside the load function 
- The case statement is the place that returns the value for this function, so it's easier to read

smartyalgo (commented):
Added comments to panic behaviour for `GetAddressFromBridgeRegistry`

cmd/ebrelayer/txs/registry.go
juniuszhou: The relayer should never go to this branch, and panic can make relayer exit, then we can fix the issue. If return error, relayer can do nothing, and this error not depends on any runtime data, so it will be definitely failed again. the relayer can't work normally anymore.

PR #2737, opened 2022-05-09
Separate unpauser

jprupp:
I'm pretty sure there's all sorts of wrong here, but please take a look anyways.


kleimanb81 (approved):

kleimanb81 (commented):
Very nice start , booting up my PC

Brando753:
So on review, this task may have been a lot harder ask than initially expected. The BridgeBank is Pausable, Pausable is PauserRole and PauserRole has no gap defined... which means while this change is technically correct if this was a new deployment it is not upgrade safe in its current form. Without the gap defined, I believe the storage layout will change between compiles. Will investigate how to best resolve this.

Brando753:
Rocmmend closing out this PR in favor of #3298

PR #2746, opened 2022-05-11
Improved code style in ante handler

begmaroman:


app/ante/ante.go
```diff
@@ -56,52 +88,54 @@ func NewAdjustGasPriceDecorator() AdjustGasPriceDecorator {
 // AnteHandle adjusts the gas price based on the tx type.
 func (r AdjustGasPriceDecorator) AnteHandle(ctx sdk.Context, tx sdk.Tx, simulate bool, next sdk.AnteHandler) (newCtx sdk.Context, err error) {
 	msgs := tx.GetMsgs()
-	if len(msgs) == 1 && (strings.Contains(strings.ToLower(sdk.MsgTypeURL(msgs[0])), strings.ToLower(disptypes.MsgTypeCreateDistribution)) ||
-		strings.Contains(strings.ToLower(sdk.MsgTypeURL(msgs[0])), strings.ToLower(disptypes.MsgTypeRunDistribution))) {
+	if len(msgs) == 1 && slicex.StringsContain(strings.ToLower(sdk.MsgTypeURL(msgs[0])), distributionMessageTypes) {
```
sifag, line 91: How about something like that
`if len(msgs) == 1 && isDistributedMsg(msgs[0])`
where
`func isDistributedMsg(msg sdk.Msg) bool {
     return slicex.StringsContain(strings.ToLower(sdk.MsgTypeURL(msg)), distributionMessageTypes)
}`

Sorry for the terrible formatting

app/ante/ante.go
```diff
@@ -56,52 +88,54 @@ func NewAdjustGasPriceDecorator() AdjustGasPriceDecorator {
...
 	for i := range msgs {
 		msgTypeURLLower := strings.ToLower(sdk.MsgTypeURL(msgs[i]))
-		if strings.Contains(msgTypeURLLower, strings.ToLower(banktypes.TypeMsgSend)) ||
-			strings.Contains(msgTypeURLLower, strings.ToLower(banktypes.TypeMsgMultiSend)) ||
-			strings.Contains(msgTypeURLLower, "createuserclaim") ||
-			strings.Contains(msgTypeURLLower, "swap") ||
-			strings.Contains(msgTypeURLLower, "removeliquidity") ||
-			strings.Contains(msgTypeURLLower, "removeliquidityunits") ||
-			strings.Contains(msgTypeURLLower, "addliquidity") {
+		if slicex.StringsContain(msgTypeURLLower, regularMessageTypes) {
```
sifag, line 108: Same here `isRegularMessage`

tools/slicex/slicex.go
```diff
@@ -0,0 +1,13 @@
+package slicex
+
+import "strings"
+
+// StringsContain returns true if the given value contains at least one value from the given slice
+func StringsContain(val string, sl []string) bool {
+	for _, sv := range sl {
+		if strings.Contains(val, sv) {
```
sifag, line 8: I think it's better to match for the exact string we are looking for instead of accepting strings that contain the string as a substring. I can see that in this use case it doesn't matter but I'd rather be safe

sifag (changes requested):
LGTM overall

x/clp/keeper/calculations_test.go
```diff
@@ -886,8 +886,8 @@ func TestKeeper_GetSwapFee(t *testing.T) {
 	nativeAssetAmount := sdk.NewUintFromString("998")
 	externalAssetAmount := sdk.NewUintFromString("998")
 	asset := types.NewAsset("eth")
-	externalCoin := sdk.NewCoin(asset.Symbol, sdk.Int(sdk.NewUint(10000)))
-	nativeCoin := sdk.NewCoin(types.NativeSymbol, sdk.Int(sdk.NewUint(10000)))
+	externalCoin := sdk.NewCoin(asset.Symbol, sdk.NewIntFromUint64(10000))
+	nativeCoin := sdk.NewCoin(types.NativeSymbol, sdk.NewIntFromUint64(10000))
 	err := sifapp.AddCoinsToAccount(types.ModuleName, app.BankKeeper, ctx, signer, sdk.NewCoins(externalCoin, nativeCoin))
```
joneskm, line 891: Probably best to add these changes in a separate PR since they're not related to the ante handler. You could even PR these against the feature/clp-refactor branch.

joneskm (changes requested):

tools/slicex/slicex.go
begmaroman: But the original logic uses `strings.Contain` function to find substrings. Don't really want to change an actual business logic within this PR.

app/ante/ante.go
begmaroman: Good point, done

app/ante/ante.go
begmaroman: Good point, done

tools/slicex/slicex.go
sifag: You are right!

sifag (approved):

PR #2752, opened 2022-05-12
Remove whitelist flag in metadata

juniuszhou:


smartyalgo:
Would `x/tokenregistry/keeper/keeper.go::GetRegistryEntry` fail because existing entry has `is_whitelisted`, and `MustUnmarshal` fails with panic?

smartyalgo:
Theres `is_whitelisted` references in siftool. We should remove that too

x/tokenregistry/keeper/metadata.go
```diff
@@ -69,18 +69,14 @@ func (k keeper) AddTokenMetadata(ctx sdk.Context, metadata types.TokenMetadata)
 		entry = &types.RegistryEntry{}
```
smartyalgo, line 69: Comment at L60 outdated

smartyalgo (changes requested):

proto/sifnode/tokenregistry/v1/types.proto
```diff
@@ -21,20 +21,19 @@ enum Permission {
 }
 
 message RegistryEntry {
-  bool is_whitelisted = 1;
```
Brando753, line 24: Has this flag ever been stored in the production tokenRegistry? If so, we cannot remove it, only rename it.

Brando753: Looking at develop branch the registry entry looks like
```proto
message RegistryEntry {
  reserved 1;
  int64 decimals = 2;
  string denom = 3;
  string base_denom = 4;
  string path = 5;
  string ibc_channel_id = 6;
  string ibc_counterparty_channel_id = 7;
  string display_name = 8;
  string display_symbol = 9;
  string network = 10;
  string address = 11;
  string external_symbol = 12;
  string transfer_limit = 13;
  repeated Permission permissions = 15;
  ```
so we cant just delete this entry can we?

Brando753 (changes requested):
Need clarity on if this field needs to be deprecated or if its save to remove.

proto/sifnode/tokenregistry/v1/types.proto
juniuszhou: I am not sure if it is used in the production. I checked the master branch, the is_whitelisted also removed and  set it as reserved. We have different approached to handle it, via data migration, or add v2 for our proto buf. For now, I will follow the develop and master branch, set it as reserved.

smartyalgo (approved):

PR #2766, opened 2022-05-16
Siftool: replace hardhat with geth

jzvikart:
Closes #720.

To get siftool running with geth instead of hardhat, use `siftool run-env --geth`.


khdegraaf (approved):
LGTM

juniuszhou (approved):

PR #2767, opened 2022-05-17
v0.13.3

timlind:


x/clp/keeper/calculations.go
```diff
@@ -286,6 +287,12 @@ func CalculatePoolUnits(oldPoolUnits, nativeAssetBalance, externalAssetBalance,
 		return sdk.ZeroUint(), sdk.ZeroUint(), types.ErrAsymmetricAdd
 	}
 
+	ratioDiff := R.Quo(A).Sub(r.Quo(a)).Abs()
```
canercandan, line 290: @timlind I believe the formula should be instead the following: 

```
|(X/Y) - (x/y)| > t 
```

where `X` is the external asset balance, `Y` native asset balance and x and y external and native amount which means it should translate to:

```
ratioDiff := A.Quo(R).Sub(a.Quo(r)).Abs()
```

correct?

x/clp/keeper/calculations.go
```diff
@@ -286,6 +287,12 @@ func CalculatePoolUnits(oldPoolUnits, nativeAssetBalance, externalAssetBalance,
...
+
+	if ratioDiff.GT(ratioThreshold) {
+		return sdk.ZeroUint(), sdk.ZeroUint(), types.ErrAsymmetricAdd
```
canercandan, line 293: @timlind what do you think of using a different error message for the ratio asymetric add error?

canercandan (commented):
* implemented formula requires some changes
* might need to add a new error message for the ratio asymmetric add error

x/clp/keeper/calculations.go
timlind: Updated.

x/clp/keeper/calculations.go
timlind: Can do.

canercandan (dismissed):
LGTM thanks

x/clp/keeper/calculations.go
```diff
@@ -286,6 +287,12 @@ func CalculatePoolUnits(oldPoolUnits, nativeAssetBalance, externalAssetBalance,
 		return sdk.ZeroUint(), sdk.ZeroUint(), types.ErrAsymmetricAdd
```
AustinoBombino, line 287: Should the ratio change and ratioThreshold replace the old slipadjust check and symmetryThreshold?

marshsif (commented):
It looks like this PR also includes new token registry configs. Is that supposed to be the case? Should we add that as a feature for this release?

joneskm:
LGTM

joneskm (approved):

sifag (approved):

PR #2769, opened 2022-05-17
Fix a bug for remove outdated prophecy

juniuszhou:
The code change includes:
1. fix the block number comparison bug
2. add the test cases for prophecy in keeper
3. add the prophecy removal frequency, to avoid iterate all prophecies in every block


x/oracle/keeper/prophecy.go
```diff
@@ -140,14 +142,18 @@ func (k Keeper) AppendSignature(ctx sdk.Context, prophecyID []byte, ethereumAddr
 
 // CleanUpProphecy clean up outdated prophecy
 func (k Keeper) CleanUpProphecy(ctx sdk.Context) {
+	// it is low efficient to check outdated prophecy each block
+	if k.currentHeight%CleanUpFrequency != 0 {
```
Brando753, line 146: Please put spaces between your operators and variables; I assume this is modulus operator. This just reduces the frequency this gets called correct? Cant this result in going very long times without a cleanup if the function calling this is always a block offset from 1000?

x/oracle/keeper/prophecy.go
```diff
@@ -140,14 +142,18 @@ func (k Keeper) AppendSignature(ctx sdk.Context, prophecyID []byte, ethereumAddr
...
+	}
 	var prophecyInfo types.ProphecyInfo
 	currentHeight := uint64(k.currentHeight)
 
 	store := ctx.KVStore(k.storeKey)
 	iter := sdk.KVStorePrefixIterator(store, types.SignaturePrefix)
 	for ; iter.Valid(); iter.Next() {
 		k.cdc.MustUnmarshal(iter.Value(), &prophecyInfo)
-		if prophecyInfo.BlockNumber-currentHeight > ProphecyLiftTime {
+		if currentHeight > prophecyInfo.BlockNumber+ProphecyLiftTime {
```
Brando753, line 156: Please put a space between your operators

Brando753 (changes requested):

x/oracle/keeper/prophecy.go
juniuszhou: Yes, we need avoiding the keeper to check all prophecies in every block. Since the ProphecyLiftTime = 520000, 1000 just a small number compared with the life time.

juniuszhou: No, it actually works this way. If we have a prophecy for cross chain transfer in block X. Then from X + 52000, we may remove it since its ProphecyLiftTime. but it must be removed before X + 52000 + 1000. Which exact block depends on the last four digit of X.

x/oracle/keeper/prophecy.go
```diff
@@ -140,14 +142,18 @@ func (k Keeper) AppendSignature(ctx sdk.Context, prophecyID []byte, ethereumAddr
...
 	var prophecyInfo types.ProphecyInfo
 	currentHeight := uint64(k.currentHeight)
 
 	store := ctx.KVStore(k.storeKey)
 	iter := sdk.KVStorePrefixIterator(store, types.SignaturePrefix)
 	for ; iter.Valid(); iter.Next() {
 		k.cdc.MustUnmarshal(iter.Value(), &prophecyInfo)
-		if prophecyInfo.BlockNumber-currentHeight > ProphecyLiftTime {
+		if currentHeight > prophecyInfo.BlockNumber + ProphecyLiftTime {
 			storePrefix := append(types.SignaturePrefix, prophecyInfo.ProphecyId[:]...)
```
smartyalgo, line 157: I think we should put `append(types.SignaturePrefix, prophecyInfo.ProphecyId[:]...)` into it's own func, which takes `prophecyID` as parameter, for readability.

smartyalgo: Maybe in addition, put L157-158 into a `DeleteProphecyInfo` func. Same for L159-160. This makes it a lot more readable. 
It also lets us test the delete function, separate from the filter logic in L156

smartyalgo: Why don't we check if the prophecy status is `completed` instead of `pending`, before deleting?

smartyalgo (changes requested):

x/oracle/keeper/prophecy.go
juniuszhou: It is from requirement, if no enough relayers to finalize the prophecy, we just remove it even it is still pending.

smartyalgo: But in our case, the GlobalNonce is strictly increasing, ([enforced here ](https://github.com/Sifchain/sifnode/blob/future/peggy2/x/ethbridge/keeper/globalNonce.go#L39) ) so if a prophecy is removed instead of finalized, the bridge will still expect that nonce, thus the bridge halts. Am I wrong ?

juniuszhou: Yes, you are right. Since the lifeTime we set for prophecy is about one week. It is a very extreme condition for most of relayers have not worked for one week. so we don't check the status of prophecy when remove its signature and global nonce.

smartyalgo: Yes I agree this is an extremely unlikely event. If there's proposal thats unconfirmed for this long we have a much bigger problem. 
I think it is still safer to include checking if the status is completed, just because deleting a pending prophecy will require more manual fix. 
If you have bandwidth, I suggest do it now, if not, add a TODO.

smartyalgo (approved):

smartyalgo:
Added comment on pending prophecy, other than that, LGTM

x/oracle/keeper/prophecy.go
juniuszhou: will add the status check.

PR #2778, opened 2022-05-19
Fix unit test for smart contract

juniuszhou:


smart-contracts/contracts/CosmosBridge.sol
```diff
@@ -226,13 +227,16 @@ contract CosmosBridge is CosmosBridgeStorage, Oracle {
         pow += getValidatorPower(validator.signer);
       }
 
-      // Signatures must be sorted on the sifchain side, so we just
-      // need to make sure that the next validator in the array
+      // Signatures must be sorted on the relayer side, so we just
+      // need to make sure that the next witness in the array
```
jzvikart, line 231: What is the actual sorting key? (i.e. `validator.signer`)

juniuszhou: sorting key is signer, cosmosBridge need check if the duplicated signer in the signature list. sorted singer make it efficient to compare, gas saving.

jzvikart:
@Brando753 Shall we also add a test to check for missing `bridgeBank.addExistingBridgeToken()`?

juniuszhou:
we may even remove it, if don't need to do such operation in the future. all bridge tokens will be created and put into whitelist automatically.

jzvikart:
Looks good to me!

jzvikart (approved):

juniuszhou:
#2793 created to discuss the addExistingBridgeToken method.

PR #2791, opened 2022-05-21
Specify version for proto-gen Docker image

jprupp:


sifag:
Is there a test that is run to ensure the generated code is compatible with the old one? Or can we run such a test? I'm asking since the generated code in `x/clp/types/querier.pb.gw.go` differs. Though I cannot tell if it's relevant or not

timlind (approved):

jprupp:
By the way, I reviewed my assumption that the `latest` tag in Docker Hub points to the latest version tag that pushed: it doesn’t. it just another tag that is set by the developer, and calling it `latest` is just a convention.

Based on the date that the `latest` image was deployed for `proto-gen`:

https://hub.docker.com/layers/sdk-proto-gen/tendermintdev/sdk-proto-gen/latest/images/sha256-4a749a2800dbdfdd1ca4d0394bc4f59c0456f7374257704f074406d6eba3653d?context=explore

It probably doesn't point to the code in version tag `v0.7` as I assume, because that one is much more recent than the last push to `latest`, which happened two hours after tag `v0.6` was pushed.

But the code pushed to `latest` is likely not the same as the code pushed to `v0.6` either, as sizes and hashes are all different.

jprupp:
Digging deeper, I found that our use of `buf protoc` is deprecated by the Buf tool:

https://github.com/bufbuild/buf/releases/tag/v1.0.0

> Remove buf protoc. This was a pre-v1.0 demonstration to show that buf compilation
produces equivalent results to mainline protoc, however buf is working on building
a better Protobuf future that provides easier mechanics than our former protoc-based
world. buf protoc itself added no benefit over mainline protoc beyond being considerably
faster and allowing parallel compilation. If protoc is required, move back to mainline protoc
until you can upgrade to buf. See https://github.com/bufbuild/buf/pull/915 for more
details.

So, we need to move away from this. Already Cosmos SDK has. I'll see about this.

jprupp:
The reason that the generated Protocol Buffers Go code that @timlind uploaded to the repository differs from that which is generated via `make proto-gen` is that he used the underlying directly against his own installed versions of `buf` and `protoco`, and didn’t use the Docker and `make proto-gen`.

PR #2792, opened 2022-05-23
Do not run unit tests in actions for integration tests since we're already running them separately

jzvikart:


jzvikart:
There is no need to run unit tests in actions `integrationtests1.yml` and `integrationtests2.yml`. We're already running unit tests in a separate action, so running them here is a waste of time and resources. Moreover, if the unit tests are failing (like they are now) their failures will mask results of integration tests which otherwise might be passing. This makes failure analysis unnecessarily more complex.

juniuszhou (approved):

jzvikart:
What unit tests did you have in mind? At the moment we have:
1. Golang CI = `go.yml` -> `go test -v ./...` (in /)
2. Smart Contracts = `smart-contracts.yml` -> `make tests` (in smart-contracts)
3. Run integration tests 1 = `integrationtests1.yml` -> `siftool test test/integration/src/peggy2/test_sifnode_transfers.py`
4. Run integration tests 2 for any chain (peg) = `integrationtests2.yml` -> `siftool test test/integration/src/peggy2/test_eth_transfers.py`
5. Run localnet tests on localnet environment = `localnet.yml` -> not sure what these do on peggy2, they were added by @canercandan

(Other workflows are either disabled or not tests.)

PR #2794, opened 2022-05-23
limit pagination on all balances query

timlind:
This PR limits the pagination for the "all balances" query on the bank module.

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -7,6 +7,7 @@ import (
 	"os"
 
 	sifchainAnte "github.com/Sifchain/sifnode/app/ante"
+	sifbank "github.com/Sifchain/sifnode/x/bank"
 	"github.com/Sifchain/sifnode/x/clp"
 	clpkeeper "github.com/Sifchain/sifnode/x/clp/keeper"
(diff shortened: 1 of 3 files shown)


jzvikart:
The changes in `grpc_query.go` seem reasonable to me, but why do we need entirely new `module.go` for this?\

Also, is there a way to achieve consistent result on client side when the balances would change between requests for individual pages? We talked about having an optional parameter "block number", but I don't see it.

What about edge cases - if, say, an account has 100 balances and client gets 100, how does the client diferentiate between the case of "max limit of results reached" and "no more results"?

timlind:
> The changes in `grpc_query.go` seem reasonable to me, but why do we need entirely new `module.go` for this?
>

To wrap the entry points to the querier. 
 
> Also, is there a way to achieve consistent result on client side when the balances would change between requests for individual pages? We talked about having an optional parameter "block number", but I don't see it.
> 

The parameter is called height and is built in to cosmos for all queries so it's not shown here.

> What about edge cases - if, say, an account has 100 balances and client gets 100, how does the client diferentiate between the case of "max limit of results reached" and "no more results"?

Cosmos pagination has a NextKey attribute that specifies the pointer for more results.

PR #2807, opened 2022-05-26
whitelist new erc20 tokens for devnet and testnet

pgoos:


timlind:
@pgoos can you commit the changes to generate-erc20-jsons as well?

timlind (approved):

PR #2819, opened 2022-05-30
Feature/v45 with develop

Jedi2002:
THe PR provides an upgrade handler for the release version 0.14.0


scripts/run.sh
```diff
@@ -2,4 +2,4 @@
 
 killall sifnoded
 
-sifnoded start --trace
+sifnoded start --trace --home ~/.sifnoded/ --p2p.laddr 0.0.0.0:27655  --grpc.address 0.0.0.0:9096 --grpc-web.address 0.0.0.0:9093 --address tcp://0.0.0.0:27659 --rpc.laddr tcp://127.0.0.1:26657
```
Jedi2002, line 5: Added explicit parameters to manage the script better, and use it while running multiple networks while testing 
Default ports are used for the ports used for broadcasting in our test scripts.

cmd/sifnoded/cmd/root.go
```diff
@@ -49,7 +49,6 @@ func NewRootCmd() (*cobra.Command, params.EncodingConfig) {
 		Use:   "sifnoded",
 		Short: "app Daemon (server)",
 		PersistentPreRunE: func(cmd *cobra.Command, _ []string) error {
-			initClientCtx = client.ReadHomeFlag(initClientCtx, cmd)
```
canercandan, line 52: @Jedi2002 any reason why we are getting rid of this one?

Jedi2002: The ReadHomeFlag was removed by the SDK in this version. 
We are already setting the home directory to the default when initializing `initClientCtx` so this should not be a problem

canercandan (approved):
LGTM

PR #2919, opened 2022-06-08
remove todo for issue 2823

juniuszhou:
Currently, all queries are provided from ethbridge module. It is not so necessary to move these queries to oracle module.


smartyalgo (approved):
LGTM. Please merge after tests are passing

PR #2927, opened 2022-06-09
Fix margin interest rate computation

AustinoBombino:


AustinoBombino:
looks like one issue was being caused by unnecessary sdk type conversions in the target rate computation

canercandan (approved):
LGTM

PR #2938, opened 2022-06-13
Fix how we compute the prefix for the block number

banshee:


x/ethbridge/keeper/globalNonce.go
```diff
@@ -75,7 +75,7 @@ func (k Keeper) SetGlobalSequenceToBlockNumber(
 	blockNumber uint64) {
 
 	store := ctx.KVStore(k.storeKey)
-	prefix := k.getGlobalSequenceToBlockNumberPrefix(ctx, networkDescriptor, globalNonce)
+	prefix := k.getGlobalSequenceToBlockNumberPrefix(ctx, networkDescriptor)
```
juniuszhou, line 78: Don't understand why remove the globalNonce from prefix of the key. The keeper is supposed to store the mapping from (network, globalNonce) to block number. Then relayer can know the from block number in sifnode side, to avoid query from very begining. 
Without the globalNonce, it will record the last block number for lock/burn happened. Then relayer will get the from block number large than it should query the events from sifnode, can't get the continued global sequence.

banshee: The issue was that the block number isn't read correctly - I see this:

```
{"level":"info","ts":1655246922.7568674,"caller":"relayer/cosmos.go:174","msg":"ProcessLockBurnWithScope","globalSequence":1,"fromBlockNumber":0,"toBlockNumber":21}
{"level":"info","ts":1655246923.871993,"caller":"relayer/cosmos.go:174","msg":"ProcessLockBurnWithScope","globalSequence":1,"fromBlockNumber":0,"toBlockNumber":21}
{"level":"info","ts":1655246924.9656959,"caller":"relayer/cosmos.go:174","msg":"ProcessLockBurnWithScope","globalSequence":1,"fromBlockNumber":22,"toBlockNumber":22}
{"level":"info","ts":1655246930.759035,"caller":"relayer/cosmos.go:174","msg":"ProcessLockBurnWithScope","globalSequence":2,"fromBlockNumber":0,"toBlockNumber":23}
{"level":"info","ts":1655246931.8613405,"caller":"relayer/cosmos.go:174","msg":"ProcessLockBurnWithScope","globalSequence":2,"fromBlockNumber":0,"toBlockNumber":23}
```

But it's possible I solved it the wrong way.  I'll take another look.

juniuszhou: I may find out what's wrong here. The log is very useful, thanks! The problem is the lock/burn for globalSequence=2 not happened yet, so there is no such block number. the sifnode just return 0. We can add an check, if the globalSequence > 1 and fromBlockNumber = 0, then we can skip it.

banshee:
We're fixing this another way.

PR #2946, opened 2022-06-17
Fix Protocol Automatic Force Close

AustinoBombino:
abci.go was overwriting pool changes made by automatic force close. this was causing the close accounting issues that we were seeing. I moved the set pool to before the mtp close checks and things are working.

this issue was only present for the automated force close, not the normal close and force close transactions.


timlind (approved):
Nice catch

PR #2948, opened 2022-06-20
Feature/liquidity protection cleanup

joneskm:
Remove individual asset buy/sell permissions
Add isActive flag so that the feature can be enabled and disabled
Use base denominations to remove any decimal manipulations


x/clp/types/keys.go
```diff
@@ -35,16 +35,10 @@ var (
 	PmtpParamsPrefix                    = []byte{0x05} // Key to store the Pmtp params
 	RewardParamPrefix                   = []byte{0x06}
 	SymmetryThresholdPrefix             = []byte{0x07}
-	SwapAssetPermissionStorePrefix      = []byte{0x08}
 	LiquidityProtectionParamsPrefix     = []byte{0x09} // Key to store the Liquidity Protection params
 	LiquidityProtectionRateParamsPrefix = []byte{0x0a} // Key to store the Liquidity Protection rate params
```
canercandan, line 39: @joneskm I would suggest to fill the gap between `0x07` and `0x09` by decreasing two new prefixed by one

such as:

```
LiquidityProtectionParamsPrefix     = []byte{0x08} // Key to store the Liquidity Protection params
LiquidityProtectionRateParamsPrefix = []byte{0x09} // Key to store the Liquidity Protection rate params
```

x/clp/types/keys.go
```diff
@@ -81,8 +75,9 @@ func GetDefaultPmtpParams() *PmtpParams {
 
 func GetDefaultLiquidityProtectionParams() *LiquidityProtectionParams {
 	return &LiquidityProtectionParams{
-		MaxRowanLiquidityThreshold:      sdk.MustNewDecFromStr("1000"),
+		MaxRowanLiquidityThreshold:      sdk.NewUint(1000),
```
canercandan, line 78: @joneskm does it mean we can use floating numbers anymore, what if we use BTC as a threshold asset where threshold value is set in BTC and if BTC price very high and you want to rely on decimals for more accuracy, would that means that would not be possible with current implementation?

x/clp/keeper/msg_server.go
```diff
@@ -368,47 +368,37 @@ func (k msgServer) Swap(goCtx context.Context, msg *types.MsgSwap) (*types.MsgSw
...
 		} else {
-			// otherwise retrieve asset pool to retrieve native swap price
-			maxRowanLiquidityThresholdAssetPool, err := k.GetPool(ctx, maxRowanLiquidityThresholdAsset)
+			pool, err := k.GetPool(ctx, maxRowanLiquidityThresholdAsset)
 			if err != nil {
 				return nil, types.ErrMaxRowanLiquidityThresholdAssetPoolDoesNotExist
 			}
-			maxRowanLiquidityThresholdAssetNativePrice = *maxRowanLiquidityThresholdAssetPool.SwapPriceNative
-		}
+			sentValue, err = CalcRowanValue(&pool, pmtpCurrentRunningRate, msg.SentAmount)
```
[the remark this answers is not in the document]
joneskm: The reason for not using the SwapPriceNative is that it doesn't use the base denominations. I also took the opportunity to change to using the spot price  = X/Y instead of XY/(1+X)**2 as used in the SwapPriceNative . It's not clear whether we should calculate the spot price for every pool on every block or simply calculate the spot price when needed. I went for the second option.

x/clp/keeper/msg_server.go
```diff
@@ -528,19 +518,40 @@ func (k msgServer) Swap(goCtx context.Context, msg *types.MsgSwap) (*types.MsgSw
...
+				}
+			}
+
+			// This is equivalent to currentRowanLiquidityThreshold := sdk.MinUint(currentRowanLiquidityThreshold.Add(emitValue), maxRowanLiquidityThreshold)
+			// except it prevents any overflows when adding the emitValue
+			if maxRowanLiquidityThreshold.Sub(currentRowanLiquidityThreshold).LT(emitValue) {
+				currentRowanLiquidityThreshold = maxRowanLiquidityThreshold
+			} else {
+				currentRowanLiquidityThreshold = currentRowanLiquidityThreshold.Add(emitValue)
+			}
```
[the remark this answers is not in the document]
joneskm: same reason as above

x/clp/types/keys.go
joneskm: agreed, will push a change

canercandan (approved):
LGTM, thanks

PR #2953, opened 2022-06-20
feat: 🎸 add sdk 045 feature toggle

canercandan:


app/app.go
```diff
@@ -465,48 +468,139 @@ func NewSifApp(
...
+			distrtypes.ModuleName,
+			slashingtypes.ModuleName,
+			evidencetypes.ModuleName,
+			stakingtypes.ModuleName,
+			ibchost.ModuleName,
+			dispensation.ModuleName,
+			clptypes.ModuleName,
+		)
+	}
+	if featureToggleSdk045 {
```
Jedi2002, line 514: It might a good idea to have only 1  if/else block and set all functions in that

Jedi2002:
Its a good start , but I feel this design will be difficult to scale well

canercandan:
following @Jedi2002 review, I have updated the PR so it uses the go build tags in all the files with changes

Jedi2002:
LGTM , might be a good idea to wait for at least one more review from sifnode

AustinoBombino (approved):

marshsif (approved):

PR #2962, opened 2022-06-23
Feature/margin 1 feature toggle

canercandan:


x/margin/types/types.go
```diff
@@ -1,3 +1,6 @@
+//go:build FEATURE_TOGGLE_MARGIN_CLI_ALPHA
+// +build FEATURE_TOGGLE_MARGIN_CLI_ALPHA
+
```
timlind, line 3: Are these build tags needed in all files or just one's that are imported by the build path? If the dependency path to here is disabled higher up surely these tags aren't needed?

timlind (approved):
Can address linting issues.

AustinoBombino (approved):
nice job! lgtm

PR #2967, opened 2022-06-24
Feature/margin 1 removal queue

timlind:


x/clp/keeper/msg_server.go
```diff
@@ -520,16 +520,37 @@ func (k msgServer) RemoveLiquidity(goCtx context.Context, msg *types.MsgRemoveLi
...
+
 	err = k.Keeper.UseUnlockedLiquidity(ctx, lp, lp.LiquidityProviderUnits.Sub(lpUnitsLeft), false)
 	if err != nil {
 		return nil, err
 	}
 
+	futurePool := pool
+	futurePool.NativeAssetBalance = futurePool.NativeAssetBalance.Sub(withdrawNativeAssetAmount)
+	futurePool.ExternalAssetBalance = futurePool.ExternalAssetBalance.Sub(withdrawExternalAssetAmount)
+	if k.GetMarginKeeper().CalculatePoolHealth(&futurePool).GT(k.GetMarginKeeper().GetRemovalQueueThreshold(ctx)) {
```
canercandan, line 549: @timlind that condition should do the opposite instead no? as a healthy pool starts at a value of `1` and goes down to `0`. and I would assume if the new pool health value is less than the threshold only then we want to queue the request, correct?

x/clp/keeper/msg_server.go
```diff
@@ -652,11 +674,30 @@ func (k msgServer) RemoveLiquidityUnits(goCtx context.Context, msg *types.MsgRem
...
```
canercandan, line 691: @timlind same here I would expect this condition to do the opposite.

x/clp/keeper/msg_server.go
```diff
@@ -652,11 +674,30 @@ func (k msgServer) RemoveLiquidityUnits(goCtx context.Context, msg *types.MsgRem
...
+		k.QueueRemoval(ctx, &types.MsgRemoveLiquidity{
```
canercandan, line 692: @timlind I see this method is used in both `RemoveLiquidity` and `RemoveLiquidityByUnits` although the `msg` type used and that is later passed to emit the event use the same for both, would that be a problem to have accurate event data?

x/clp/keeper/removalqueue.go
```diff
@@ -0,0 +1,129 @@
...
+		k.cdc.MustUnmarshal(it.Value(), &request)
+
+		if msg.ExternalAsset.Equals(*request.Msg.ExternalAsset) {
+			lp, err := k.GetLiquidityProvider(ctx, request.Msg.ExternalAsset.Symbol, request.Msg.Signer)
+			if err != nil {
+				continue
+			}
+
+			requestUnits := ConvWBasisPointsToUnits(lp.LiquidityProviderUnits, request.Msg.WBasisPoints)
+			withdrawUnits := sdk.MinUint(requestUnits, perRequestUnits)
```
canercandan, line 69: @timlind if this is a cap to limit the quantity of units one can withdraw per request according to `perRequestUnits` then we should replace `MinUint` by `MaxUint` so if `requestUnits` is greater than `perRequestUnits` it will use `perRequestUnits` correct?

x/clp/keeper/removalqueue.go
```diff
@@ -0,0 +1,129 @@
...
+func (k Keeper) SetProcessedRemovalRequest(ctx sdk.Context, request types.RemovalRequest, pointsProcessed sdk.Int, rowanRemoved sdk.Uint) {
+	request.Msg.WBasisPoints = request.Msg.WBasisPoints.Sub(pointsProcessed)
+	store := ctx.KVStore(k.storeKey)
+	store.Set(types.GetRemovalRequestKey(request), k.cdc.MustMarshal(&request))
+
+	queue := k.GetRemovalQueue(ctx)
+	queue.TotalValue = queue.TotalValue.Sub(rowanRemoved)
+	k.SetRemovalQueue(ctx, queue)
+
+	if request.Msg.WBasisPoints.LT(sdk.ZeroInt()) {
```
canercandan, line 98: @timlind less than or equal in case the new WBasisPoints value go straight to zero?

canercandan (commented):
@timlind added some comments

x/clp/keeper/removalqueue.go
timlind: No, the aim is to withdraw no more than requestUnits, and no more than perRequestUnits.

x/clp/keeper/msg_server.go
```diff
@@ -520,16 +520,37 @@ func (k msgServer) RemoveLiquidity(goCtx context.Context, msg *types.MsgRemoveLi
 		return nil, types.ErrAsymmetricRemove
 	}
 
+	//  ensure requested removal amount is less than available - what is already on the queue
+	lpQueuedUnits := k.GetRemovalQueueUnitsForLP(ctx, lp)
+	msgUnits := ConvWBasisPointsToUnits(lp.LiquidityProviderUnits, msg.WBasisPoints)
+	if msgUnits.GT(lp.LiquidityProviderUnits.Sub(lpQueuedUnits)) {
```
Jedi2002, line 526: Do you think it makes sense to add a check so that two liquidity removals cannot be sent in the same block?

x/clp/keeper/removalqueue.go
canercandan: got it thanks for clarifying that

Issue #2884, opened 2022-06-07
Unexpected behavior in BridgeBank multiLockBurn


Brando753:
While burn() allows for the burning of tokens with an empty denom, _burnTokens() called by multiLockBurn() reverts with the same input

Brando753:
Confirmed still in the current code base, we should verify these functions behave exactly the same.

Brando753:
BridgeBank should NOT allow burns of empty string; Unit tests should be updated. SetCosmosDenom and like functions should revert if attempted to be set to an empty string.

pandaring2you:
Significant amount of refactoring from javascript to typescript and treating erowan not being a bridgetoken . ETA tomorrow to have Elliot's remediations done by friday

Brando753:
PR #2976 closes this issue out.

Brando753:
This issue was also flagged by Certik audit:

# Description
In the `smart-contracts/contracts/BridgeBank/BridgeBank` contract, the `burn()` function does not check the token's denom while the validation of token denom is processed in the `_burnTokens()` function. The lack of handling token denom check may cause the denom to be an empty string in `LogBurn` event.

As a result, in the `x/ethbridge/keeper/keeper.go`, minting tokens will fail if the denom is an empty string.
```go
  case types.ClaimType_CLAIM_TYPE_BURN:
		coins = sdk.NewCoins(sdk.NewCoin(claim.CosmosDenom, claim.Amount))
		err = k.bankKeeper.MintCoins(ctx, types.ModuleName, coins)
```
# Recommendation
Recommend adding the validation for denom. For example:
```solidity
  require(keccak256(abi.encodePacked(denom)) != keccak256(abi.encodePacked("")), "INV_DENOM");
```

PR #2976, opened 2022-06-25
Make the burn and multiLockBurn methods consistent in how they work.

Brando753:
Closes issues #2884.


Brando753:
This PR now closes out issues #2884 #3002 #3031

smart-contracts/contracts/BridgeBank/BridgeBank.sol
```diff
@@ -67,20 +67,27 @@ contract BridgeBank is BankStorage, CosmosBank, EthereumWhiteList, CosmosWhiteLi
...
+   * @param _rowanTokenAddress The address of the Rowan ERC20 contract on this network
    */
   function initialize(
     address _operator,
     address _cosmosBridgeAddress,
     address _owner,
     address _pauser,
-    int32 _networkDescriptor
+    int32 _networkDescriptor,
+    address _rowanTokenAddress
```
banshee, line 90: Why are we adding a parameter to initialize?  Can't ever be called in production.

Brando753: This contract will be deployed on other networks, and while the address will almost certainly be the zero address on other chains, it should be something that is set on deployment. Since this is an open-source project, it makes sense to me that if someone wanted to deploy these contracts, the initialized code would still be updated with our changes.

banshee (approved):

PR #2979, opened 2022-06-29
Feature/margin 1

canercandan:
PR to merge Margin feature branch to master along with Feature Toggle


marshsif (approved):
I'll approve this, but noting that it is a very large change-set and I've not gone through it item by item.

cmd/sifnoded/Dockerfile
```diff
@@ -29,6 +29,7 @@ RUN go mod download
 
 COPY . .
 
+ENV FEATURE_TOGGLE_MARGIN_CLI_ALPHA=1
```
timlind, line 32: @canercandan this needs to be removed.

canercandan: @timlind good catch Tim, updating it now thanks

proto/sifnode/clp/v1/types.proto
```diff
@@ -25,15 +25,39 @@ message Pool {
     (gogoproto.nullable) = false,
     (gogoproto.moretags) = "yaml:\"pool_units\""
   ];
-  string swap_price_native = 5 [
+  string external_liabilities = 5 [
```
timlind, line 28: See docs for breaking proto changes, I don't think we should be changing field ids.

canercandan: @timlind I've tested using a localnet with existing chain data then updated the field IDs and nothing broke, the structure fields were still using their relevant values. Is it really a concern?

canercandan: my test followed those steps:
1. create a chain with current margin proto structure and field IDs (init and run chain) and create pools
2. stop chain
3. update proto/sifnode/clp/v1/types.proto Pool message struct to use different fields IDs
4. call make proto-all
5. call make install
6. run chain again
7. retrieve existing pools
8. pool field values are the same than what they were before the code changes

```
height: "188"
pool:
  external_asset:
    symbol: cusdc
  external_asset_balance: "488439008293"
  external_custody: "1000000000000000000"
  external_liabilities: "0"
  health: "0.999999989602279511"
  interest_rate: "0.600000000000000000"
  native_asset_balance: "96174925132652547206383901"
  native_custody: "5078652337075426"
  native_liabilities: "1000000000000000000"
  pool_units: "96174925423929435353999282"
  reward_period_native_distributed: "0"
  swap_price_external: "196.902629601115062319"
  swap_price_native: "0.005078652337075426"
```

x/margin/keeper/migrations.go
```diff
@@ -0,0 +1,20 @@
+package keeper
```
timlind, line 1: This file needs build tags, it's breaking the tests @canercandan

canercandan: thanks, added the build tags to the file

AustinoBombino (dismissed):

gzukel (dismissed):
+1

khdegraaf (dismissed):

AustinoBombino (approved):

PR #2991, opened 2022-07-04
feat: custom leverage decimal per mtp

timlind:
Changes leverage to a decimal and allows specifying leverage when opening a position.


AustinoBombino (approved):
nice, looks good

PR #3010, opened 2022-07-11
fix issue 2898, add wrong from account in error

juniuszhou:


x/ethbridge/keeper/blacklist.go
```diff
@@ -19,7 +20,7 @@ func (k Keeper) SetBlacklist(ctx sdk.Context, msg *types.MsgSetBlacklist) error
 	}
 
 	if !k.adminKeeper.IsAdminAccount(ctx, admintypes.AdminType_ETHBRIDGE, from) {
-		return oracletypes.ErrNotAdminAccount
+		return fmt.Errorf("%s %s", from.String(), "not an admin account")
```
smartyalgo, line 23: I think you should use cosmos sdk's error, and use wrap to provide additional info
See https://github.com/Sifchain/sifnode/pull/3014/files#diff-ee097124641d93e6a1eada444f07ec8571326ca882dca0fc5d95a81de4e4055cR38

juniuszhou: will use this method, thanks

smartyalgo (approved):

PR #3011, opened 2022-07-12
fix issue 2905, correct error code for metadata check

juniuszhou:


x/tokenregistry/types/msgs.go
```diff
@@ -274,11 +274,11 @@ func (msg TokenMetadataAddRequest) ValidateBasic() error {
 	}
```
smartyalgo, line 274: Should we also validate NetworkDescriptor here?

juniuszhou: will check the NetworkDescriptor

smartyalgo (approved):

Issue #2901, opened 2022-06-08
Denom validation by cosmos SDK methods


Brando753:
File Affected:  x/clp/types/asset.go: 23-34. Because the Amount field of sdk.Coin type is always sdk.OneInt(), building a coin type just to validate just the denom is redundant. Use sdk.ValidateDenom() and return a boolean based on any thrown error instead of using coin.Validate().

Brando753:
This same ticket also affects the file: x/tokenregistry/types.msgs.go: 31-40, 80-89, 127-133.

juniuszhou:
Just find out the MaxSymbolLength used in the clp module is different with cosmos sdk. So I suggest we won't fix it.

pandaring2you:
Need 2 more confirmations from either @smartyalgo @Brando753 @banshee

pandaring2you:
When we get 2 confirmations, will close the PR without merging.

pandaring2you:
Note: @juniuszhou Should reach out to sifnode team why MaxSymbolLength is different than CosmosSDK

banshee:
I'm wondering why we trim whitespace:

```
func (a Asset) Validate() bool {
	if !VerifyRange(len(strings.TrimSpace(a.Symbol)), 0, MaxSymbolLength) {
		return false
	}
	coin := sdk.NewCoin(a.Symbol, sdk.OneInt())
	return coin.IsValid()
}
```

This seems like a historical misfeature.  The original code returned false if the symbol just contained whitespace.  The new code checks the length of the symbol after removing whitespace, so a symbol thats "<1000 spaces>SYM<1000 spaces>" is valid.  Seems weird.

banshee:
Believe @Brando753 meant x/tokenregistry/types/msgs.go, where we construct coins just to validate denom.

banshee:
Also see https://github.com/Sifchain/sifnode/pull/3085

juniuszhou:
wait the input from sifnode team if MaxSymbolLength=71 is reasonable and should we remove it.

juniuszhou:
@timlind and @AustinoBombino agree on extend it. #3028 ready for code review.

banshee:
Merged #3028

juniuszhou:
close it after PR #3028 merged.

PR #3028, opened 2022-07-14
use ValidateDenom to check denom

juniuszhou:


juniuszhou:
will close it without merge if we agree on won't fix for #2901

banshee (approved):

PR #3034, opened 2022-07-19
reserve error code

juniuszhou:
The error code 1 should be reserved for cosmos system.

### DIFF
--- a/x/clp/types/errors.go
+++ b/x/clp/types/errors.go
@@ -5,11 +5,11 @@ import (
 )
 
 var (
-	ErrInvalid                                         = sdkerrors.Register(ModuleName, 1, "invalid")
-	ErrPoolDoesNotExist                                = sdkerrors.Register(ModuleName, 2, "pool does not exist")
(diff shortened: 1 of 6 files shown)


Brando753:
@juniuszhou it looks reasonable to me but we really need signoff from Sifnode and data services

juniuszhou:
> @juniuszhou it looks reasonable to me but we really need signoff from Sifnode and data services
split the commit, just keep peggy related x modules.

smartyalgo (approved):

juniuszhou:
@Jedi2002 @canercandan @timlind Can you approve and merge this into master?

Jedi2002 (commented):
LGTM . 
Curious though, as to why we should reserve error code 1 ?

juniuszhou:
> LGTM . Curious though, as to why we should reserve error code 1 ?

The issue is from audit report, error code 1 is reserved for cosmos sdk. 
```
	// errInternal should never be exposed, but we reserve this code for non-specified errors
	errInternal = Register(UndefinedCodespace, 1, "internal")
```

banshee (approved):

PR #3047, opened 2022-07-20
Fix/liqp rate update event

canercandan:
We are using the wrong event type for LIQP rates params update flow, created a new one for this scenario and using it in-place of the previous one


joneskm (approved):
lgtm

PR #3062, opened 2022-07-25
add data in genesis

juniuszhou:


banshee:
Can we switch

```
func (k Keeper) getKeyViaNetworkDescriptorGlobalNonce(networkDescriptor types.NetworkDescriptor,
	globalSequence uint64) []byte {
	bs1 := make([]byte, 4)
	binary.LittleEndian.PutUint32(bs1, uint32(networkDescriptor))

	bs2 := make([]byte, 8)
	binary.LittleEndian.PutUint64(bs2, globalSequence)

	storeKey := append(append(types.GlobalNonceProphecyIDPrefix, bs1[:]...), bs2[:]...)
	return storeKey
}

```

to also use bigendian?  If we use bigendian here, there's no littleendian in our code at all.

juniuszhou:
> Can we switch
> 
> ```
> func (k Keeper) getKeyViaNetworkDescriptorGlobalNonce(networkDescriptor types.NetworkDescriptor,
> 	globalSequence uint64) []byte {
> 	bs1 := make([]byte, 4)
> 	binary.LittleEndian.PutUint32(bs1, uint32(networkDescriptor))
> 
> 	bs2 := make([]byte, 8)
> 	binary.LittleEndian.PutUint64(bs2, globalSequence)
> 
> 	storeKey := append(append(types.GlobalNonceProphecyIDPrefix, bs1[:]...), bs2[:]...)
> 	return storeKey
> }
> ```
> 
> to also use bigendian? If we use bigendian here, there's no littleendian in our code at all.

yes, it is better to use bigendian. will update it.

smartyalgo:
Curious, why is bigEndian better?

juniuszhou:
> Curious, why is bigEndian better?

just use the same endian in the code, I have other PR #2748 to use the protobuf. I think it is better solution, can avoid some mistake for different endian usage.

x/ethbridge/keeper/ethereumLockBurnSequence.go
```diff
@@ -36,10 +36,42 @@ func (k Keeper) GetEthereumLockBurnSequence(ctx sdk.Context, networkDescriptor o
...
+	for ; iterator.Valid(); iterator.Next() {
+		key := iterator.Key()
+		value := iterator.Value()
+		sequences[string(key)] = binary.BigEndian.Uint64(value)
+	}
+	return sequences
+}
+
+// SetSequenceViaRawKey used in import sequence from genesis
+func (k Keeper) SetSequenceViaRawKey(ctx sdk.Context, key []byte, newSequence uint64) {
```
smartyalgo, line 71: This looks very error prone. 
I suggest write a function to extract NetworkDescriptor and ValidatorAddress, and use those to call `SetEthereumLockBurnSequence`

proto/sifnode/ethbridge/v1/types.proto
```diff
@@ -62,4 +62,9 @@ enum ClaimType {
 }
 
 // GenesisState for ethbridge
-message GenesisState { string crosschain_fee_receive_account = 1; }
+message GenesisState {
```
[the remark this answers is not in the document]
juniuszhou: CrosschainFeeReceiveAccount already checked, for other three items, they are map, so that they are empty is correct if no any data before in peggy1.0. Do you mean we should check each entry in these maps?

proto/sifnode/ethbridge/v1/types.proto
```diff
@@ -62,4 +62,9 @@ enum ClaimType {
 }
 
 // GenesisState for ethbridge
-message GenesisState { string crosschain_fee_receive_account = 1; }
+message GenesisState { 
+    string crosschain_fee_receive_account = 1; 
+    map<string, uint64> ethereum_lock_burn_sequence = 2;
```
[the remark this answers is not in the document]
juniuszhou: 1. if ethereum_lock_burn_sequence not set before, then no data will be included in genesis, it doesn't generate any empty/default value
2. 0 is expected if no vallue exists, to avoid such confusion, we set the lockBurnSequence starting from 1, also in contract

x/ethbridge/keeper/ethereumLockBurnSequence.go
juniuszhou: Ok, will add extract method.

banshee: Are these comments still current?

juniuszhou: I implemented a new method DecodeKey to extract NetworkDescriptor and ValidatorAddress according to comment. please check.

juniuszhou:
need deal with map issue first.

PR #3075, opened 2022-07-28
Feature/sqmodifier

sifag:


timlind (approved):

x/margin/types/genesis.go
```diff
@@ -20,6 +20,7 @@ func DefaultGenesis() *GenesisState {
 			EpochLength:           1,
 			MaxOpenPositions:      10000,
 			Pools:                 []string{},
+			SqModifier:            sdk.MustNewDecFromStr("10000000000000000000000000"),
```
canercandan, line 23: @sifag do we have a spec somewhere about this default value? thanks

sifag: If you look at the diff, it was the default value that was set

canercandan (approved):

x/margin/types/genesis.go
timlind: @caner this default value still needs to be decided.

PR #3100, opened 2022-08-03
Include Margin Debt in CLP Swap

AustinoBombino:


timlind (approved):

x/clp/keeper/calculations.go
```diff
@@ -16,53 +17,27 @@ func SwapOne(from types.Asset,
 	sentAmount sdk.Uint,
 	to types.Asset,
 	pool types.Pool,
-	normalizationFactor sdk.Dec,
-	adjustExternalToken bool,
-	pmtpCurrentRunningRate sdk.Dec) (sdk.Uint, sdk.Uint, sdk.Uint, types.Pool, error) {
+	pmtpCurrentRunningRate sdk.Dec,
+	marginEnabled bool) (sdk.Uint, sdk.Uint, sdk.Uint, types.Pool, error) {
```
canercandan, line 21: @AustinoBombino this need to live within the `calculations_FEATURE_TOGGLE_MARGIN_CLI_ALPHA.go` file instead as you are changing the signature of the function that if remains in this file will be available for production code, so instead as you want to make it available only when the feature toggle Margin is enabled you could one of those two options:

Option A:
1) move the entire `SwapOne` function with the margin changes to `calculations_FEATURE_TOGGLE_MARGIN_CLI_ALPHA.go`
2) create a new `calculations_NO_FEATURE_TOGGLE_MARGIN_CLI_ALPHA.go` file that will contain the header:
```
//go:build !FEATURE_TOGGLE_MARGIN_CLI_ALPHA
// +build !FEATURE_TOGGLE_MARGIN_CLI_ALPHA
```
and the original `SwapOne` function

Option B:
1) You follow the steps in Option A
2) You create an inner function that will live within `calculations.go` that will contain the code logic that are common between those two versions of the SwapOne function so there are less code to maintain in cases changes occur in this part of the code.

x/clp/keeper/calculations.go
```diff
@@ -78,37 +53,19 @@ func CalcSwapPmtp(toRowan bool, y, pmtpCurrentRunningRate sdk.Dec) sdk.Dec {
...
-	normalizationFactor sdk.Dec,
-	adjustExternalToken bool,
-	pmtpCurrentRunningRate sdk.Dec) sdk.Uint {
-	X, x, Y, toRowan := SetInputs(sentAmount, to, pool)
-	swapResult, err := CalcSwapResult(toRowan, normalizationFactor, adjustExternalToken, X, x, Y, pmtpCurrentRunningRate)
-	if err != nil {
-		// this branch will never be reached as err will always be nil
-		return sdk.Uint{}
+	pmtpCurrentRunningRate sdk.Dec,
+	marginEnabled bool) sdk.Uint {
```
canercandan, line 60: @AustinoBombino similarly here we should move `GetSwapFee` to the margin feature toggle file as the signature of the function changes and preserve the previous one in none feature toggle file, you can see more details on how to implement that in my previous comment.

x/clp/keeper/msg_server.go
```diff
@@ -474,8 +473,8 @@ func (k msgServer) Swap(goCtx context.Context, msg *types.MsgSwap) (*types.MsgSw
 	// Check if its a two way swap, swapping non native fro non native .
 	// If its one way we can skip this if condition and add balance to users account from outpool
 	if !msg.SentAsset.Equals(nativeAsset) && !msg.ReceivedAsset.Equals(nativeAsset) {
-		normalizationFactor, adjustExternalToken := k.GetNormalizationFactor(decimals)
-		emitAmount, lp, ts, finalPool, err := SwapOne(*sentAsset, sentAmount, nativeAsset, inPool, normalizationFactor, adjustExternalToken, pmtpCurrentRunningRate)
+		marginEnabled := k.getMarginKeeper().IsPoolEnabled(ctx, inPool.String())
+		emitAmount, lp, ts, finalPool, err := SwapOne(*sentAsset, sentAmount, nativeAsset, inPool, pmtpCurrentRunningRate, marginEnabled)
```
canercandan, line 477: @AustinoBombino same here for msg_server functions that do not use the same function's signature

canercandan (commented):
@AustinoBombino please review my comments, thanks

canercandan (approved):
LGTM

PR #3111, opened 2022-08-08
fix: use safety factor in place of force close

timlind:


x/margin/client/cli/tx.go
```diff
@@ -260,7 +260,7 @@ func GetUpdateParamsCmd() *cobra.Command {
 	_ = cmd.MarkFlagRequired("interest-rate-increase")
 	_ = cmd.MarkFlagRequired("interest-rate-decrease")
 	_ = cmd.MarkFlagRequired("health-gain-factor")
-	_ = cmd.MarkFlagRequired("force-close-threshold")
+	//_ = cmd.MarkFlagRequired("force-close-threshold")
```
canercandan, line 263: @timlind can we remove that altogether?

x/margin/keeper/msg_server.go
```diff
@@ -282,7 +286,7 @@ func (k Keeper) ForceCloseLong(ctx sdk.Context, msg *types.MsgForceClose) (*type
 	}
 
 	// check MTP health against threshold
-	forceCloseThreshold := k.GetForceCloseThreshold(ctx)
+	forceCloseThreshold := k.GetSafetyFactor(ctx)
```
canercandan, line 289: @timlind I don't we needed to do additional calculation along with SF to be able to use it as a force close threshold am I missing something?

canercandan: @timlind you can ignore that comment as @Rubz2021 confirmed SF should be used instead in-place of previous force close threshold, thanks

PR #3113, opened 2022-08-08
Report block time via CLP module

sifag:


joneskm (approved):

joneskm (approved):

x/clp/abci.go
```diff
@@ -136,3 +137,17 @@ func BeginBlocker(ctx sdk.Context, k kpr.Keeper) {
...
+func MeasureBlockTime(ctx sdk.Context, k kpr.Keeper) {
+	now := time.Now()
+	if blockTime == nil {
+		blockTime = &now
+		return
+	}
+
+	elapsed := now.Sub(*blockTime)
+	blockTime = &now
+	k.Logger(ctx).Info(fmt.Sprint("Block took ", elapsed.Seconds(), "s to execute"))
```
timlind, line 152: Might want to put the block number in this log too.

timlind (approved):

x/clp/abci.go
sifag: Block height is printed in the very next line. I think it's fine to not include it here
```
12:29PM INF Block took 5.014953372s to execute module=x/clp
12:29PM INF executed block height=4 module=state num_invalid_txs=0 num_valid_txs=0
```

PR #3123, opened 2022-08-11
add new argument maxMessagesInSifnodeTransaction 

juniuszhou:
In load testing, the TPS from Ethereum to Sifnode just 1. add new argument maxMessagesInSifnodeTransaction in ebrelayer, then relayer can put more messages in a single transaction.

### DIFF
--- a/cmd/ebrelayer/main.go
+++ b/cmd/ebrelayer/main.go
@@ -33,16 +33,18 @@ import (
 )
 
 const (
-	networkDescriptorFlag             = "network-descriptor"
-	tendermintNodeFlag                = "tendermint-node"
-	web3ProviderFlag                  = "web3-provider"
-	bridgeRegistryContractAddressFlag = "bridge-registry-contract-address"
-	validatorMonikerFlag              = "validator-moniker"
-	maxFeePerGasFlag                  = "maxFeePerGasFlag"
-	maxPriorityFeePerGasFlag          = "maxPriorityFeePerGasFlag"
-	ethereumChainIdFlag               = "ethereum-chain-id"
-	sifnodeGrpcEntryPointFlag         = "sifnode-grpc-endpoint"
-	defaultSifnodeGrpc                = "0.0.0.0:9090"
+	networkDescriptorFlag                  = "network-descriptor"
+	tendermintNodeFlag                     = "tendermint-node"
+	web3ProviderFlag                       = "web3-provider"
+	bridgeRegistryContractAddressFlag      = "bridge-registry-contract-address"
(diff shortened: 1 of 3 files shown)


jzvikart:
I feel like this parameter should be set automatically and not by user. Because honestly we're not solving the problem here, we're just shifting it to some poor guy who will have no idea or interest what it's all about. For example, what would life look like if we had to manually set window size for every TCP connection? Or if we had to manually adjust the fuel/air mixture ratio in the car engine? I'm not saying that we should implement a full-blown solution like New Reno right here and now, but at least it would be a good idea to think about it from architectural point of view and have some concept about a possible solution.

So the question is - is there a way to determine the optimal value of this parameter automatically and dynamically during run-time? Some kind of algorithm that adapts to the moment-in-time situation, or perhaps just a fixed number that takes into account the max message size/worst case scenario? Since this challenge is not specific to peggy/cosmos, how do other software such as IBC relayers deal with it? etc. Having a single "fast and reliable send" function would probably benefit us in many ways...

Also, what about the Sifchain -> Ethereum flow? How are we dealing with it there? Is it subject to the same limitations?

juniuszhou:
> I feel like this parameter should be set automatically and not by user. Because honestly we're not solving the problem here, we're just shifting it to some poor guy who will have no idea or interest what it's all about. For example, what would life look like if we had to manually set window size for every TCP connection? Or if we had to manually adjust the fuel/air mixture ratio in the car engine? I'm not saying that we should implement a full-blown solution like New Reno right here and now, but at least it would be a good idea to think about it from architectural point of view and have some concept about a possible solution.
> 
> So the question is - is there a way to determine the optimal value of this parameter automatically and dynamically during run-time? Some kind of algorithm that adapts to the moment-in-time situation, or perhaps just a fixed number that takes into account the max message size/worst case scenario? Since this challenge is not specific to peggy/cosmos, how do other software such as IBC relayers deal with it? etc. Having a single "fast and reliable send" function would probably benefit us in many ways...
> 
> Also, what about the Sifchain -> Ethereum flow? How are we dealing with it there? Is it subject to the same kind of limitations?

I think the current value 5 is enough for our current traffic and I aslo set it as default value, the relayer can start without set it explicitly. It is just maximum value for messages included in a transaction, so if traffic is low, no need for auto ajustment during run-time. The argument may just be used in load testing or relayer down for a long time, need to catch up after restart. I checked the Gravity bridge, no such parameter.
From Sifchain -> Ethereum, we also have the similar argument. 
// send the prophecy by batch, maximum is 5 prophecies in each batch

jzvikart (approved):

PR #3130, opened 2022-08-12
Kevin/fix lint2

khdegraaf:
Remove deprecated linters.  Don't want to merge it into master until after 0.15, due to risk of changing error strings downstream.


canercandan (approved):
LGTM

PR #3135, opened 2022-08-15
Test/fix go version

canercandan:


sheokapr (approved):
Straightforward Change there to upgrade golang version.

PR #3179, opened 2022-08-27
register MsgRegisterAll and MsgDeregisterAll

juniuszhou:


smartyalgo (approved):

smartyalgo:
Do you know what's the use case for DeregisteringAll ?

juniuszhou:
> DeregisteringAll

It is related to business and operation, not sure about it. The condition could be like for CEX, sometimes we need to remove some inactive exchange pair.

smartyalgo (approved):

PR #3184, opened 2022-08-30
Add admin close and tests

Jedi2002:


x/margin/keeper/msg_server.go
```diff
@@ -328,6 +326,31 @@ func (k Keeper) ForceCloseLong(ctx sdk.Context, msg *types.MsgForceClose, isAdmi
...
+	signer, err := sdk.AccAddressFromBech32(msg.Signer)
+	if err != nil {
+		return nil, err
+	}
+	if !k.AdminKeeper().IsAdminAccount(ctx, admintypes.AdminType_MARGIN, signer) {
+		return nil, sdkerrors.Wrap(admintypes.ErrPermissionDenied, fmt.Sprintf("signer not authorised: %s", msg.Signer))
+	}
+
+	params := k.GetParams(ctx)
+	if msg.CloseAll {
```
joneskm, line 340: Do we need this boolean? If it's set to false then does the AdminClose just change the ForceCloseFundPercentage?

x/margin/keeper/msg_server.go
```diff
@@ -328,6 +326,31 @@ func (k Keeper) ForceCloseLong(ctx sdk.Context, msg *types.MsgForceClose, isAdmi
...
+		params.SafetyFactor = sdk.MustNewDecFromStr("100.00")
```
joneskm, line 341: Might be cleaner to use sdk.NewDec(100)

x/margin/keeper/msg_server.go
Jedi2002: Yeah , and then we can just use the existing force close to close individual MTP's
But I think this can be discussed. I do feel if Repay is not needed at all  , for any ForceClose , I can just change the ForceClose function to update this parameter as well , and that would make things much cleaner .

x/margin/keeper/msg_server.go
```diff
@@ -328,6 +326,31 @@ func (k Keeper) ForceCloseLong(ctx sdk.Context, msg *types.MsgForceClose, isAdmi
...
+	}
+	params.ForceCloseFundPercentage = sdk.ZeroDec()
```
AustinoBombino, line 343: @Jedi2002 
I don't think we want to set the system-wide forceclosefund percentage for closing an individual mtp. May make more sense to change the call to Repay() in forcecloselong to pass a a variable for takeInsurance instead of always true

AustinoBombino (approved):

PR #3188, opened 2022-08-31
Min Borrow Amount

AustinoBombino:
Adds check in mtp open comparing liabilities with current minimum interest rate. 

If the liabilities are too low to compute an actual interest payment an error is returned: `(ModuleName, 9, "borrowed amount is too low")`

In cases where InterestRateMin has been lowered with outstanding positions open, payments may resolve to less than 1. The interest computation will now round up to the smallest possible unit.

When changing InterestRateMin we run the risk of overcharging via rounding mentioned above, this can be avoided by closing all outstanding positions prior to lowering InterestRateMin.


AustinoBombino:
fixing tests

canercandan (dismissed):
LGTM

canercandan (approved):
LGTM

PR #3189, opened 2022-08-31
Bugfix/3152/mismatch consensus threshold

smartyalgo:
In this PR:

No longer return error when a witness adds claim to a SUCCESS prophecy. This error is ignored up the chain, except in SignProphecy, where it will cause the msg_server to treat it as an error and not update the witness' global sequence

This isnt strictly related to the ticket, but I found this while working on 3152

To reproduce, set witness count such that at least one witness will validate a prophecy that has passed consensusThreshold
e.g. Threshold = 66%, set 4 witness/relayers


juniuszhou:
It is a big issue. I am wondering if we should remove the error type ErrProphecyFinalized, instead we add a new status like already_success. Then we also can avoid the relayer to retry the same transaction.

x/oracle/keeper/keeper.go
```diff
@@ -119,6 +122,7 @@ func (k Keeper) AppendValidatorToProphecy(ctx sdk.Context, networkDescriptor typ
 		return prophecy.Status, types.ErrProphecyFinalized
 
 	default:
+		// This is possible IFF GetProphecy returned with UNSPECIFIED
```
smartyalgo, line 125: I think we should ensure Prophecy Stored in Keeper never has UNSPECIFIED. UNSPECIFIED is 0, the default in golang if no value is passed when constructing Prophecy object

smartyalgo:
That's a good point. For instance, in Sifchain to EVM flow, we check [old and new status](https://github.com/Sifchain/sifnode/blob/future/peggy2/x/oracle/keeper/keeper.go#L254) so we only emit the event once.

x/oracle/keeper/keeper.go
juniuszhou: I agree, the UNSPECIFIED as returned value is not good, we can add a new status enum. but it is not a big issue, the second value is an error, so it means UNSPECIFIED should not be used by caller.

juniuszhou (approved):

PR #3194, opened 2022-09-02
Smart contract function to update consensusThreshold

smartyalgo:
In this PR:

Add privileged function to update CosmosBridge.sol::consensusThreshold to be in sync with value stored in sifnoded

### DIFF
--- a/smart-contracts/.prettierrc
+++ b/smart-contracts/.prettierrc
@@ -27,7 +27,8 @@
         "tabWidth": 2,
         "useTabs": false,
         "singleQuote": false,
-        "semi": false
+        "semi": true,
+        "trailingComma": es5
       }
     }
   ]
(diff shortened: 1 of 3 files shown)


juniuszhou (approved):

smart-contracts/.prettierrc
```diff
@@ -27,7 +27,8 @@
         "tabWidth": 2,
         "useTabs": false,
         "singleQuote": false,
-        "semi": false
+        "semi": true,
```
banshee, line 30: This should stay false, and fixing this will remove a bunch of other changes in this PR

banshee (changes requested):

smart-contracts/.prettierrc
smartyalgo: If false it will remove all semicolons, I think we want semicolons in ts?

banshee: Let's do this in another PR.

PR #3216, opened 2022-09-07
MTP Open Safety Factor Check

AustinoBombino:


x/margin/keeper/msg_server.go
```diff
@@ -211,6 +211,18 @@ func (k msgServer) OpenLong(ctx sdk.Context, msg *types.MsgOpen) (*types.MTP, er
...
+	lr, err := k.UpdateMTPHealth(ctx, *mtp, pool)
+
+	if err != nil {
+		return nil, err
+	}
+
+	if lr.LTE(safetyFactor) {
+		return nil, types.ErrMTPUnhealthy
+	}
+
```
timlind, line 225: Do we want to be checking mtp health here or pool health after opening? There is a pool open threshold we could maybe check against.

AustinoBombino: @timlind we had a call yesterday with a bunch of us and Barlin at blockscience and the decision was to base it on the posterior mtp health

canercandan:
Test case:

Opening a position with the entire pool depth to make sure it drops LR under safety factor to trigger the error:

```
- external_asset:
    symbol: cusdc
  external_asset_balance: "611316557165"
  external_custody: "0"
  external_liabilities: "9000000000"
  health: "0.985491278773643856"
  interest_rate: "0.000000010999999999"
  last_height_interest_rate_computed: "15"
  native_asset_balance: "79978669084012644657289180"
  native_custody: "1306450091106295598863316"
  native_liabilities: "0"
  pool_units: "81285119183944010790286697"
  reward_period_native_distributed: "0"
  swap_price_external: "131.038125995881888844"
  swap_price_native: "0.007631366767496559"
```

Trying to open long position against USDC with a big position size (all the liquidity available in the pool):

```
sifnoded tx margin open   --borrow_asset=rowan   --collateral_asset=cusdc   --collateral_amount=61131655716   --position=long   --leverage=10   $'--keyring-backend=test' $'--node=http://localhost:26657' $'--chain-id=localnet' $'--from=account-sif1syavy2npfyt9tcncdtsdzf7kny9lh777yqc2nd' -y   $'--fees=100000000000000000rowan'   $'--broadcast-mode=block'   --account-number=0   --sequence=8   
```

that throws the expected error:

```
raw_log: 'failed to execute message; message index: 0: mtp health would be too low
  for safety factor'

```

canercandan (approved):
LGTM

PR #3219, opened 2022-09-08
add validate for blacklist

juniuszhou:


x/ethbridge/types/msgs.go
```diff
@@ -671,6 +672,20 @@ func (msg MsgSetBlacklist) GetSignBytes() []byte {
 }
 
 func (msg *MsgSetBlacklist) ValidateBasic() error {
+	if msg.From == "" {
```
smartyalgo, line 675: Should we also check if it is valid bech32 address?
`sdk.AccAddressFromBech32(msg.From)`

x/ethbridge/types/msgs.go
```diff
@@ -671,6 +672,20 @@ func (msg MsgSetBlacklist) GetSignBytes() []byte {
...
+		return sdkerrors.Wrap(sdkerrors.ErrInvalidAddress, msg.From)
+	}
+
+	if len(msg.Addresses) == 0 {
+		return ErrEmptyBlackList
```
smartyalgo, line 680: Why isnt an empty blacklist allowed? For example, when we init, or when we want to clear the list

x/ethbridge/types/msgs.go
juniuszhou: will add it.

x/ethbridge/types/msgs.go
juniuszhou: It is my wrong assumption, I thought it is an incremental function. will remove this check.

smartyalgo: Can you add a test such to verify empty list wont throw err?

juniuszhou: sure, will add it.

PR #3221, opened 2022-09-08
remove ErrcrossChainFee

juniuszhou:


smartyalgo:
Let's delete it instead of commenting it out

juniuszhou:
> Let's delete it instead of commenting it out
forget to remove it, will do.

PR #3273, opened 2022-09-20
pool balance vs module account invariant check

timlind:
WIP: This isn't working yet. On creating the first pool without enough funds to do so, the module account seems to be high while pool balance is zero.

### DIFF
--- a/x/clp/abci.go
+++ b/x/clp/abci.go
@@ -40,6 +40,11 @@ func EndBlocker(ctx sdk.Context, keeper kpr.Keeper) []abci.ValidatorUpdate {
 		}
 	}
 
+	res, stop := kpr.BalanceModuleAccountCheck(keeper)(ctx)
+	if stop {
+		panic(res)
+	}
+
 	return []abci.ValidatorUpdate{}
 }
 
--- a/x/clp/keeper/invariants.go
+++ b/x/clp/keeper/invariants.go
@@ -1,0 +1,44 @@
+package keeper
+
+import (
+	"fmt"
+
+	"github.com/Sifchain/sifnode/x/clp/types"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+)
+
+func RegisterInvariants(registry sdk.InvariantRegistry, k Keeper) {
+	// registry.RegisterRoute(types.ModuleName, "balance-module-account-check", BalanceModuleAccountCheck(k))
+}
+
+func BalanceModuleAccountCheck(k Keeper) sdk.Invariant {
+	return func(ctx sdk.Context) (string, bool) {
+		// Get Rowan Balance from CLP Module
+		clpModuleTotalNativeBalance := k.GetBankKeeper().GetBalance(ctx, types.GetCLPModuleAddress(), types.GetSettlementAsset().Symbol)
+		clpModuleTotalNativeBalanceUint := sdk.NewUintFromString(clpModuleTotalNativeBalance.Amount.String())
+
+		pools := k.GetPools(ctx)
+		poolsTotalNativeBalanceUint := sdk.ZeroUint()
+		for _, pool := range pools {
+			clpModuleTotalExternalBalance := k.GetBankKeeper().GetBalance(ctx, types.GetCLPModuleAddress(), pool.ExternalAsset.Symbol)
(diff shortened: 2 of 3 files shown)


canercandan:
# Test of the CLP module wallet invariant checkers

tested against v1.0-beta.11 (betanet version) with the swap fix:

## :red_circle: Swap Rowan to CUSDC while margin position on
* no monetary policy on
* cusdc pool enabled for margin
* open cusdc position
* swap rowan to cusdc
* throw an invariant error:
```
external balance mismatch in pool cusdc (module: 603861131258 != pool: 607861131258)
```

## :red_circle: Swap CUSDC to Rowan while margin position on
* no monetary policy on
* cusdc pool enabled for margin
* open cusdc position
* swap cusdc to rowan
* throw an invariant error:
```
external balance mismatch in pool cusdc (module: 612316557165 != pool: 616316557165)"
```

## :heavy_check_mark: Add liquidity to CUSDC while margin position on
* no monetary policy on
* cusdc pool enabled for margin
* open cusdc position
* add liquidity to USDC pool
* does not throw any invariant error

## :heavy_check_mark: Remove liquidity from CUSDC while margin position on
* no monetary policy on
* cusdc pool enabled for margin
* open cusdc position
* remove liquidity from USDC pool
* does not throw any invariant error

## :heavy_check_mark: LPD from CUSDC while margin position on
* no Rewards or LIQP on
* cusdc pool enabled for margin
* open cusdc position
* enable LPD distribution to distribute at each block
* does not throw any invariant error

## :heavy_check_mark: Rewards to CUSDC while margin position on
* no LPD or LIQP on
* cusdc pool enabled for margin
* open cusdc position
* enable Rewards to CUSDC pool
* does not throw any invariant error

## :heavy_check_mark: Rewards + LPD on while margin position on
* no LIQP on
* cusdc pool enabled for margin
* open cusdc position
* enable LPD distribution to distribute at each block
* enable Rewards to CUSDC pool
* does not throw any invariant error

canercandan (dismissed):

canercandan (approved):

AustinoBombino (approved):
lgtm

Jedi2002:
I have some thoughts 
- It seems like a lot of repetitive code, which does the same thing , we can thinking about using a  post handler 
https://docs.cosmos.network/v0.46/core/baseapp.html#posthandler

- Also, I do feel the end block check is enough to handle the situation, It is not very clear to me how returning from messages would protect us, because we have a lot of state changes that are happening in the begin blocker/end blocker, such as monetary policies and rewards . 

- It is also difficult to see how these tx level checks might get affected if there are two transactions in the same block. This might not be a big issue, but we should test with at least doing multiple add liquidity and swaps on the same bool in the same block.

- It might actually be better to just keep the invariant check with the logger for now and deploy it, and closely monitor the system . If errors do not get logged for a long duration, We can just convert the logging to a panic. If we do see errors, we can decide what to do. Maybe adding more logging such as  pools states or active mtps etc might be usefull

AustinoBombino:
> I have some thoughts
> 
> * It seems like a lot of repetitive code, which does the same thing , we can thinking about using a  post handler
>   https://docs.cosmos.network/v0.46/core/baseapp.html#posthandler
> * Also, I do feel the end block check is enough to handle the situation, It is not very clear to me how returning from messages would protect us, because we have a lot of state changes that are happening in the begin blocker/end blocker, such as monetary policies and rewards .
> * It is also difficult to see how these tx level checks might get affected if there are two transactions in the same block. This might not be a big issue, but we should test with at least doing multiple add liquidity and swaps on the same bool in the same block.
> * It might actually be better to just keep the invariant check with the logger for now and deploy it, and closely monitor the system . If errors do not get logged for a long duration, We can just convert the logging to a panic. If we do see errors, we can decide what to do. Maybe adding more logging such as  pools states or active mtps etc might be usefull

I agree the post handler looks like a good way to condense this code.

timlind:
Post handler seems like a good option for this one, but for future checks that may be more heavy weight we might want to hand select where it's run.

The benefit of doing the check at message level instead of just end block is that we can prevent bad messages from being committed without halting the chain in end blocker, but agreed that end blocker is needed as well to check the other block related code.

timlind:
Looks like PostHandler is only introduced in sdk v0.46

PR #3289, opened 2022-09-28
add address check

juniuszhou:


smartyalgo (approved):
LGTM, can you add unit test to verify this behavior?

unit test added.

PR #3293, opened 2022-10-01
fix: 🐛 beta12 invariant check + rowan as collateral asset

canercandan:


AustinoBombino (approved):
current live hotfix build

PR #3302, opened 2022-10-05
feat: add ledger support on cmd line

timlind:


Makefile
```diff
@@ -11,7 +11,7 @@ DOCKER ?= docker
 DOCKER_BUF := $(DOCKER) run --rm -v $(CURDIR):/workspace --workdir /workspace bufbuild/buf
 
 GOFLAGS:=""
-GOTAGS:=
+GOTAGS:=ledger
```
pgoos, line 14: is this the only change needed to make ledger work with margin open/close?

timlind: This is just to enable it for the command line binary for any transactions. The backend support for margin is already working.

joneskm (approved):

PR #3306, opened 2022-10-05
LP value liabilities

AustinoBombino:
Account for liabilities in LP value query. Rename usage of balances to depths.


timlind (approved):
lgtm

canercandan (approved):
LGTM

PR #3327, opened 2022-10-14
feat: upgrade cosmos v0.45.x

timlind:
[x] upgrade to 0.45.9 with dragonberry patch


go.mod
```diff
@@ -3,11 +3,11 @@ module github.com/Sifchain/sifnode
...
 
 require (
-	github.com/BurntSushi/toml v1.1.0
+	github.com/BurntSushi/toml v1.2.0
 	github.com/MakeNowJust/heredoc v1.0.0
 	github.com/cespare/cp v1.1.1 // indirect
-	github.com/cosmos/cosmos-sdk v0.45.6
-	github.com/cosmos/ibc-go/v2 v2.0.2
+	github.com/cosmos/cosmos-sdk v0.45.9
+	github.com/cosmos/ibc-go/v2 v2.4.1
```
colin-axner, line 10: hey, just wanna double check you are aware ibc-go minor releases may have [state machine breaking changes](https://github.com/cosmos/ibc-go/releases/tag/v2.3.0) (requires coordinated upgrade)

timlind: @colin-axner thanks for the heads up, we'll likely punt this for now then. What do you mean by a coordinated upgrade?

joneskm (dismissed):

go.mod
colin-axner: A coordinated upgrade requires all validators to upgrade at the same block otherwise they will reach app hash mismatch. Does that make sense?

colin-axner: If you schedule an upgrade for a specific height, then it is totally fine to bump the ibc-go minor version. But if validators are meant to upgrade when they can and not tied to any height, then I'd advise to leave the minor version as it was

PR #3332, opened 2022-10-18
Feature/parameterized swap fee rates

joneskm:


proto/sifnode/clp/v1/params.proto
```diff
@@ -106,8 +106,8 @@ message SwapFeeParams {
 
 message SwapFeeTokenParams {
   string asset = 1;
-  string min_swap_fee = 2 [
-      (gogoproto.customtype) = "github.com/cosmos/cosmos-sdk/types.Uint",
+  string swap_fee_rate = 2 [
+      (gogoproto.customtype) = "github.com/cosmos/cosmos-sdk/types.Dec",
```
timlind, line 110: When changes types of a field it's probably a good idea to have a migrator overwrite the value before any attempt to load it.

AustinoBombino: This is my only concern, but if @joneskm think's we're fine then everything else looks good to me.

AustinoBombino (dismissed):

AustinoBombino (approved):

proto/sifnode/clp/v1/params.proto
joneskm: Agreed, I'll overwrite it in the upgrade handler.

timlind (approved):

PR #3346, opened 2022-10-25
cosmos-sdk update

faddat:
This should be in the next release.


snobbee:
duplicates of https://github.com/Sifchain/sifnode/pull/3422

PR #3361, opened 2022-11-04
feat: dynamic propsal fee

timlind:


x/admin/keeper/keeper.go
```diff
@@ -100,3 +100,21 @@ func (k Keeper) GetAdminAccounts(ctx sdk.Context) []*types.AdminAccount {
...
 	return accounts
 }
+
+func (k Keeper) SetParams(ctx sdk.Context, params *types.Params) {
+	store := ctx.KVStore(k.storeKey)
+	store.Set(types.ParamsStorePrefix, k.cdc.MustMarshal(params))
+}
+
+func (k Keeper) GetParams(ctx sdk.Context) *types.Params {
+	defaultSubmitProposalFee := sdk.NewUintFromString("500000000000000000000")
```
canercandan, line 110: @timlind the value seems to be equal to 500 rowan instead of 5000 is it intentional?

canercandan:
@timlind Test scenarios as described in the test plan are all successful: https://www.notion.so/sifchain/Testplan-Release-v1-0-15-beta-Asymmetric-Adds-275b1c9195e1492a83a6050a4bcbe70b#20701f3332fa49078392374be69eb4c6

there is unit test that seems to fail though, if you can have a look and fix it thanks.

canercandan (approved):
LGTM

PR #3423, opened 2023-10-27
ci: integrate go releaser

snobbee:
Integrate goreleaser to improve release process.


jzvikart (commented):
What exactly does goreleaser do and how does it benefit us?
Our current official way of building is "make install". This should always produce the official sifnoded binary. We should only have one way of building, and it should run locally in exactly the same way as in GitHub actions.

We already have too many different ways to do the same thing, so before we introduce any new tool like this it's important to make sure we don't introduce even more ways, but streamline the process and clean up everything except "the single source of truth".

snobbee:
> What exactly does goreleaser do and how does it benefit us? Our current official way of building is "make install". This should always produce the official sifnoded binary. We should only have one way of building, and it should run locally in exactly the same way as in GitHub actions.
> 
> We already have too many different ways to do the same thing, so before we introduce any new tool like this it's important to make sure we don't introduce even more ways, but streamline the process and clean up everything except "the single source of truth".

I understand the importance of maintaining a single, consistent build process. The intention behind proposing `goreleaser` is not to introduce another method of building, but to enhance our current one. Here's why:

1. **Cross-Platform Binaries:** While "make install" works well, it may not be the most efficient way to produce binaries for all major platforms. `goreleaser` excels in this regard, enabling us to easily generate binaries for multiple platforms without the complexity that might come with a Makefile alone.

2. **Adoption in the Cosmos Ecosystem:** Several projects within the Cosmos ecosystem have already adopted `goreleaser`. For instance, both [Osmosis](https://github.com/osmosis-labs/osmosis/blob/main/.goreleaser.yaml) and [Cosmos SDK](https://github.com/cosmos/cosmos-sdk/blob/main/.goreleaser.yml) use it. This not only showcases its reliability but also hints at the potential benefits it offers, especially in terms of standardizing builds across the ecosystem.

3. **Streamlined Release Process:** Take a look at the [Osmo release](https://github.com/osmosis-labs/osmosis/releases/tag/v20.1.0) to see the output `goreleaser` can produce. It not only builds the binaries but also structures them in a clean, organized manner, which can be extremely beneficial for our users and for ensuring clarity in our release process.

Integrating `goreleaser` doesn't mean we're introducing a new build method. Instead, we're enhancing our existing one, ensuring that it remains the single source of truth, but is more efficient, consistent, and in line with what's becoming a standard in the Cosmos ecosystem.

jzvikart:
@snobbee That's a great sales pitch, but it doesn't really answer the question of how it benefits us. For example, due to Peggy our build process has lot of dependencies and complications, including npm, docker, smart contracts, yarn, protobuf etc., They are highly complex and they break all the time, causing us a lot of grief. The real source of our pain is technical debt and goreleaser will most definitely not help us with that. The proper thing to do would be to focus on redesigning our build process from the ground up, in a way that covers all our needs holistically, succintly and comprehensively. It will take a lot of effort to pay off the debt, and until then, any partial solution will have a very limited benefits (if at all). I'm not opposing using goreleaser, my concern is that the longer we wait the higher the price we are paying.

pgoos (approved):
LGTM

pgoos:
> LGTM



> @snobbee That's a great sales pitch, but it doesn't really answer the question of how it benefits us. For example, due to Peggy our build process has lot of dependencies and complications, including npm, docker, smart contracts, yarn, protobuf etc., They are highly complex and they break all the time, causing us a lot of grief. The real source of our pain is technical debt and goreleaser will most definitely not help us with that. The proper thing to do would be to focus on redesigning our build process from the ground up, in a way that covers all our needs holistically, succintly and comprehensively. It will take a lot of effort to pay off the debt, and until then, any partial solution will have a very limited benefits (if at all). I'm not opposing using goreleaser, my concern is that the longer we wait the higher the price we are paying.

We don't even know if we're coming back to peggy so maybe let's postpone any efforts fixing the peggy build process. @jzvikart  what do you think?

jzvikart:
@pgoos Absolutely, if we're not going back to Peggy, dropping support for it would be a massive relief. But until we get a decision to permanently drop Peggy, we have to keep supporting it. Let's talk about this during the next meeting.

jzvikart:
@snobbee If you're going to uses this, please also update `make install` target so that it uses goreleaser and keeps builds consistent across different environments rather than just adding another target.

pgoos (approved):
LGTM

PR #3427, opened 2023-11-07
feat: add new CLP tx to add liquidity to rewards bucket

snobbee:
* add new CLP tx to add liquidity to rewards bucket
* add rewards bucket keeper functions
* add bank keeper mock object to mock bank keeper functions
* add unit tests for rewards bucket keeper functions
* add integration tests for the tx message
* add REST post endpoint


pgoos (approved):
LGTM looks great 🚀

PR #3429, opened 2023-11-14
feat: increase min gas price and min fee

snobbee:
* increase min gas price regardless of tx type
* increase high and low min fee
* fix security issue where one can stack up multiple msgs in same transaction and depending on the order of the msgs can benefits from a low fee, fixed by always using the max fee available


pgoos (dismissed):
Great work

pgoos (dismissed):

timlind (approved):

pgoos (approved):

PR #3430, opened 2023-11-15
Feat/rewards abci

snobbee:
* introduce new module epochs to manage time-based scheduled tasks for daily reward distribution
* create CLP hook for distribution of available rewards in reward buckets
* calculation of reward share for eligible lps
* eligible lp is an LP that locked liquidity in a pool for at least `lockPeriod`
* introduced new `lockedPeriod` reward params (set by default at 14 days)
* calculation of absolute reward amount to distribute for each eligible lp
* distribute reward amount to lp and substract the amount from reward bucket
* updated CLP cil to update reward params that includes both lockPeriod and epochIdentifier
* add `rewardsDistribute` param (default to false)
* if `rewardsDistribute` is false then the rewards get added to liquidity pool and eligible LP pool units get re-calculated and increased.
* keep track of all the rewards distributed or added to liquidity provider pool share at both LP and pool level


x/clp/keeper/liquidityprovider.go
```diff
@@ -189,3 +193,26 @@ func (k Keeper) GetAllLiquidityProvidersPaginated(ctx sdk.Context,
...
+	all, err := k.GetAllLiquidityProviders(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	return partitionLPsbyAsset(k.filterLiquidityProviders(ctx, all)), nil
+}
+
+// filterLiquidityProviders should only keep the LPs where their last updated block is at least 100000 blocks old from current block height
+func (k Keeper) filterLiquidityProviders(ctx sdk.Context, lps []*types.LiquidityProvider) []*types.LiquidityProvider {
```
timlind: Name can specify what the filter is.

x/clp/keeper/liquidityprovider.go
```diff
@@ -16,6 +16,10 @@ func (k Keeper) SetLiquidityProvider(ctx sdk.Context, lp *types.LiquidityProvide
 	if !lp.Validate() {
 		return
 	}
+	// if lp last updated block is 0, set it to current block height
+	if lp.LastUpdatedBlock == 0 {
+		lp.LastUpdatedBlock = ctx.BlockHeight()
```
timlind, line 21: Wouldn't you want to always update LastUpdatedBlock?

timlind: Also how would you want this to be affected on distribution?

snobbee: @timlind good catch, I added the `LastUpdatedBlock` update to the `AddLiquidity` keeper function

x/clp/keeper/liquidityprovider.go
snobbee: @timlind done

pgoos (dismissed):
Lgtm

timlind (approved):

pgoos (approved):

jzvikart (approved):

PR #3433, opened 2023-12-04
feat: update rewards params to update new reward parameters

snobbee:
* reward update msg to update new rewards distribute parameter
* reward update msg to update new rewards epoch identifier parameter
* reward update msg to update new rewards lock period parameter


pgoos (approved):
LGTM

